PR #11, opened 2019-07-26
My very first pull request

aliyss:
I came over your repo and since I found it particularly upto date I was wondering if you mind me adding some stuff or not. Here's a list of changes I made:

Added: isOnline and kind
I added the isOnline and the kind. Although kind only returns "chat" and "group" it might be good in future to check for enterprise-chats or similar in case it comes out. In my opinion isOnline can be very helpful.

Removed: fromMe check
Adding to that I removed the fromMe as I believe this should be an if else provided by the user directly.

Added: Return of the message, currently it only returns the string success, but i plan on making it work so that it returns the message sent.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -105,7 +105,9 @@ class Client extends EventEmitter {
 
 		// Register events
 		await page.exposeFunction('onAddMessageEvent', msg => {
+
 			if (!msg.isNewMsg) return;
+			console.log(msg);
 			this.emit(Events.MESSAGE_CREATE, new Message(this, msg));
 		});
 
@@ -137,14 +139,35 @@ class Client extends EventEmitter {
 	 * @param {string} message
 	 */
 	async sendMessage(chatId, message) {
-		let last_message = await this.pupPage.evaluate(async (chatId, message) => {
-			await Store.SendMessage(Store.Chat.get(chatId), message);
+		let msg = {};
+		if (message && typeof message === 'object' && message.constructor === Object) {
+			msg = message;
+			msg.mentions = [];
+			for (let i = 0; i < message.mentions.length; i++) {
+				let user = await Store.Contact.serialize().find(x => x.id.user === message.mentions[i]);
+				if (user && message.body.includes("@" + message.mentions[i] + " ")) {
+					msg.mentions.push(user.id)
+				}
+			}
+		} else {
+			msg = {
+				linkPreview: null,
+				mentions: [],
+				quotedMsg: null,
+				quotedMsgAdminGroupJid: null,
+				body: message
+			}
+		}
+		let last_message = await this.pupPage.evaluate(async (chatId, msg) => {
+			await Store.SendMessage(Store.Chat.get(chatId), msg.body, {linkPreview : msg.linkPreview, mentionedJidList : msg.mentions, quotedMsg : msg.quotedMsg, quotedMsgAdminGroupJid : msg.quotedMsgAdminGroupJid});
 			return Store.Chat.get(chatId).msgs._last.serialize()
-		}, chatId, message);
+
+		}, chatId, msg);
 		last_message = new Message(this, last_message);
-		if (last_message.body === message) {
+		if (last_message.body === msg.body) {
 			return last_message
 		}
+
 	}
 
 	/**
--- a/src/Client.js
+++ b/src/Client.js
@@ -14,160 +14,162 @@ const Message = require('./structures/Message');
  * @extends {EventEmitter}
  */
 class Client extends EventEmitter {
-    constructor(options = {}) {
-        super();
+	constructor(options = {}) {
+		super();
 
-        this.options = Util.mergeDefault(DefaultOptions, options);
+		this.options = Util.mergeDefault(DefaultOptions, options);
 
-        this.pupBrowser = null;
-        this.pupPage = null;
-    }
+		this.pupBrowser = null;
+		this.pupPage = null;
+	}
 
-    /**
-     * Sets up events and requirements, kicks off authentication request
-     */
-    async initialize() {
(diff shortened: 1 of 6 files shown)


Aliyss:
as promised, i added a return. After sending the message will be retrieved and returned to the user.

basically allows the user to use the following format:

let m = await client.sendMessage(message.from, content);

src/util/Injected.js
```diff
@@ -176,21 +176,27 @@ exports.LoadCustomSerializers = () => {
...
+
+        res.kind = chat.kind;
         res.isGroup = chat.isGroup;
         res.formattedTitle = chat.formattedTitle;
 
         if(chat.groupMetadata) {
             res.groupMetadata = chat.groupMetadata.serialize();
         }
 
+        if (!res.isGroup) {
```
pedroslopez: This `if` here isn't really needed since all chat objects have a presence, even though it will always be `false` for groups.

src/structures/Chat.js
```diff
@@ -17,6 +17,8 @@ class Chat extends Base {
         this.id = data.id;
 
         this.name = data.formattedTitle;
+        this.isOnline = data.isOnline;
```
pedroslopez: Since this only really applies for private chats, consider patching this in the `PrivateChat` model, instead of the more general `Chat` one.

src/Client.js
```diff
@@ -137,9 +137,12 @@ class Client extends EventEmitter {
      * @param {string} message 
      */
     async sendMessage(chatId, message) {
-        await this.pupPage.evaluate((chatId, message) => {
-            Store.SendMessage(Store.Chat.get(chatId), message);
-        }, chatId, message)
+        let last_message = await this.pupPage.evaluate(async (chatId, message) => {
+            await Store.SendMessage(Store.Chat.get(chatId), message);
+            return Store.Chat.get(chatId).msgs._last.serialize()
```
pedroslopez: This is actually a pretty nice idea, however, this implementation is error prone since you don't have any guarantees that the last message in the chat is the message you sent. 

I was able to break this with the example in the repo by putting my other phone in airplane mode, sending `!ping` along with a couple more messages, and then turning airplane mode off. All the messages arrive to our client seemingly at the same time, but in between the time it takes to send a reply to the `!ping` command and when you save that `last_message`, another one of the messages that were coming in was received, so the message sent by the client is in fact *not* the last message in the chat.

pedroslopez (changes requested):
Hey there! Thanks a lot for contributing! I really appreciate it. 

I've left you some comments you should take a look at before accepting this PR, but overall really nice ideas and additions.

In the future please remember to make the PR titles more descriptive, indicating a bit about what's contained inside. 😄

src/Client.js
Aliyss: Thanks for the fast reply. Will work on the other stuff after the weekend and plan on adding a lot more. Any idea on how it would be possible to successfully retrieve the sent message?because returning Store.SendMessage only outputs the string "success".

Aliyss: @pedroslopez Pinging you in case you didn't see it. Btw. do you use discord by any chance?

pedroslopez: Hey! Hadn't seen this. I'm not sure to be honest how we would be able to reliably get the latest message. Maybe it can be done matching the last message based on the content itself and things like user? Not sure.

Yeah I'm on discord, nacho#6294 if you want to message me.

src/Client.js
```diff
@@ -134,12 +135,37 @@ class Client extends EventEmitter {
...
+                mentions: [],
+                quotedMsg: null,
+                quotedMsgAdminGroupJid: null,
+                body: message
+            }
+        }
+        let last_message = await this.pupPage.evaluate(async (chatId, msg) => {
+            msg.mentionedJidList = [];
+            for (let i = 0; i < msg.mentions.length; i++) {
+                let user = await Store.Contact.serialize().find(x => x.id.user === msg.mentions[i].split("@")[0]);
```
pedroslopez: Since you're using Contact, would this work if the user isn't one of your contacts / if you don't have a chat open with them?

Also, maybe we should just pass an extra argument to the sendMessage function that's an object with these other properties, leaving the first argument as a string as is

src/Client.js
pedroslopez: I'm still not convinced on this issue... maybe leave this out until we find a more definitive way to get the sent message?

pedroslopez (changes requested):
Sorry it took so long to review this... I got insanely busy. Anyway, we're close to merging :)

Kouthk:
@Aliyss This function would be extremely useful to me, do you still think about finishing it? say yes pls haha

Aliyss:
@Kouthk afaik I integrated it and split up the pull requests. And as I didn't know how to use pull requests before this one contains a lot of diverse stuff.

If you can tell me what you need I can check if it has been added.

Kouthk:
I would like the function to display the last message sent by my device, something like "Last message from me". 
I developed a function for this, but it is not very beautiful or performative, as I am a beginner in the universe of programming...
I don't want to use the one developed by me outside the test environment

Aliyss:
@Kouthk You can check out the code below to see how to save your last message into a variable.


```js
let last_message;

client.on('message_create', (msg) => {
  // Fired on all message creations, including your own
  if (msg.fromMe) {
      last_message = msg
  }
});
```

or 

client.sendMessage(number, message);

```js
let last_message;

    console.log('MESSAGE RECEIVED', msg);

    if (msg.body === '!ping reply') {
        // Send a new message as a reply to the current one
        last_message = await msg.reply('pong');
    }
})
```

If you need help with your code you can try asking in the [discord server](https://discord.gg/wyKybbF).

Issue #17, opened 2019-10-09
Module Conn is not found


JRakhimov:
Module raid cannot found any elements in Conn module. there is error `Cannot read property 'default' of undefined` @pedroslopez

pedroslopez:
Yep, seems like `Conn` is gone. I've made some comments on the linked PR. If it's not addressed in the next day I'll implement the fix myself and issue a new release since it's kind of high priority.

Thanks for the report :)

pedroslopez:
Solved in 46a7d02 and 512b21e. Thanks for reporting!

PR #18, opened 2019-10-09
workaround for module raid not founding 'Conn' module 

renato-macedo:
and puppeteer update

This PR addresses Issue #17 
Found this problem today, maybe something changed on WhatsApp's side

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,17 +5,17 @@
   "requires": true,
   "dependencies": {
     "agent-base": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
-      "integrity": "sha512-JVwXMr9nHYTUXsBFKUqhJwvlcYU/blreOEUkhNR2eXZIvwd+c+o5V4MgDPKWnMS/56awN3TRzIP+KoPn+roQtg==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.3.0.tgz",
+      "integrity": "sha512-salcGninV0nPrwpGNn4VTXBb1SOuXQBiqbrNXoeizJsHrsL6ERFM2Ne3JUSBWRE6aeNJI2ROP/WEEIDUiDe3cg==",
       "requires": {
         "es6-promisify": "^5.0.0"
       }
     },
     "async-limiter": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/async-limiter/-/async-limiter-1.0.0.tgz",
-      "integrity": "sha512-jp/uFnooOiO+L211eZOoSyzpOITMXx1rBITauYykG3BRYPu8h0UcxsPNB04RR5vo4Tyz3+ay17tR6JVf9qzYWg=="
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/async-limiter/-/async-limiter-1.0.1.tgz",
+      "integrity": "sha512-csOlWGAcRFJaI6m+F2WKdnMKr4HhdhFVBk0H/QbJFMCr+uO2kwohwXQPxw/9OCxp05r5ghVBFSyioixx3gfkNQ=="
     },
     "balanced-match": {
       "version": "1.0.0",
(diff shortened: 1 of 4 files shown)


pedroslopez:
Hey! Thanks for PRing. 

Indeed, it seems WhatsApp have removed the `Conn` model. This module was only used to find the overall Store, so it's not really a big deal.

You approach seems to work, but it could be done in a more straightforward way. You could do the same thing I was doing before, but with another model instead of `Conn` (maybe `Msg`, `Chat`, or any others that are still referenced in the store).

pedroslopez:
Looks good. Can you squash commits into one?

JRakhimov:
Great, now instead of Conn we use Chat module, but then we cannot access to the me property.. Any idea how can we get account phone number?

JRakhimov:
so I found the solution for getMe

```js
page.getMe = async () =>
      await page.evaluate(() => {
        const me = Store.Contact.models.find(contact => contact.isMe);

        if (me) {
          return me.id.user;
        }

        return null;
      });
```

pedroslopez:
> Great, now instead of Conn we use Chat module, but then we cannot access to the me property.. Any idea how can we get account phone number?

How were you getting `me` before the change?

Also, I've just noticed we were using `Conn` to fire an event here:
(Client.js, lines 120-123)
```js
await page.evaluate(() => {
        Store.Msg.on('add', onAddMessageEvent);
        Store.Conn.on('change:connected', onConnectionChangedEvent);
})
```
I wonder if this has been moved somewhere else or if we won't be able to do this in this way... I can't check right now, will look more into it tonight.

JRakhimov:
> @JRakhimov How were you getting me before the change?

@pedroslopez before changes I was getting it through Conn module:
```js
page.getMe = async () => await page.evaluate(() => Store.Contact.get(window.Store.Conn.me).__x_id);
```
> Also, I've just noticed we were using Conn to fire an event here: 
@pedroslopez Chat module also has change:connected event but when I am logging out from sessions it doesnt emit it

pedroslopez:
> @pedroslopez before changes I was getting it through Conn module:
> 
> ```js
> page.getMe = async () => await page.evaluate(() => Store.Contact.get(window.Store.Conn.me).__x_id);
> ```

Oh I see... In that case, that wasn't really part of whatsapp-web.js itself. A PR could definitely be made to include that functionality in the Client (maybe client.getMe() function?), or I can add it when I get some time.

A greater issue right now is the connection event being gone from Conn. @JRakhimov have you seen when Chat is triggering its `change:connected` event?

JRakhimov:
> A greater issue right now is the connection event being gone from Conn. @JRakhimov have you seen when Chat is triggering its change:connected event?

@pedroslopez I have tried but unfortunately I could not find when it triggers, but my colleague have found another solution, I think it is better solution:

we can load another module STREAM with mR:

```js
const [{ STATE, STREAM, default: AppState }] = mR.findModule("STREAM")
```

```js
STATE: {
  CONFLICT: "CONFLICT"
  CONNECTED: "CONNECTED"
  DEPRECATED_VERSION: "DEPRECATED_VERSION"
  OPENING: "OPENING"
  PAIRING: "PAIRING"
  PROXYBLOCK: "PROXYBLOCK"
  SMB_TOS_BLOCK: "SMB_TOS_BLOCK"
  TIMEOUT: "TIMEOUT"
  TOS_BLOCK: "TOS_BLOCK"
  UNLAUNCHED: "UNLAUNCHED"
  UNPAIRED: "UNPAIRED"
  UNPAIRED_IDLE: "UNPAIRED_IDLE"
}

STREAM: {
  CONNECTED: "CONNECTED"
  DISCONNECTED: "DISCONNECTED"
  RESUMING: "RESUMING"
  SYNCING: "SYNCING"
}

AppState: {
  _events: {
    change:stream: [...],
    change:state: [...]
  },
  state: "CONNECTED",
  stream: "CONNECTED"
}
```

and then listen for changes:
```js
AppState.on("change:state change:stream", () => {})
```

JRakhimov:
So with this solution we can catch the moment when we lose internet connection with phone, in this case state will be Timeout and stream is Connected, by the way research is required

pedroslopez:
Awesome, thanks @JRakhimov!

Thank you @renato-macedo as well for the PR. I'm merging this now and implementing a fix for the disconnect event right after.

PR #20, opened 2019-10-23
Added events which are fired when deleting messages && Added Contact Support similar to PrivateChat and GroupChat

aliyss:
- check the example.js for examples on how to use it
- message_revoke_everyone: is fired when the message is deleted for everyone.
- message_revoke_me: is fired when the message is deleted from the clients view.


pedroslopez:
Hey! Thanks for contributing. I've now been able to test this, and it seems only the MESSAGE_REVOKE_ME event is being triggered. I am unable to reproduce the `everyone` variant.

Following the same line, maybe a more generalized message deleted event would also be good to have.

Aliyss:
> Hey! Thanks for contributing. I've now been able to test this, and it seems only the MESSAGE_REVOKE_ME event is being triggered. I am unable to reproduce the `everyone` variant.

Fixed it now... should work now.

Aliyss:
@pedroslopez should work now.

Aliyss:
Its alright will post the updated version tomorrow as a new git pull

pedroslopez:
This looks really good and I’d like to get this merged. However, there’s a lot here. Can you please separate this into a couple PRs based on feature? Like have one PR for the Contacts feature, another for the delete events, another for mentions, etc. Thanks!

Aliyss:
Yes ofc. I totally forgot. Will do. Thanks for noticing.

Aliyss:
Finished with the other PRs

PR #22, opened 2019-11-12
Receive images feature

renato-macedo:
What do you think of this implementation? I'm using puppeteer to navigate through the page and convert the blob of the image to base 64.
But I'm still not happy with this because if the user send two or more images at once only captures the last one.

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,73 +1,73 @@
-const { Client } = require('./index')
+const fs = require('fs');
+const path = require('path');
+const { Client } = require('./index');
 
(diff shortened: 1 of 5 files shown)


pedroslopez:
Hi @renato-macedo ! 

Thank you very much for your contribution! Your method inspired me to try implementing this in a more native way. This can now be found on release 0.3.0 of the library. You can check out more info on how it's used in the release: https://github.com/pedroslopez/whatsapp-web.js/releases/tag/v0.3.0.

PR #28, opened 2019-12-19
Added function sendMessageToId to send message without chat before

SimiPrambos:
same as **sendMessage** function but if you don't have conversation before, you can use **sendMessageToId** .. check the example.js how to use this function

### DIFF
--- a/example.js
+++ b/example.js
@@ -35,10 +35,19 @@ client.on('message', async msg => {
         // Send a new message to the same chat
         client.sendMessage(msg.from, 'pong');
 
+    } else if (msg.body.startsWith('!sendto ')) {
+        // Direct send a new message to specific id
+        let number = msg.body.split(' ')[1]
+        let messageIndex = msg.body.indexOf(number) + number.length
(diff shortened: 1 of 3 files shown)


GliderCodes:
Hey there, just wondering what the id is? is it the number or what else could it be? Also if I have a number how do I take an ID from it if it's different.

pedroslopez (changes requested):
Sorry for the long wait! I finally had the chance to check this out. It works great, thanks for contributing!

A little has changed in regards to sending messages, so please rebase with the latest changes and take a look at how the new `sendMessage` function works, specifically the change from `Store.SendMessage` to `WWebJS.sendMessage` and the returning of the new message. It shouldn't be too much of a change.

Just a little comment was added as well. Thanks again for this!

tarikhagustia:
any update on this PR? i need this function too :)

pedroslopez:
Hi @tarikhagustia ! It seems the original PR author has gone inactive. I will clean this up and merge it later tonight.

pedroslopez:
superseeded by #52

PR #29, opened 2019-12-24
auto login from saved session data

gabriel-tandil:
save and read session data for login in a json file

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,6 +1,11 @@
+var fs = require('fs');
 const { Client } = require('./index')
 
-const client = new Client({puppeteer: {headless: false}});
+if (fs.existsSync('./config.json')) {
+  var sessionCfg = require('./config.json');
+}
+
+const client = new Client({puppeteer: {headless: false} ,session:sessionCfg});
 // You can use an existing session and avoid scanning a QR code by adding a "session" object to the client options.
 // This object must include WABrowserId, WASecretBundle, WAToken1 and WAToken2.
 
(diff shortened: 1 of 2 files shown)


gabriel-tandil:
you have a beautiful project here, a very small contribution.

pedroslopez:
Thanks for the contribution! 

Personally, I think I would prefer the library not write these things to disk and opt for whoever is using it to take care of that instead. In my eyes, there are a couple of things that may cause it to not work as intended, so I would leave it up to the user to decide what is best for what they're trying to do.

pedroslopez:
@gabriel-tandil Looking at this again, I think we can do it. For whatever reason when I first reviewed this I assumed it was going into the library, but upon checking again I see that this is merely a contribution to the example. In this case, I think it's ok we provide something as a base for other people to build upon. Regardless, this whole example.js thing is less than ideal and I'd like to have a proper guide up at some point. 

If you want, you can reopen this and we can get it merged. My only comments would be to rename `config.json` to maybe `session.json` and try to avoid the use of `var`, instead opting for `const` whenever possible and `let` where it's not.

iamshouvikmitra:
I want to auto generate new session if the session details found in ./session.json is old. How is that possible?

pedroslopez:
@iamshouvikmitra An "auth_failure" event is emitted when a session restore is not possible. You can use this to detect it and delete the existing session.

iamshouvikmitra:
Yeah, @pedroslopez  would you mind sharing a snippet of the same implementation. I am a little confused about killing the previous session. And defining a new client.


Ex - 
```
let client = new Client({session : cfg})


client.on('auth_failure', () => {
    client = new Client ()
})

doesn't seems to be an working solution.
```

PR #33, opened 2020-02-02
Added the delete event

aliyss:
Works for delete for me and delete for everyone. Tested on 2020.02.02.

### DIFF
--- a/example.js
+++ b/example.js
@@ -135,7 +135,7 @@ client.on('message_revoke_everyone', async (before, after) => {
 
 client.on('message_revoke_me', async (msg) => {
     // Fired whenever a message is only deleted in your own view.
-    console.log(msg); // message before it was deleted.
+    console.log(msg.body); // message before it was deleted.
 });
 
 client.on('disconnected', () => {
--- a/example.js
+++ b/example.js
@@ -127,10 +127,10 @@ client.on('message_create', (msg) => {
     }
 });
 
-client.on('message_revoke_everyone', async (before, after) => {
+client.on('message_revoke_everyone', async (after, before) => {
     // Fired whenever a message is deleted by anyone (including you)
-    console.log(before.body); // message before it was deleted.
     console.log(after.type); // message after it was deleted.
+    console.log(before.body); // message before it was deleted.
 });
 
 client.on('message_revoke_me', async (msg) => {
--- a/example.js
+++ b/example.js
@@ -127,6 +127,17 @@ client.on('message_create', (msg) => {
     }
 });
 
+client.on('message_revoke_everyone', async (before, after) => {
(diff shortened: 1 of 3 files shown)


Aliyss:
@pedroslopez Okay delete_event should now work. Please don't delete the other pull request yet.
Since i haven't figured git out yet, I am copy pasting the stuff back and forth (Don't judge, just too lazy to learn it). :/

pedroslopez:
The only think I don't quite like in this implementation would be the bit where it gets the deleted message based on some past changed event. In testing I see that it seems like some other change event gets called for that specific message right before the revoke occurs, but I'm not sure if there could be other factors that may lead to some other change event happening on some other message. At least you've properly guarded it by checking the message ID, but that means it's kind of unreliable and you may not fire the event.

There's a tradeoff here between always having the previous message available when we fire the revoke event vs always being able to receive the revoke event and potentially not have the deleted message.

I think we should opt for the latter: fire the event anyway, even though the previous message was not found, just to ensure that all revokes are caught, and I guess offering that previous message as a bonus. The implementation would be pretty much the same as what you have right now, but moving the `msg.id.id === last_message.id.id` inside the conditional and reversing the order of the arguments returned in the event. (It would end up being `client.on(revokedMsg, originalMsg)`, not necessarily with those var names, but where `revokedMsg` is the current instance of the message and `originalMsg` is the old one that you've saved based on the previous change event.)

Thoughts?

Aliyss:
The problem is mainly with the message body. If I fire the Event without saving the last message it will fire the event, but there is no way of actually retrieving the content at any point after it has been revoked. Thus the "Added bonus" will actually be no bonus at all.

The only possibility I see is adding a MESSAGE_CHANGED event and having all changes being sent through that. Then we could leave the checks to the user-side.

pedroslopez:
> The problem is mainly with the message body. If I fire the Event without saving the last message it will fire the event, but there is no way of actually retrieving the content at any point after it has been revoked. Thus the "Added bonus" will actually be no bonus at all.

Well, I'm not saying _not_ to "cache" the last_message, I'm just suggesting firing the event even if the last_message isn't the message being revoked, but just not having the message in that case.

It would look something like this

```js
let last_message;

 await page.exposeFunction('onChangeMessageEvent', (msg) => {

            if (last_message && msg.type === 'revoked') {
                const message = new Message(this, msg);
                let revoked_msg = undefined;
                if(msg.id.id === last_message.id.id) {
                        revoked_msg = new Message(this, last_message);
                }
                this.emit(Events.MESSAGE_REVOKED_EVERYONE, message, revoked_msg);
            } else {
                if (msg.type !== 'revoked') {
                    last_message = msg;
                }
            }
 });
```
(I typed this up on github so I have no idea if this actually works)

PR #34, opened 2020-02-02
Added Contact Model.

aliyss:
Allows one to use the contact model. This can be useful for getting more/other information about a chat/contact.

### DIFF
--- a/src/structures/Contact.js
+++ b/src/structures/Contact.js
@@ -1,0 +1,43 @@
+'use strict';
+
+const Base = require('./Base');
+
+/**
+ * Represents a Contact on WhatsApp
+ * @extends {Base}
+ */
+class Contact extends Base {
+    constructor(client, data) {
+        super(client);
+
+        if(data) this._patch(data);
+    }
+
+    _patch(data) {
+        this.id = data.id;
+        this.isBusiness = data.isBusiness;
+        this.isEnterprise = data.isBusiness;
+        this.labels = data.labels;
+        this.name = data.name;
+        this.pushname = data.pushname;
+        this.sectionHeader = data.sectionHeader;
+        this.shortName = data.shortName;
+        this.statusMute = data.statusMute;
+        this.type = data.type;
+        this.verifiedLevel = data.verifiedLevel;
+        this.verifiedName = data.verifiedName;
+
+        return super._patch(data);
+    }
+    
+    /**
+     * Gets the Profile Image of the User
+     */
+    async getProfileImageUrl() {
+        return this.client.pupPage.evaluate(contactId => {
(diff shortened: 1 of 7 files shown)


src/structures/Contact.js
```diff
@@ -0,0 +1,43 @@
...
+    constructor(client, data) {
+        super(client);
+
+        if(data) this._patch(data);
+    }
+
+    _patch(data) {
+        this.id = data.id;
+        this.isBusiness = data.isBusiness;
+        this.isEnterprise = data.isBusiness;
```
pedroslopez: Typo here? Should be `data.isEnterprise`

pedroslopez (changes requested):
While we're at it I think we can include more things like `isGroup`, `isMe`, `mentionName` and whatever you see that may be useful

src/structures/Contact.js
Aliyss: Yes ofc. My bad.

Aliyss: fixed

src/structures/Contact.js
```diff
@@ -0,0 +1,43 @@
...
+        this.verifiedLevel = data.verifiedLevel;
+        this.verifiedName = data.verifiedName;
+
+        return super._patch(data);
+    }
+    
+    /**
+     * Gets the Profile Image of the User
+     */
+    async getProfileImageUrl() {
```
pedroslopez: This doesn't seem to be working. Two issues I see here:

1. Store.ProfilePicThumb seems to be lazily loaded in some way, which could cause the `get` to not return anything. 
2. When it does find it, I'm not getting back a profile image url, but rather simply an object with the contact ID.

On another note, `Contact` has a `profilePicThumb` property, so we can probably load this in when it's serialized with `WWebJS.getContactModel`, removing the need to use an additional async function here to get the profile image

pedroslopez (changes requested):

src/structures/Contact.js
pedroslopez: I took another look at this and Contact.profilePicThumb isn't reliable either. They can potentially not be set. However, I found another module we can use:

`mR.findModule("profilePicFind")[0]`
 
It has a couple functions related to getting and setting profile images. With the following we can ensure to always get the profile image URL for any WhatsApp user. The only caveat is if the profile has their profile picture privacy settings set to ignore the current user, it will return a 401.

Aliyss: I rechecked... Seems only to work with Contacts that are on the screen.
For now i can remove the profilepicThumb and it can be integrated later on?

pedroslopez: Yeah that sounds ok. I have an idea for how to do it so I could take a crack at it.

PR #35, opened 2020-02-05
Get Mentions

aliyss:
I added it like this, but honestly it isn't really optimal. The first set returns id's only. The function converts the ids to Contacts and returns it as an array of Contacts. Do you see a more likeable solution?


Aliyss:
On the other hand I could make it it's own object. As a collection of Contacts?

src/structures/Message.js
```diff
@@ -90,6 +90,16 @@ class Message extends Base {
          */
         this.hasQuotedMsg = data.quotedMsg ? true : false;
 
+        /**
+         * Indicates the mentions in the message body.
+         * @type {[id]}
```
pedroslopez:
```suggestion
         * @type {Array<string>}
```

src/structures/Message.js
```diff
@@ -113,6 +123,19 @@ class Message extends Base {
         return this.client.getContactById(this._getChatId());
     }
 
+    /**
+     * Returns the Contacts mentioned in this message
+     * @returns {[Promise<Contact>]}
```
pedroslopez:
```suggestion
     * @returns {Promise<Array<Contact>>}
```

src/structures/Message.js
```diff
@@ -90,6 +90,16 @@ class Message extends Base {
          */
         this.hasQuotedMsg = data.quotedMsg ? true : false;
 
+        /**
+         * Indicates the mentions in the message body.
+         * @type {[id]}
+         */
+        this.mentions = [];
```
pedroslopez: Rename `mentions` to `mentionedIds` to indicate the difference between this and the other function.

pedroslopez (changes requested):
This looks fine to me! Just a couple comments.

src/structures/Message.js
```diff
@@ -113,6 +123,19 @@ class Message extends Base {
         return this.client.getContactById(this._getChatId());
     }
 
+    /**
+     * Returns the Contacts mentioned in this message
+     * @returns {[Promise<Contact>]}
+     */
+    async getMentions() {
+        let mentions = [];
+        for (let i = 0; i < this.mentions.length; i++) {
```
pedroslopez: I would prefer using `map` over loops like these. Makes it clearer.

Aliyss:
Okok will push the changes in the morning.

Aliyss:
Doesn't give an error if own user is mentioned, but doesn't return clean data.

PR #46, opened 2020-02-11
Starting the browser without a new tab

fernandolcardoso:
Would there be a possibility to change the script so as not to start the browser with a new tab? This would save a little memory.


pedroslopez:
Doing a quick review, it seems this does not work. Page is returning null.

PR #63, opened 2020-02-19
Archive chats

aliyss:
I wasn't sure how to implement it. So I added a ```archive``` and an ```unarchive``` method.

The other two possibilities:
- Adding a toggleArchive method. Saves coding space and leaves the checking to the user.
- Passing true/false as a parameter in a single method (but I couldn't think of a name that wouldn't be misleading).

//Edit: Grammer


pedroslopez:
This doesn't seem to work. While setting it to archived appears to work on the whatsapp web app, if you look on the phone or restart WhatsApp Web, you will see that it doesn't really archive it.

pedroslopez (changes requested):
This doesn't seem to work. While setting it to archived appears to work on the whatsapp web app, if you look on the phone or restart WhatsApp Web, you will see that it doesn't really archive it.

On another note, the `!unarchive` command example doesn't quite make sense. It seems like the function is expecting a chatId, but in practice it's just sending `!unarchive` to it.

Aliyss:
Good point will see what I can do.

pedroslopez:
While looking for a way to delete messages (1c60c83) I found some functions that can be of use to you. `Store.Cmd` is now exposed, which has a function `.deleteOrExitChat()` and another `.archiveChat()`

The archive function seems to work fine, but the deletion one shows a popup instead of directly deleting it, so we'll have to take a look at what that calls internally after accepting the popup.

PR #71, opened 2020-02-24
Added Chat Events.

aliyss:
- Group join
- Group leave
- Group events like description change, Name change.
- Not implemented: Change of Admin.

### DIFF
--- a/example.js
+++ b/example.js
@@ -178,6 +178,18 @@ client.on('message_create', (msg) => {
     }
 });
 
+client.on('chat_joined', (msg) => {
+    msg.reply('User joined.');
+});
+
+client.on('chat_left', (msg) => {
+    msg.reply('User left.');
+});
+
+client.on('chat_event', (msg) => {
+    msg.reply('Chat did something.');
+});
+
 client.on('message_revoke_everyone', async (after, before) => {
     // Fired whenever a message is deleted by anyone (including you)
     console.log(after); // message after it was deleted.
--- a/src/Client.js
+++ b/src/Client.js
@@ -148,9 +148,18 @@ class Client extends EventEmitter {
         // Register events
         await page.exposeFunction('onAddMessageEvent', msg => {
             if (!msg.isNewMsg) return;
-
             const message = new Message(this, msg);
 
+            if (msg.type === 'gp2') {
+                if (msg.subtype === 'remove') {
+                    this.emit(Events.CHAT_LEAVE, message);
+                } else if (msg.subtype === 'add') {
+                    this.emit(Events.CHAT_JOIN, message);
+                } else {
+                    this.emit(Events.CHAT_EVENT, message);
+                }
+                return;
+            }
             /**
              * Emitted when a new message is created, which may include the current user's own messages.
              * @event Client#message_create
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -108,6 +108,16 @@ class Message extends Base {
             this.mentionedIds = data.mentionedJidList;
         }
 
+        /**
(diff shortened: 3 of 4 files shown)


Aliyss:
Probably it would be nice to have a separate Message Constructor for the Group Events. Something like MessageInfo.

pedroslopez:
I think we should handle this type of message differently to avoid confusion, following your idea of creating a new class for them. Maybe this new class shouldn't even extend `Message` to make sure they're treated different.

My idea is making a GroupAction or GroupNotification class (i'm not set on the name yet) for this case, instead of a regular Message. This class would have the following properties:

- **chat**: Points to the `Chat` that this is for
- **type** : The gp2 message subtype (add, invite, remove, leave, subject, description, picture, announce...)
- **author**: The Contact that performed this action. This is the author prop on the gp2 message, and is set on everything except for subtypes `invite` and `leave`, as far as I could tell, because this is not executed by anyone.
- **targets**: The Contacts that were affected by this action. This is the `recipients` prop on the gp2 message in whatsapp (In case of subtypes add, invite, remove and leave).
- **value**: The `body` prop on the gp2 message. As far as I could tell this is only user with the `subject`, `announce` and `restrict` subtypes. In case of `subject`, it contains the new subject. In case of `announce` and `restrict`, it contains "on" or "off", indicating if the option has been disabled or enabled. Take a look at the section at the end of this post to see more about what the announce and restrict subtypes mean.

For `author` and `targets` initially let's return the Contact but later I think we should make a GroupParticipant class that can possibly extend Contact, so we have more info on it like if the user is an admin as well as take action directly on them via methods like `.kick()` or `.promote()`.

As far as the events themselves go, I think we could group them like this:
- **group_join**: emitted when subtype is either `add` or `invite`.
- **group_leave**: emitted when subtype is either `remove` or `leave`.
- **group_update**: emitted when subtype is either `subject, description, picture, announce,` or `restrict`.

A little info on what each of the gp2 subtypes mean:
- **add**: A user has been added to the chat by an admin
- **invite**: A user has joined the chat by using an invitation link
- **remove**: A user has been kicked from the chat by an admin
- **leave**: A user has left the chat voluntarily
- **subject**: The group subject has been changed.
- **description**: The group description has been changed.
- **picture**: The group picture has been changed.
- **announce**: The group settings option for "Send messages" has been modified. On/off in the body indicates if only admins can send messages in the chat or not.
- **restrict**: The group settings option for "Edit group info" has been modified. On/off in the body indicates if only admins can edit group subject, description and picture.

I can't say for certain if these are the only subtypes available, but these are the ones I've found so far.

PetrosM263:
Hi, i am trying to setup this whatsapp API Javascript  project.Below is how i setup my javascript code but i am getting no response. Is the anyone with a solution. Please help me i want to catch up with you. 

`
<script src="https://github.com/pedroslopez/whatsapp-web.js.git"></script>
<script>
const { Client } = require('whatsapp-web.js');
const client = new Client();

client.on('qr', (qr) => {
    // Generate and scan this code with your phone
    console.log('QR RECEIVED', qr);
});

client.on('ready', () => {
    console.log('Client is ready!');
});

client.on('message', msg => {
    if (msg.body == '!ping') {
        msg.reply('pong');
    }
});

client.initialize();
</script>
`


You can come to my whatsapp inbox for personal inboxassistance if you don't mind. [https://wa.me/263783059999](url)

Aliyss:
@PetrosM263 you have to use it as a Nodejs application and not in the web. Google for Nodejs app. That should help for a start. Or follow the Readme instructions.

PetrosM263:
So i can only implement this project on a Nde.js server not just a general server

Issue #43, opened 2020-02-10
Is there a function like isOnline() or isConnected()?


fernandolcardoso:
Hello person!

Is there a way for me to check if the Client is online / connected before forwarding messages?

pedroslopez:
Hi! Are you referring to a particular Contact or the Client itself?

The client emits a `ready` event when everything is ready to go, if that's what you're asking.

fernandolcardoso:
I am referring to the Client himself.

I am conducting tests with an iPhone and I noticed that several times, the connection is interrupted and at the time of shipment, it is that the Customer realizes that he is disconnected.

At this moment, as I use a database for sending, my system updates the message, as if it were forwarded.

I hope you understand.

pedroslopez:
Can you describe a bit more your implementation/issue so I can understand?
My understanding is, when you go call the `sendMessage` function the client disconnects. Is this correct?

As a general rule of thumb, the device should always be on and connected to the internet.

fernandolcardoso:
I am testing a script in parallel, known as "sulla-hotfix", in this script there is an isOnline() function that returns true or false, making sure that the Client is connected.

In the script "whatsapp-web.js", I witnessed situations in which I should be connected, but when forwarding the message, a disconnected message was presented.

fernandolcardoso:
My understanding is, when you go call the sendMessage function the client disconnects. Is this correct?

@pedroslopez correct

pedroslopez:
Are you able to reproduce this situation? Please give me steps or example scenario/code so I can test and see if it's something wrong with the library dropping connections. 

The client should not disconnect when sending a message.

pedroslopez:
Also, what disconnect reason are you getting?

fernandolcardoso:
@pedroslopez I will try to explain what happens to me.

I added the "mysql2" package and developed a small script about setInterval that performs periodic queries in a database.

When there are messages in the table, the information is captured, projected in the script and then an update is made that the message has been forwarded.

In some cases, the script only detects that it is disconnected after the sendMessage command is used, which generates unnecessary updates to the table.

I think the design of "whatsapp-web.js" is much better than that of "sulla-hotfix", however, when it disconnects, the script is declared as closed.

I think there should be an isOnline to check if it is active, if it is not, it should auto-reconnect.

Sorry for my English, I'm from Brazil and I'm using the translator.

pedroslopez:
[...]

PR #77, opened 2020-02-27
Add Force state update

vlachdev:
Hello

Whatsapp web will not update state until some action, this will force update client state.
First will set state to TIMEOUT, then to current state (or stay timeouted)

I just utilized information provided by @pedroslopez 
https://github.com/pedroslopez/whatsapp-web.js/issues/43#issuecomment-584988672

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -433,10 +433,10 @@ class Client extends EventEmitter {
     /**
      * Force reset of connection state for the client
     */
-   async resetState(){
-    await this.pupPage.evaluate(() => {
-        window.Store.AppState.phoneWatchdog.shiftTimer.forceRunNow();
-    });
+    async resetState(){
+        await this.pupPage.evaluate(() => {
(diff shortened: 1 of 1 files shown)


LinkTree3:
I have an issue that when the whatsapp web client is disconnected due to no internet and then the internet returns it takes up to 6-7 minutes sometimes for the client to get connected again.
It's either a few seconds to reconnect or 6-7 mins

`resetState` doesn't seem to be working after a internet disconnect

`window.Store.AppState.phoneWatchdog.shiftTimer.forceRunNow();`
also seems to return an uncaught type error when called directly in the browser
`Uncaught TypeError: Cannot read property 'cast' of undefined`

It only works if the client is already connected.


The way I solved it for now is with an interval that checks every 30 seconds the state of the connection using `getState` and if after 3 times it hasn't reconnected yet I close the session using `logout` and check the internet connection and wait for it to return and then run `initialize` again. This way I only need to wait about 90 seconds instead of 6-7 minutes. But I don't like the fact I have to reset the whole session.

PR #79, opened 2020-02-27
better sesion data save

gabriel-tandil:
For more consistency whenever I have a successful authentication I write
to the json file the session data

### DIFF
--- a/example.js
+++ b/example.js
@@ -20,14 +20,12 @@ client.on('qr', (qr) => {
 
 client.on('authenticated', (session) => {
     console.log('AUTHENTICATED', session);
-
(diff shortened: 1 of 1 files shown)


iamshouvikmitra:
Using userDataDir args for puppeteer to store the browser state also works like a charm. We could remove all the session code altogether itself. It also helps increase the boot time for the browser. Something you guys could look into.

pedroslopez:
> Using userDataDir args for puppeteer to store the browser state also works like a charm. We could remove all the session code altogether itself. It also helps increase the boot time for the browser. Something you guys could look into.

Oh I was actually not aware of this, thanks for pointing it out. I’ll take a look and probably add it to the documentation I’m writing.

PR #87, opened 2020-02-29
typing in chat

### LINKED ISSUE #72: Simulate '...typing' in the chat. sendChatstateComposing
Code used in the sulla-hotfix project https://github.com/smashah/sulla/

{ id: "WapQuery", conditions: (module) => (module.queryExist) ? module : ((module.default && module.default.queryExist) ? module.default : null) },

/**
 * Simulate '...typing' in the chat.
 *
 * @param {string} chatId '000000000000@c.us'
 * @param {boolean} on true to turn on similated typing, false to turn it off //you need to manually turn this off.
 */
window.WAPI.simulateTyping = async function (chatId, on) {
    if (on) await Store.WapQuery.sendChatstateComposing(chatId)
    else await Store.WapQuery.sendChatstateComposing(chatId)
};

andersondeoliveiramachado:
I cannot access the module using:

window.Store.WapQuery = window.mR.findModule('WapQuery').default;

only using:

 window.Store.WapQuery = webpackJsonp([], null, ["begfaahgdb"]).default;

module in  : 

app.a6bc2c573f664d4f1feb.js

### DIFF
--- a/example.js
+++ b/example.js
@@ -168,6 +168,9 @@ client.on('message', async msg => {
     } else if (msg.body === '!archive') {
         const chat = await msg.getChat();
         chat.archive();
+    } else if (msg.body === '!typing') {
+        const chat = await msg.getChat();
+        chat.typing();        
     }
 });
 
--- a/src/structures/Chat.js
+++ b/src/structures/Chat.js
@@ -102,6 +102,16 @@ class Chat extends Base {
     async unarchive() {
         return this.client.unarchiveChat(this.id._serialized);
     }
+   
+    /**
+     * typing in chat
+     */
+    async typing() {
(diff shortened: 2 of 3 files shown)


pedroslopez:
I believe your “WapQuery” may be the same as “Wap” that I’ve already exposed? Please check if the function you want to use is present there.

andersondeoliveiramachado:
> I believe your “WapQuery” may be the same as “Wap” that I’ve already exposed? Please check if the function you want to use is present there.

You're right. The function is already part of the Wap module

pedroslopez:
Just in case someone is looking at this PR, I made some changes right after in commit b07b38b that also sends recording state and allows clearing.

PR #94, opened 2020-03-02
Send Sticker

andersondeoliveiramachado:


InstantlyMoist (approved):

Aliyss:
Does this not work?

pedroslopez:
Well, I would say it's incomplete. I should have commented on this much earlier, I apologize for that. From the original issue, it seemed like it was a WIP PR and was never finished. Since it was more of a work in progress, I kind of let it be, but I should have followed up.

This PR adds a `sendSticker` function to the injected functions on the whatsapp web page, but does not expose it in any way to the library itself. Furthermore, it's unclear how you would actually use this to send stickers. What should the `sticker` parameter look like? Is it an image? Do you need to send an already existing sticker from a pack? Lastly it doesn't really conform to the way messages are being sent with the library. Sending messages should ideally be done through the general `sendMessage` function and return back the message that was just sent. 

If anybody wants to tackle this and take over, please do so since it seems that the original author has gone inactive for this project. Right now I don't have enough time to dedicate towards the feature, but I will get to it at some point.

andersondeoliveiramachado:
image

ncesar:
@pedroslopez I think it was finished by the previous author, wasnt it?

Aliyss:
@ncesar I think the main idea was to make it more like WWebJS -> Thus implementing the functionality in the already existing sendMessage function.

PR #102, opened 2020-03-10
feat: included call to generate the qrcode at the terminal

carlosrodj:
qrcode-terminal module included to generate qrcode.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "whatsapp-web.js",
-  "version": "1.0.2-post",
+  "version": "1.1.0-post",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


pedroslopez:
Hi Carlos! Thanks for contributing.

While this is exactly how I do it and exactly the library I use, I don't want to bake this into this library itself, and would rather leave it up to the user. 

This could be used for projects that may not even log anything to the console at all or in projects where multiple clients are being started, which would render different QR codes to the terminal without proper distinction and may lead to confusion.

I'm open to hearing different opinions and feedback on this approach.

pedroslopez:
We could include this as part of the example, but I wouldn't want to add an extra dependency for something that's only used in the example. I've actually already included this in the upcoming guide I'm working on (https://waguide.pedroslopez.me/) but that's still in progress.

PR #110, opened 2020-03-20
Add a reply to all whatsapp groups fuction

PetrosM263:
I was thinking , it would be great if we can have a function that allows us to reply to all groups at once without to reply one by one.


pedroslopez:
Seems like this PR should have been an issue instead? Also, please clarify what you mean by replying to all groups. Messages are replied to individually.

PR #112, opened 2020-03-22
Added "restartOnAuthFail" as an option

### LINKED ISSUE #109: Handle 'auth_failure' more gracefully
In the current example, `auth_failure` when a sessions file is supplied leads to a roadblock. You need to kill the application and the file. However the browser shows that the WhatsApp client has no such issue with it, rather it instead shows a QR code as expected but the 'qr' event does not fire as mentioned in the comment: 

`// NOTE: This event will not be fired if a session is specified.`

Currently my method is to kill the sessions file and kill the app, then restarting it with something monitoring the process.
```
client.on('auth_failure', msg => {
	console.error('AUTHENTICATION FAILURE', msg);
	console.log('Deleting the sessions file: ('+ SESSION_FILE_PATH+ ')\nPlease restart the app.');
	fs.unlinkSync(SESSION_FILE_PATH);
	process.exit();
});
```

Would it be possible to enhance this so that `QR` event can be fired and/or if we can re-initialize with sessions file omitted as a recovery option?

stevefold:
- When true, the sessions passed is discarded when reinitialized.

Addresses this issue: https://github.com/pedroslopez/whatsapp-web.js/issues/109

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -75,7 +75,11 @@ class Client extends EventEmitter {
                      */
                     this.emit(Events.AUTHENTICATION_FAILURE, 'Unable to log in. Are the session details valid?');
                     browser.close();
-
+                    if (this.options.restartOnAuthFail) {
+                        // session restore failed so try again but without session to force new authentication
+                        this.options.session = null;
+                        this.initialize(this.options);
+                    }
                     return;
                 }
 


pedroslopez:
LGTM! Thanks for contributing.

src/Client.js
```diff
@@ -75,7 +75,11 @@ class Client extends EventEmitter {
                      */
                     this.emit(Events.AUTHENTICATION_FAILURE, 'Unable to log in. Are the session details valid?');
                     browser.close();
-
+                    if (this.options.restartOnAuthFail) {
```
AviFix: Should be: this.options.puppeteer.restartOnAuthFail

src/Client.js
```diff
@@ -75,7 +75,11 @@ class Client extends EventEmitter {
...
+                        // session restore failed so try again but without session to force new authentication
+                        this.options.session = null;
+                        this.initialize(this.options);
```
AviFix: initialize func takes no params

AviFix (commented):
2 comments and The restart doesn't work when the timeout error is thrown

src/Client.js
stevefold: Why? We're checking the options passed to client()...

src/Client.js
stevefold: Yes it does.

stevefold:
Hey @AviFix, I've tested this and it works every time. Even when timeout is thrown (which it probably shouldn't, I can make it check to see if the QRCode is shown and if so, **then** we restart without auth)... I'll make it even more _graceful_ at some point.

src/Client.js
pedroslopez: This seems fine to me

src/Client.js
pedroslopez: @AviFix is right, initialize does not take any params. It uses the options object on the client. I'll make a quick commit to change this.

stevefold: My mistake! :)

PR #121, opened 2020-03-29
Hide the info bar in the browser

pauloboc:


pedroslopez:
Hi Paulo

Is there a reason why this should be hidden?

pauloboc:
I don't see the need to use the navigation bar, the browser is probably not the default. 
But feel free to refuse.

PR #127, opened 2020-03-30
Unmute & Mute

aliyss:


src/structures/Chat.js
```diff
@@ -57,6 +57,12 @@ class Chat extends Base {
          */
         this.archived = data.archive;
 
+        /**
+         * Indicates if the Chat is muted
+         * @type {boolean}
+         */
+        this.isMuted = data.mute.isMuted;
```
Aliyss: Not sure if this works. But in the browser `mute` is definitely an object.
I forgot if `data` is serialized or not.

src/structures/Chat.js
```diff
@@ -112,6 +118,21 @@ class Chat extends Base {
         return this.client.unarchiveChat(this.id._serialized);
     }
 
+    /**
+     * mutes this chat
+     * @param {Number} timestamp
```
pedroslopez: Can you describe what this time stamp refers to?

Aliyss: Timestamp is the exparation time

pedroslopez: Oh, can it really be any arbitrary date not in those intervals?

Also, maybe this should be a JS date and do the time stamp conversion internally for convenience.

Aliyss:
@pedroslopez

Mind testing it for me. I am not even sure it works correctly. And can't really test on my current system.

Aliyss:
After testing this does not work as intended. I will try to fix it.

Aliyss:
Deprecated

PR #132, opened 2020-04-05
added catch in sendMessage function for handling the error

dan89farhan:
added 
.catch(error => { throw error }); 
at line 389

please approve it for better handling of errors

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -388,7 +388,7 @@ class Client extends EventEmitter {
                 msg = await window.WWebJS.sendMessage(chat, message, options, sendSeen);
             }
             return msg.serialize();
-        }, chatId, content, internalOptions, sendSeen).catch(error => { throw error });
+        }, chatId, content, internalOptions, sendSeen).catch(error => { throw error; });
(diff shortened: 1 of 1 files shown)


pedroslopez:
Thanks for contributing. Is there a specific case where you are getting errors that I can try out?

dan89farhan:
yes, send a message with the wrong number. for e.g: 666666666677. since it is an invalid number then it should return an error message. but now the application gets crashed please update the code with catch functionality.

pedroslopez:
I see... but this fix doesn't seem to do anything since it's throwing the same error back. If this is something that happens in your case you should catch the error yourself when calling the sendMessage function, or check if the number is valid before by using `client.isRegisteredUser()`

dan89farhan:
how do I suppose to catch the error when client.js in not throwing any error it just get crashed with error UnhandledPromiseRejectionWarning: Error

Aliyss:
adding a ```try { your_code } catch (e) { console.log(e) }``` should work good enough.

PR #137, opened 2020-04-06
fix can't close the browser when it fails to connect to the internet

kurob1993:
move this.pupBrowser variable to line 51 to resolve not being able to close the browser when it is not connected to the internet

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -48,6 +48,8 @@ class Client extends EventEmitter {
         const page = (await browser.pages())[0];
         page.setUserAgent(UserAgent);
 
+        this.pupBrowser = browser;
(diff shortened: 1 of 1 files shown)


kurob1993:
`let ses = await session();
        const client = new Client({ puppeteer: { headless: false }, session: ses });
        client.initialize().then(function (res) {
            console.log(res.name);
        }).catch(function (err) {
            errRes = true;
            console.log(err.name);
            sails.sockets.broadcast('qrSockets', 'qr', { err: err.name });
            client.destroy();
        });`

image

pedroslopez:
Thanks for contributing, LGTM!

PR #146, opened 2020-04-08
Added MediaKey in order to see which sticker has been sent

InstantlyMoist:
This will give the option to see the MediaKey, every sticker has a MediaKey. This will open up a way to react to stickers in the future.


InstantlyMoist:
Forgot to mention, the MediaKey is consistent.

PR #154, opened 2020-04-11
Fix for inconsistent MediaKey.

InstantlyMoist:
As of my old PR, where I thought MediaKey would be consistent. It broke a sober 30 minutes ago.
Looking into the MediaKey and reverse engineering Whatsapp Web. It happens that the end-to-end encryption resets every session, thus changing the MediaKey. The ClientUrl, however, should be consistent at all times.


pedroslopez:
clientUrl is used to download media in general as well... what exactly are you trying to achieve here? Detect if a certain sticker is being referred to?

InstantlyMoist:
Trying to find a reference to the sticker used indeed. clientUrl was most likely to be correct in this case. There might be a better solution I'm missing, though.

PR #159, opened 2020-04-14
Fixed: No return if Contact not found

pranavms13:
No return if contact isnt found


pranavms13:
overrode by manual catch.
similar to https://github.com/pedroslopez/whatsapp-web.js/pull/132

PR #166, opened 2020-04-18
Pin xoxo unpin

aliyss:
chat.pin()
chat.unpin()

client.pinChat(chatId)
...


pedroslopez (changes requested):
Pinning/unpinning seems to work, but the return value isn't as expected (docstrings indicate it returns a boolean with the pin state). While testing I found pinning/unpinning returns a number that seems like it may be a timestamp, and I even got `undefined` at one point. Can you verify this?

pedroslopez (changes requested):
The return still doesn't seem to be working right. When I first pin a chat I get `false`, and unpinning gives me `true`, which would seem backwards. 

Calling `pin()` or `unpin()` more than once should also be limited or simply ignored. Doing this more than once seems to bug it out and it can get to a state where trying to unpin gives the message "Couldn't unpin chat", and just stays stuck until a restart of whatsapp web. Furthermore, doing this reverts the return values, causing `pin()` to return `true` and `unpin()` to return `false`.

Aliyss:
Ok. Will add a check later. So the state is checked before it tries to pin/unpin.

pedroslopez:
@Aliyss have you been able to take a look at this? Still seems like the pin state isn't too reliable

Aliyss:
@pedroslopez I didn't get to it. Very Sorry. Currently a lot to do.

src/Client.js
```diff
@@ -563,6 +563,40 @@ class Client extends EventEmitter {
...
+     * Enables and returns the pin state of the Chat
+     * @returns {boolean}
+     */
+    async pinChat(chatId) {
+        return await this.pupPage.evaluate(async chatId => {
+            let chat = await window.Store.Chat.get(chatId);
+            if (chat.pin) {
+                return true;
+            }
+            let maxPinned = await window.Store.Chat.models[2].pin;
```
pedroslopez, line 576: Could there be a better way to get the max number of pinned chats allowed / current number of pinned chats for this check?

This seems like it could be error prone if maybe there are less than 3 chats

pedroslopez (changes requested):

src/Client.js
Aliyss: Yeah... I searched quite a bit. I'll try again with a fresh mind. But couldn't find anything promising.

Aliyss:
@pedroslopez I couldn't find anything. I've been searching for quite a while, but there's no clear variable.

pedroslopez:
@Aliyss I think it's ok to assume it's 3 as long as we clearly indicate what that number means (e.g `const MAX_PIN_COUNT = 3` or something similar)

However, we still need to get the current number of pinned chats and compare it to prevent the issue.

src/Client.js
```diff
@@ -610,6 +610,43 @@ class Client extends EventEmitter {
         }, chatId);
     }
 
+    /**
+     * Enables and returns the pin state of the Chat
+     * @returns {boolean}
+     */
+    async pinChat(chatId) {
+        return await this.pupPage.evaluate(async chatId => {
```
pedroslopez, line 618:
```suggestion
        return this.pupPage.evaluate(async chatId => {
```

https://eslint.org/docs/rules/no-return-await I'm going to enable this rule soon

src/Client.js
```diff
@@ -610,6 +610,43 @@ class Client extends EventEmitter {
...
+            let chat = await window.Store.Chat.get(chatId);
+            if (chat.pin) {
+                return true;
+            }
+            const MAX_PIN_COUNT = 3;
+            if (window.Store.Chat.models.length > MAX_PIN_COUNT) {
+                let maxPinned = await window.Store.Chat.models[MAX_PIN_COUNT - 1].pin;
```
pedroslopez, line 625:
```suggestion
                let maxPinned = window.Store.Chat.models[MAX_PIN_COUNT - 1].pin;
```

No need for `await` here.

src/Client.js
```diff
@@ -610,6 +610,43 @@ class Client extends EventEmitter {
...
+        }, chatId);
+    }
+
+    /**
+     * Changes and returns the pin state of the Chat
+     * @returns {boolean}
+     */
+    async unpinChat(chatId) {
+        return await this.pupPage.evaluate(async chatId => {
+            let chat = await window.Store.Chat.get(chatId);
```
pedroslopez, line 641:
```suggestion
            let chat = window.Store.Chat.get(chatId);
```

No need for `await` here.

src/Client.js
```diff
@@ -610,6 +610,43 @@ class Client extends EventEmitter {
         }, chatId);
     }
 
+    /**
+     * Enables and returns the pin state of the Chat
+     * @returns {boolean}
+     */
+    async pinChat(chatId) {
+        return await this.pupPage.evaluate(async chatId => {
+            let chat = await window.Store.Chat.get(chatId);
```
pedroslopez, line 619:
```suggestion
            let chat = window.Store.Chat.get(chatId);
```

No need for `await` here. Getting a chat does not return a promise.

src/Client.js
```diff
@@ -610,6 +610,43 @@ class Client extends EventEmitter {
...
+            return true;
+        }, chatId);
+    }
+
+    /**
+     * Changes and returns the pin state of the Chat
+     * @returns {boolean}
+     */
+    async unpinChat(chatId) {
+        return await this.pupPage.evaluate(async chatId => {
```
pedroslopez, line 640:
```suggestion
        return this.pupPage.evaluate(async chatId => {
```

pedroslopez (changes requested):
Just a couple details with `await` usage

Aliyss:
Fixed and tested. Still works.

Aliyss:
Keeping it open... Until officially closed? All requested changes have been made.

pedroslopez (approved):

PR #168, opened 2020-04-19
Unmute v3

aliyss:


Aliyss:
So guess what... mute is added.
- It's a bit buggy for some unknown reason.
Using the command will mute it however it may not unmute on it's own.
In the Browser Console this works fine. However when using it with WWebJS it doesn't automatically unregister.

Aliyss:
@pedroslopez Can you replicate this. Maybe it's just my client.

Basically wondering why it work in the console but not directly.

PR #220, opened 2020-05-25
Open window

aliyss:


src/structures/Chat.js
```diff
@@ -177,6 +177,20 @@ class Chat extends Base {
...
+        return this.client.interface.openChatWindow(this.id._serialized);
+    }
+
+    /**
+     * Opens the Chat Drawer of the chat
+     */
+    async openDrawer() {
+        return this.client.interface.openChatDrawer(this.id._serialized);
+    }
+
```
pedroslopez, line 193: Everything else looks fine, but I would remove these to keep all interface-related things in the interface controller. 

```suggestion
```

pedroslopez (changes requested):

pedroslopez (commented):
LGTM
__
disregard this review, UI seems to be bugged

PR #252, opened 2020-06-09
feat: add typescript definition file

### LINKED ISSUE #125: Types in typescript?
Hello, I have a query, Do types available in typescript? Is there a way to install them as well as npm @types/whatsapp-web.js?

stefanfuchs:
Add typescript definitions.
Tested with npm pack locally, and using the compiled package in another project.

Resolves #125

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -660,4 +660,4 @@ declare namespace WAWebJS {
   }
 }
 
-export = WhatsappWeb
+export = WAWebJS
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ import puppeteer = require('puppeteer')
 declare namespace WAWebJS {
 
   export class Client extends EventEmitter {
-    constructor(options: WhatsappWeb.ClientOptions)
+    constructor(options: ClientOptions)
 
     /**Accepts an invitation to join a group */
     acceptInvite(inviteCode: string): Promise<void>
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,664 @@
+
+import { EventEmitter } from 'events'
+import puppeteer = require('puppeteer')
+
+declare namespace WhatsappWeb {
+
+  export class Client extends EventEmitter {
+    constructor(options: WhatsappWeb.ClientOptions)
+
+    /**Accepts an invitation to join a group */
+    acceptInvite(inviteCode: string): Promise<void>
+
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,664 @@
+
+import { EventEmitter } from 'events'
+import puppeteer = require('puppeteer')
+
+declare namespace WhatsappWeb {
```
pedroslopez, line 5:
```suggestion
declare namespace WAWebJS {
```

pedroslopez (changes requested):
This seems good to me, but maybe `WAWebJS` would be a better name instead of `WhatsappWeb`? Just to reflect the name of the library. It's a bit of a nitpick, but let me know :)

index.d.ts
stefanfuchs: Ok, no problem. This might be a better name. 
In the end there would be no practical difference, because we can import the lib (and the types) with a custom name. E.g.:

```typescript
import WhatsappWebCustom = require('whatsapp-web.js')
import { Client as CustomClient, MessageId } from 'whatsapp-web.js'
```

stefanfuchs: Wait, I forgot we need to also rename all the WhatsappWeb.* entries in the file.

index.d.ts
```diff
@@ -0,0 +1,663 @@
+
+import { EventEmitter } from 'events'
+import puppeteer = require('puppeteer')
+
+declare namespace WAWebJS {
+
+  export class Client extends EventEmitter {
+    constructor(options: WhatsappWeb.ClientOptions)
```
stefanfuchs, line 8:
```suggestion
    constructor(options: ClientOptions)
```

stefanfuchs (commented):
Rename WhatsappWeb.ClientOptions to ClientOptions

index.d.ts
```diff
@@ -0,0 +1,663 @@
...
+     */
+    _serialized: string,
+  }
+
+  export interface PrivateChat extends Chat {
+
+  }
+}
+
+export = WhatsappWeb
```
stefanfuchs, line 663:
```suggestion
export = WAWebJS
```

stefanfuchs:
I adjusted the file with the new namespace name. Everything seems ok now.

pedroslopez:
Great! LGTM, let's do this. Thanks for contributing.

PR #253, opened 2020-06-11
Added check for null in Message constructor

### LINKED ISSUE #250: Creating new instance of Message fails if author is null
In Version 1.6.1 I found a bug. It occurs when you want to create a new instance of Message and the author is null, it fails because it tests for type of data.author === "object" and null is typeof "object" but you can't access properties on null, so it fails

[screenshot of the code that produces the bug in Message.js](https://cdn.discordapp.com/attachments/698610475931271249/719488454446022676/unknown.png)
[My Hotfix for the issue](https://cdn.discordapp.com/attachments/698610475931271249/719488534284730459/unknown.png)

I just set the author to undefined when it's null.

mwalterskirchen:
As I described in my issue, that I opened a few days ago, you get an error when you want to create a new message when the author is null, which happens quite frequently. The problem lies in the typeof check because null is an object. I fixed the issue by adding an additional check for null in the constructor for all the fields that previously only checked for typeof === object. This should fix the errors.

Resolves #250

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -64,7 +64,7 @@ class Message extends Base {
          * ID for the Chat that this message was sent to, except if the message was sent by the current user.
          * @type {string}
          */
-        this.from = (typeof (data.from) === 'object' && data.author !== null) ? data.from._serialized : data.from;
+        this.from = (typeof (data.from) === 'object' && data.from !== null) ? data.from._serialized : data.from;
 
         /**
          * ID for who this message is for.
@@ -73,7 +73,7 @@ class Message extends Base {
          * If the message is sent by another user, it will be the ID for the current user. 
          * @type {string}
          */
-        this.to = (typeof (data.to) === 'object' && data.author !== null) ? data.to._serialized : data.to;
+        this.to = (typeof (data.to) === 'object' && data.to !== null) ? data.to._serialized : data.to;
 
         /**
          * If the message was sent to a group, this field will contain the user that sent the message.
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -64,7 +64,7 @@ class Message extends Base {
          * ID for the Chat that this message was sent to, except if the message was sent by the current user.
          * @type {string}
          */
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -64,7 +64,7 @@ class Message extends Base {
          * ID for the Chat that this message was sent to, except if the message was sent by the current user.
          * @type {string}
          */
-        this.from = typeof (data.from) === 'object' ? data.from._serialized : data.from;
+        this.from = (typeof (data.from) === 'object' && data.author !== null) ? data.from._serialized : data.from;
```
pedroslopez, line 67:
```suggestion
        this.from = (typeof (data.from) === 'object' && data.from !== null) ? data.from._serialized : data.from;
```

src/structures/Message.js
```diff
@@ -73,13 +73,13 @@ class Message extends Base {
          * If the message is sent by another user, it will be the ID for the current user. 
          * @type {string}
          */
-        this.to = typeof (data.to) === 'object' ? data.to._serialized : data.to;
+        this.to = (typeof (data.to) === 'object' && data.author !== null) ? data.to._serialized : data.to;
```
pedroslopez, line 76:
```suggestion
        this.to = (typeof (data.to) === 'object' && data.to !== null) ? data.to._serialized : data.to;
```

pedroslopez (changes requested):
This looks okay, but I believe you meant to check for nulls on these other properties instead of author. Can you check?

src/structures/Message.js
Alexcraftish: Yes, sorry, my bad. Good old copy and paste error. Should I change it and resubmit a new commit?

pedroslopez: No worries. You should be able to just make the changes and push a new commit to your branch and it should show up here

Alexcraftish:
This should fix the mistakes you pointed out

pedroslopez:
LGTM!

PR #271, opened 2020-06-28
Added Option to Custom User Agent in Client

iamyyjs:
I want the user agent to be different from the options available, so that I can personally equate the user agent with my personal computer which is linux.

Ignore if you feel unnecessary this PR.


pedroslopez (changes requested):
This looks good to me, but I would change `UserAgent` to be `userAgent` so it matches the camelCase convention

iamyyjs:
> This looks good to me, but I would change `UserAgent` to be `userAgent` so it matches the camelCase convention

thank you, i have updated to camelCase

pedroslopez (approved):

PR #278, opened 2020-07-02
fix: add more type definitions

stefanfuchs:
- add ClientInfo type definition
- fix some `@todo` comments in the docs


pedroslopez:
Thanks for this!

Could you make the relevant changes in the jsdoc comments?

stefanfuchs:
So, the main purpose of this PR was to add the ClientInfo in the .d.ts file, because I believe it to be an important part of the lib.
And using that oportunity, I fixed some other small problems in the docs too.

I have fixed the JSDoc comments for the easiest ones for now (those that I removed `@todo`).
- getWWebVersion
- unmuteChat
- isRegisteredUser

I have looked into the other `@todo` cases, but I didn't have time yet to fix them (also I don't have much experience with JSDoc).
But I may try fixing these cases when I have the time.

pedroslopez:
Sounds good! Thanks again for contributing.

PR #284, opened 2020-07-11
Implement option allowMsgFromMe

leodutra:
This allows the user to handle messages from himself, 
enabling users to create "personal assistant" chatbots.

This was tested and the effect is as expected.
Also, I ran the `generate-docs`.


pedroslopez:
Hi Leo, thanks for contributing!

If what you're looking for is a way to handle both received and sent messages, you can already do this by using the `message_create` event instead of the `message` event. https://pedroslopez.me/whatsapp-web.js/Client.html#event:message_create

This is what it seems like you're trying to achieve with this, please correct me if I'm wrong.

leodutra:
Let me try. I'll close if that suffice.

leodutra:
`message_create` is enough for my case. 
Thanks for the help @pedroslopez.

As the functionality already exists, I'm closing.

PR #285, opened 2020-07-13
fix: Constants export in index.js

stefanfuchs:
I came across an error while trying to use the enums present in index.d.ts; Example:
```typescript
import WhatsappWeb = require('whatsapp-web.js')
if (messageType === WhatsappWeb.MessageType.TEXT) // Error: cannot read property 'TEXT' of undefined
```

I found out that even though MessageType is defined in index.d.ts, it should also be exported in index.js, in order for it to be present at runtime.
This PR exports all the Constants in index.js and fixes the issue.

Note: I used the spread operator to export all the constants at once. If you think it should be better to export them individually, please let me know and I can try doing that also.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -298,6 +298,12 @@ declare namespace WAWebJS {
 
   }
 
+  /** whatsapp web url */
+  export const WhatsWebURL: string
+  
+  /** default client options */
+  export const DefaultOptions: ClientOptions
+
   /** Chat types */
   export enum ChatTypes {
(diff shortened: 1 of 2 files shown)


stefanfuchs:
Note2: I also added some missing contants that were in the Constants.js but not index.d,ts:
- WhatsWebURL: string
- DefaultOptions: ClientOptions

pedroslopez:
LGTM

Thank you :)

PR #286, opened 2020-07-13
chore: add missing types in JSDoc/TSDoc (ClientOptions, ClientInfo, MessageSendOptions, Session)

stefanfuchs:
Added missing types in the JSDoc and in the TSDoc files.
- ClientOptions
- ClientInfo
- MessageSendOptions
- Session

Also remove @todo from createGroup (was already solved).

Note: I added the 'info' property in Client, but it doesn't seem to be finding the ClientInfo type, so it is being set as 'any' in the generated docs. If you know how to solve this, please let me know.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -500,6 +500,8 @@ declare namespace WAWebJS {
     mentions?: Contact[]
     /** Send 'seen' status */
     sendSeen?: boolean
+    /** Media to be sent */
+    media?: MessageMedia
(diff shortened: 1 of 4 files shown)


pedroslopez:
I made some changes to properly set type on clientinfo as well as reusing the MessageSendOptions when sending messages through the chat and replying to messages directly.

Thanks so much for this!

PR #301, opened 2020-08-05
fix: MessageMedia interface to class (index.d.ts)

stefanfuchs:
- Change MessageMedia from an interface to a class, to correctly correspond to the underlying type
- Add documentation comments for MessageMedia


pedroslopez:
LGTM!

PR #304, opened 2020-08-08
fix: add null checks for Message, Chat, Contact

stefanfuchs:
Add null checks, to avoid errors like 'cannot read property xxxxx of undefined'.

I was getting this kind of error while trying to use `getChat(chatId)` but the chat did not exist.
Now it returns null instead of throwing an error.


pedroslopez:
Thanks for this. To be honest, I've been a little bit reluctant to merge this because it may introduce a breaking change if people are using something like a try-catch to handle invalid chats/contacts. Returning `null` won't produce an error, which could cause issues for them. These breaking changes should be avoided and only done if necessary on major releases, following [semantic versioning](https://semver.org/) rules. 

Some of these changes are fine, like the check on sending media or the one on deleting messages- I'm more concerned with `getChat()` and `getContact()`. Maybe in this case, to make it clearer, we should return our own error. This is something I've kind of wanted to do for a while, which is define our own error type and throw this ourselves so that users of the library can catch these and handle them properly. 

Thoughts?

src/util/Injected.js
```diff
@@ -185,7 +185,7 @@ exports.LoadUtils = () => {
         const chat = window.Store.Chat.get(chatId);
         if (!chat) return null;
 
-        return window.WWebJS.getChatModel(chat);
+        return await window.WWebJS.getChatModel(chat);
```
stefanfuchs, line 188: why `return await` here?

stefanfuchs:
@pedroslopez what you said makes sense, I agree that this is potentially a breaking change for those who already use try catch in their codes.
The reason I made this PR is because I believe it makes more sense to simply return null when a chat/contact does not exist, instead of throwing an error (which for me appears to be an unexpected behaviour).

stefanfuchs:
Think about a database. When you query something by id X, but there is no such object with id X, an undefined or null value is returned.
An error is only thrown when an actual error occurs, like connection problems or internal database error.

src/util/Injected.js
stefanfuchs: some insight: https://eslint.org/docs/rules/no-return-await

pedroslopez: Yep, you're right. Adding `async` alone was enough here. This was changed on another commit- I'm just merging changes back, so I'm resolving this here. I'll do some cleanup soon when I do a slight refactor I'm planning of a couple things.

pedroslopez:
Yeah I totally understand your reasoning, and I agree with you. What bothers me the most is the breaking change, and I wouldn't want to release v2 now just for this one change.

I think we should instead opt for deprecating the current behavior, but providing this new `null` return as an option. I've seen something similar with Mongoose, where you will get deprecation warnings when using it the old way, but you can set an option when connecting to opt-in to the new functionality. Examples here: https://mongoosejs.com/docs/deprecations.html

This way we can safely include this without causing a breaking change on a minor/patch release and just remove the deprecated functionality when v2 is released.

stefanfuchs:
Deprecating sounds like a good alternative. Then removing it in the next major release.

PR #312, opened 2020-08-13
Events and constants to emit pairing situation

guicuton:
Additional files that i've miss on first PR with new event and constants to check if client is paired or not

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ typings/
 
 # next.js build output
 .next
+
+# client files
+client/
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ typings/
 
 # next.js build output
 .next
+
+#ignore client folder
+client/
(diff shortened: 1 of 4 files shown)


pedroslopez:
Hi Guilherme. I'm not entirely sure what's going on here. It seems like your proposed changes are spread throughout two separate PRs. Can you consolidate them into a single pull request so It can be more easily reviewed and merged as a unit? 

On another note, does the `ready` event that is already being emitted not address your use case? Can you explain a bit more why you may need this event?

guicuton:
> Hi Guilherme. I'm not entirely sure what's going on here. It seems like your proposed changes are spread throughout two separate PRs. Can you consolidate them into a single pull request so It can be more easily reviewed and merged as a unit?

My fail... i'm new with github on ATOM and then i've made some sh***

> On another note, does the `ready` event that is already being emitted not address your use case? Can you explain a bit more why you may need this event?

The `ready` event is emitted only when the user is paired and with this new event a new message will be emitted with current pairing status, differently of getState() that is needed to be called

pedroslopez:
Hmm... I'm only seeing you added a function `getSerializedPhoneId`, which seems unrelated. The changes to the `.gitignore` also seem to be out of place, and contain what seems like merge stuff from git. 

Could you check and test this PR and make sure the correct changes have been committed? I would also appreciate if you could keep whitespace changes to a minimum to make reviewing easier, as well as addressing any linting issues.

guicuton:
Bug changes. Ignore

PR #323, opened 2020-08-29
Add limitTimestamp option to Chat.js fetchMessages()

wemersonrv:
This PR add `limitTimestamp` filter option to `fetchMessages` method on chats;

The argument is optional, and if ommited, filter will skip it and just use `searchOptions.limit`

### DIFF
--- a/src/structures/Chat.js
+++ b/src/structures/Chat.js
@@ -131,14 +131,19 @@ class Chat extends Base {
      * Loads chat messages, sorted from earliest to latest.
      * @param {Object} searchOptions Options for searching messages. Right now only limit is supported.
      * @param {Number} [searchOptions.limit=50] The amount of messages to return. Note that the actual number of returned messages may be smaller if there aren't enough messages in the conversation. Set this to Infinity to load all messages.
+     * @param {Number} [searchOptions.limitTimestamp] Unix timestamp to limit results until a specific time period. If ommited, only searchOptions.limit will be used.
      * @returns {Promise<Array<Message>>}
      */
     async fetchMessages(searchOptions) {
-        if (!searchOptions || !searchOptions.limit) {
-            searchOptions = { limit: 50 };
-        }
-        let messages = await this.client.pupPage.evaluate(async (chatId, limit) => {
-            const msgFilter = m => !m.isNotification; // dont include notification messages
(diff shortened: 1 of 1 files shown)


pedroslopez (changes requested):
Thanks for contributing! Unfortunately this doesn't seem to be working as expected. 

Consider the following scenarios:

**First scenario**
1. We are working with an old Chat that has thousands of messages.
2. It is currently Aug 28 at 10 PM
3. We want to get messages sent in the last 24 hours (From Aug 27 at 10PM)
4. Let's say there are only 10 messages in this time period.
5. We execute `chat.fetchMessages({limitTimestamp: XXXXX})` with the corresponding timestamp
6. The function takes forever to run, loading all messages in the chat until it was created.

This occurs because it keeps trying to give you the default `limit` of messages, which is 50. Since it can't find 50 messages that match the condition, it keeps trying and trying, loading the entire conversation, until it runs out of messages. The same issue will occur if we give it any `limit` bigger than the number of messages available.

**Second scenario**
Let's take the same example as before, but this time we'll modify point # 4 to say there are 100 messages in this time period, and we stay with the default limit of 50 messages. In this case, because of the way `limit` works, it will return the _last_ 50 messages in the chat instead of the first 50 messages starting with the specific date, as would be expected.

**Other suggestions**
1. Maybe use `afterTimestamp` instead of `limitTimestamp`, to be more clear with the fact that it will return messages _after_ that specific timestamp.
2. I see you're turning the timestamp into a JS Date object for validation, but consider that Unix timestamps are expressed in _seconds_ while the parameter for the JS Date object is expressed in _milliseconds_. If you wanted to convert unix timestamp to js timestamp, you'll need to divide by 1000 (`unixTimestamp / 1000`) to create the proper date.

wemersonrv:
> **First scenario**
> This occurs because it keeps trying to give you the default `limit` of messages, which is 50. Since it can't find 50 messages that match the condition, it keeps trying and trying, loading the entire conversation, until it runs out of messages. The same issue will occur if we give it any `limit` bigger than the number of messages available.

Yeah, i see now after your explanation. Here in my tests i used `limit` with **Infinity** value and it takes a long time, more than 40 seconds when it's good. Now i know that even when i pass argument to filter timestamps after limitTimestamp, it still get all data before filter.

PR #324, opened 2020-08-29
feat: added typing on forward messages

caioagiani:


pedroslopez:
Thanks for contributing.

This isn't correct... the forward function does not return the created message.

caioagiani:
Fixed @pedroslopez

PR #351, opened 2020-09-22
Chat mute expiration

Leoalmeida95:
The lib provides the mute () and unmute () methods, but there is no property in Chat that indicates whether the conversation is already mutated. I took a look and this property is just for that.

### DIFF
--- a/src/structures/Chat.js
+++ b/src/structures/Chat.js
@@ -58,7 +58,7 @@ class Chat extends Base {
         this.archived = data.archive;
 
         /**
-         * Timestamp for when the mute expiration
+         * Unix timestamp for when the mute expires
          * @type {number}
          */
         this.muteExpiration = data.muteExpiration;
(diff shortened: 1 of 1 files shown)


Leoalmeida95:
Hello @pedroslopez, this is a simple change. Do you think it is possible to add it to the library? It is important, because without this property, you cannot know when a chat is mutated. The mute () and unmute () methods work perfectly, but today there is no way to distinguish between mutated and non-mutated chats.

pedroslopez:
LGTM! Thanks for contributing :)

pedroslopez:
FYI, I've also added similar functionality in ee21e145472c53f7ba381a3705ebf3 that's just a boolean that indicates whether or not the chat is muted

Leoalmeida95:
Perfect! So we can know if it is mutated and its duration. very good! I will try to contribute more :)

PR #373, opened 2020-10-11
Account information added to login

oguzhan51511:
After everything is ready, I'm on the slope of transferring the information of the account. Thus, I don't have to send the account number as a parameter in multiple use.


pedroslopez:
Hi, thanks for contributing.

Is there a reason why you couldn't get the info from `client.info` and use the `authenticated` event for session details?

oguzhan51511:
There is only session information in the authenticated event.
I could not access client.info data. What I need is the number of the account logged in. If there is any place I can reach, it works.

PR #378, opened 2020-10-13
feat: Wpp Business's Labels support

dyegoaurelio:
We're now able to work with Wpp Business's Labels

I added methods to get all labels, get all labels a chat is marked and to fetch all chats a Label is marking

### DIFF
--- a/src/util/Injected.js
+++ b/src/util/Injected.js
@@ -204,6 +204,8 @@ exports.LoadUtils = () => {
 
     window.WWebJS.getMessageModel = message => {
         const msg = message.serialize();
+        msg.isStatusV3 = message.isStatusV3;
+
         delete msg.pendingAckUpdate;
         return msg;
     };
--- a/index.d.ts
+++ b/index.d.ts
@@ -444,6 +444,8 @@ declare namespace WAWebJS {
     body: string,
     /** Indicates if the message was a broadcast */
     broadcast: boolean,
+    /** Indicates if the message was a status update */
+    isStatus: boolean,
     /** ID for the Chat that this message was sent to, except if the message was sent by the current user */
     from: string,
     /** Indicates if the message was sent by the current user */
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -88,6 +88,12 @@ class Message extends Base {
         this.isForwarded = data.isForwarded;
 
         /**
+         * Indicates if the message is a status update
+         * @type {boolean}
+         */
+        this.isStatus = data.isStatusV3;
+
+        /**
          * Indicates if the message was a broadcast
          * @type {boolean}
          */


src/Client.js
```diff
@@ -741,6 +741,41 @@ class Client extends EventEmitter {
         return { gid: createRes.gid, missingParticipants };
     }
 
+    /**
+     * Returns all current Labels
+     * @returns {Promise<Array<Label>>}
+     */
+    async getLabels() {
+        return JSON.parse( await this.pupPage.evaluate(async () => {
```
pedroslopez, line 749: No need to encode/decode JSON here. Puppeteer takes care of serializing the models.
Also, no need to `return await`. Opt for `return` instead. The https://eslint.org/docs/rules/no-return-await rule will be enabled on eslint soon.

src/Client.js
```diff
@@ -741,6 +741,41 @@ class Client extends EventEmitter {
...
+            return JSON.stringify(await window.WWebJS.getLabels());
+        }) );
+    }
+
+    /**
+     * Returns Label of specified chat 
+     * @returns {Promise<Label>}
+     */
+    async getChatLabels(chatId){
+        return JSON.parse( await this.pupPage.evaluate(async (chatId) => {
```
pedroslopez, line 759: Same as before, JSON conversion and no `return await`.

src/Client.js
```diff
@@ -741,6 +741,41 @@ class Client extends EventEmitter {
...
+            return JSON.stringify(await window.WWebJS.getChatLabels(chatId));
+        }, chatId) );
+    }
+
+    /**
+     * Returns all Chats of specified label
+     * @returns {Array<ChatId>}
+     */
+    async getAllChatsFromLabel(labelId){
+        return ( await this.pupPage.evaluate(async (labelId) => {
```
pedroslopez, line 769: Unneeded `return await`

src/structures/Chat.js
```diff
@@ -228,6 +228,14 @@ class Chat extends Base {
     async getContact() {
         return await this.client.getContactById(this.id._serialized);
     }
+
+    /**
+     * Returns array of all labels this Chat is marked
+     * @returns {Array<Labels>}
+     */
+    async getLabels() {
+        return await this.client.getChatLabels(this.id._serialized);
```
pedroslopez, line 237: Unneeded `return await`

src/util/Injected.js
```diff
@@ -317,6 +317,21 @@ exports.LoadUtils = () => {
         return true;
     };
 
+    window.WWebJS.getLabels = async () =>{
+        return window.Store.Label.models.map( function (label){  return  { 'name': label.name , 'hexColor': label.hexColor , 'id':label.id }; });
```
pedroslopez, line 321: Prefer arrow functions here to keep consistent with the rest of the code.

src/util/Injected.js
```diff
@@ -317,6 +317,21 @@ exports.LoadUtils = () => {
...
+    };
+
+    window.WWebJS.getChatLabels = async (chatId) => {
+        let chatLabelsIndex = await window.WWebJS.getChat(chatId);
+        chatLabelsIndex = chatLabelsIndex.labels;
+        //console.log(chatLabelsIndex)
```
pedroslopez, line 327: Remove commented code

src/util/Injected.js
```diff
@@ -317,6 +317,21 @@ exports.LoadUtils = () => {
...
+        let res = [];
+        for (var i in chatLabelsIndex){
+            res.push({ 'name': window.Store.Label.get(chatLabelsIndex[i]).name , 'hexColor': window.Store.Label.get(chatLabelsIndex[i]).hexColor ,  'id' : chatLabelsIndex[i] } );
```
pedroslopez, line 330: I'm seeing in multiple places you're constructing this object representation of a Label with the same properties. To keep things consistent and remove code duplication, perhaps we could instead have a serializer function, much like we have with `getChatModel` and `getMessageModel` that takes care of returning a properly serialized representation of the label with the necessary properties.

pedroslopez (changes requested):
Thanks for contributing! Sorry it took a while to review.

Looks good overall and works functionality-wise. However, I believe instead of returning regular objects for labels in functions like `client.getChatLabels()` and `client.getLabels()`, we should opt for creating a Class that represents a ChatLabel. This way, we have a structured representation of labels and their properties that can be properly documented. It could also be useful if we want to add functions like editing labels in the future.

dyegoaurelio:
Closed this pr accidentally
reopened at #407

PR #384, opened 2020-10-16
Added hashkey for showing encrypted media

fdciabdul:
this will be importan for adding features in the future


fdciabdul:
@pedroslopez is wrong or nah ?

pedroslopez:
@fdciabdul Is this useful for anything right now? If there's a valid use case that can be accomplished at the moment introducing this functionality, I'm ok with merging it. I would just like to refrain from adding partial features.

PR #388, opened 2020-10-21
Get screenshot

dogukanberat:


dogukanberat:
u can get qr like this with screenshot

```
    wbot.on("qr", async qr => {
      await wbot.screenShot(
        `public/screen.png`
      );
}
```

parthibd:
Why do you need a screenshot ? The QR code data is already there is that event callback . You can just convert the string into a simple QR code image file . Easy.

pedroslopez:
I think, since taking screenshots isn't really a library function and isn't something I'd like to promote, it's ok to have people use it through the puppeteer page if they really want to use it. In most cases there shouldn't really be a need to take screenshots.

Issue #398, opened 2020-10-26
Receive multi_vcard


lucasrosa90:
### Bug description
When I'm receiving multi_vcard, the body is empty.

### Reproduction steps
1. Receive multiple vcards message..
...

### Expected behavior 
Show multiple VCARDs, in array or something like that.

### Environment (please complete the following information):
**WhatsApp**
 - Account type [Standard / Business]: Business
 - Device OS [iOS / Android]: Both
 - WhatsApp Web version [run `await client.getWWebVersion()`]: 2.2043.8

**Library**
 - Browser [Chrome / Chromium]: Chromium
 - NodeJS version (`node -v`): v14.8.0
 - npm or yarn version (`npm -v`): yarn 1.22.4
 - whatsapp-web.js version: 1.10.0

fdciabdul:
can you send the log of error ?

LucasRosa90:
@fdciabdul there's no error.

I put a console.log on message_create, and got this:

`Message {
  mediaKey: undefined,
  id: {
    fromMe: false,
    remote: '554891581407@c.us',
    id: '3ABE9DADC34E8FA5AC44',
    _serialized: 'false_554891581407@c.us_3ABE9DADC34E8FA5AC44'
  },
  ack: -1,
  hasMedia: false,
  body: '',
  type: 'multi_vcard',
  timestamp: 1603723095,
  from: '554891581407@c.us',
  to: '554891049495@c.us',
  author: undefined,
  isForwarded: false,
  broadcast: false,
  fromMe: false,
  hasQuotedMsg: false,
  location: undefined,
  mentionedIds: []
}`

LucasRosa90:
I'm trying to find the error.

This is the object received at `onAddMessageEvent`

```
{
  id: {
    fromMe: false,
    remote: '554891581407@c.us',
    id: '3A2805D7A139A7852747',
    _serialized: 'false_554891581407@c.us_3A2805D7A139A7852747'
  },
  type: 'multi_vcard',
  t: 1603738236,
  notifyName: '',
  from: '554891581407@c.us',
  to: '554891049495@c.us',
  self: 'in',
  ack: -1,
  invis: false,
  isNewMsg: true,
  star: false,
  recvFresh: true,
  broadcast: false,
  mentionedJidList: [],
  isVcardOverMmsDocument: false,
  vcardList: [
    {
      displayName: 'Adriana Agemed',
      vcard: 'BEGIN:VCARD\n' +
[...]

PR #400, opened 2020-10-26
feat: add received vcards to the message

lucasrosa90:
#398


pedroslopez (changes requested):
I think the issue with this solution is that the `body` property would no longer always be a `string` type, which could cause issues for people using the library if they're expecting this type.

Maybe we should just add a property that has the vcard array, like what WhatsApp is doing on their end.

LucasRosa90:
Sure, you're right.

pedroslopez (approved):

PR #402, opened 2020-10-28
fix: serialize msgs.buttons and delete msgs from chat model

yuvalfis:
@pedroslopez here's the changes


pedroslopez:
Thanks for this!

PR #404, opened 2020-10-28
documentation for block() and unblock() added

iamrenaud:


pedroslopez:
Documentation is automatically generated when the version with the changes is released. A version of the package with Block/unblock functionality has not been released on npm yet, so docs have not been regenerated. 

Thanks for this though.

PR #411, opened 2020-11-04
Added option to return qrcode as a base64 image.

victormga:


victormga:
Lint failed, Making new pr

PR #412, opened 2020-11-04
Added option to return qrcode as a base64 image.

victormga:


victormga:
anything on this?

fadlee:
Love this changes, no need to regenerate the qr code

parthibd:
We can easily encode the QR code data as an image for the event data in WWJs . I don't think this feature is necessary as it can be implemented easily in userland code .

victormga:
closing as i'm making the pr's separated.

PR #421, opened 2020-11-07
feat: Get the registered chat id in whatsapp

lucasrosa90:
Get the registered chat id in whatsapp.

**Reason?**
- In Brazil, after change the rule for cell numbers, WhatsApp didn't change they database for the already registered numbers, just to the new ones, so here we have to get what's the chatId registered in WhatsApp before send the messages.

I'll show it in codes:

Using my phone number, the actual `client.isRegisteredUser()` return true for the valid and the invalid chatId, you guys can check there:
```
> npm run shell
wwebjs> client.isRegisteredUser("5548991581407@c.us").then(result => { console.log(`isRegistered: ${result}`); });
Promise { <pending> }
wwebjs> isRegistered: true
wwebjs> client.isRegisteredUser("554891581407@c.us").then(result => { console.log(`isRegistered: ${result}`); });
Promise { <pending> }
wwebjs> isRegistered: true
```

The numbers are different, the 5548**9**91581407 is the new rule, but in WhatsApp it doesn't exists, so I have to send the message to the old one, without this additional 9.

So, to keep the compability and resolve this problem, with some help from @Guicuton I created this function at client.js:

```
> npm run shell
wwebjs> client.getRegisteredChatId("5548991581407@c.us").then(result => { console.log(`Registered chatId: ${result}`); });
Promise { <pending> }
wwebjs> Registered chatId: 554891581407@c.us
```

I dont know if other countries has this problem, but in Brazil we have, and it prevent from future problems to another countries too.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -65,6 +65,9 @@ declare namespace WAWebJS {
         /** Check if a given ID is registered in whatsapp */
         isRegisteredUser(contactId: string): Promise<boolean>
 
+        /** Get the registered chat id in whatsapp */
+        getRegisteredChatId(chatId: string): Promise<string>
+
         /**
          * Mutes the Chat until a specified date
          * @param chatId ID of the chat that will be muted
--- a/index.d.ts
+++ b/index.d.ts
@@ -65,8 +65,8 @@ declare namespace WAWebJS {
         /** Check if a given ID is registered in whatsapp */
         isRegisteredUser(contactId: string): Promise<boolean>
 
-        /** Get the registered chat id in whatsapp */
-        getRegisteredChatId(chatId: string): Promise<string>
+        /** Get the registered number id in whatsapp */
+        getNumberId(chatId: string): Promise<object>
 
         /**
          * Mutes the Chat until a specified date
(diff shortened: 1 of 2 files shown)


victormga:
lucas, do you have any additional info about if this call causes ban?
in a older version of my code, i had an check like this when importing contacts from a csv file, and i've had a banned wa account.
only happened once, and it was a new whatsapp account, but that made me worried.

anyway, i think this is duplicate because you can call: await window.WWebJS.getNumberId(id); and it returns the real wa id, even here in brazil where the phone numbers had changed.

lucasrosa90:
@victormga how can i get window.WWebJS from "new Client()"?

lucasrosa90:
I did some changes to use that function, but I didnt find another way to export that to use in my client. If there's another way to use that, tell me.

victormga:
lucas, you can evaluate on client.pupPage.

```
try {
  const whatsapp_id = client.pupPage.evaluate((number) => {
     return await window.WWebJS.getNumberId(number);
  }, number);
  console.log(whatsapp_id);
} catch(e) {
  console.error(e);
}
```

lucasrosa90:
Thanks @victormga , I didnt know that I can do that.

PR #439, opened 2020-11-19
Added option to return qrcode as a base64

victormga:


victormga:
anything?

pedroslopez:
Thanks for contributing. I think transforming the QR code data to an image or whatever format the user needs is a task that is out of scope for the library and should be handled by the program using the library.

PR #456, opened 2020-11-25
Add feature to fetch messages from a chat object, filtering after a specific timestamp

wemersonrv:
Created a function `fetchMessagesAfter` based on original function `fetchMessages`.

It has a Unix timestamp argument to be the limit, and will get all messages after this timestamp.

A couple of months ago, i've made another PR (#323) that not works fine. This is my new attempt to make it right... done some tests here and works fine!


pedroslopez:
Thanks for contributing!

I haven't tested yet, but would it be possible to integrate these changes into the existing `fetchMessages` function? Initially my goal was to have a single function to do this and just pass in parameters. In this case, maybe the option would be `fetchMessages({ after: timestamp})`.  It should keep support for `limit` option as well.

wemersonrv:
> Thanks for contributing!
> 
> I haven't tested yet, but would it be possible to integrate these changes into the existing `fetchMessages` function? Initially my goal was to have a single function to do this and just pass in parameters. In this case, maybe the option would be `fetchMessages({ after: timestamp})`. It should keep support for `limit` option as well.

Understand. I will wait your tests first and if u tell that it's fine i will merge both in same function... this way is good for u?

wemersonrv:
@pedroslopez make the update you ask and add `searchOptions.timestamp` and will accept both, one or none arguments...

- If do not provide arguments, will use only `searchOptions.limit = 50` as it was before, and ignore timestamp;
- If provide only `searchOptions.timestamp` argument, limit will not be used;
- If provide both, timestamp and limit; will use both to filter.

src/structures/Chat.js
```diff
@@ -165,29 +165,43 @@ class Chat extends Base {
...
-            while (msgs.length < limit) {
+            let filteredMessages = [];
+            do{
                 const loadedMessages = await chat.loadEarlierMsgs();
                 if (!loadedMessages) break;
-                msgs = [...loadedMessages.filter(msgFilter), ...msgs];
-            }
+
+                filteredMessages = loadedMessages.filter(msgFilter);
+                msgs = [...msgs, ...filteredMessages];
```
pedroslopez, line 195: Why was this order reversed?

src/structures/Chat.js
```diff
@@ -165,29 +165,43 @@ class Chat extends Base {
...
      */
     async fetchMessages(searchOptions) {
-        if (!searchOptions || !searchOptions.limit) {
-            searchOptions = { limit: 50 };
+        let {limit, timestamp} = searchOptions || {};
+
+        if(timestamp < 1) timestamp = null;
+        if(limit === undefined || limit === null) limit = (timestamp ? limit : 50);
+        if(timestamp){
+            const date = new Date(timestamp * 1000);
```
pedroslopez, line 177: Maybe passing in a js `Date` object instead of the unix timestamp would be easier for end users? What do you think?

pedroslopez (changes requested):
I finally had the chance to test this, and it seems to work well overall. However, I would say there might be some confusion with how `timestamp` and `limit` interact when used together.

Going into the testing, I initially expected the timestamp to work as a starting point from which the limit would be enacted. As it's implemented right now, it still starts counting back from the latest message in the conversation. Maybe it's a matter of choosing a better name for the `timestamp` parameter that better indicates what it's intended for.

In my previous comment on this issue I suggested the option name to be `{after: xxx}` which I think would clearly indicate that it would start counting from the time specified. I'm not sure what the right name would be for this other method.

Maybe changing the direction it's counting in could be an option as well? I'm not sure. What do you think?

mpirescarvalho:
> I finally had the chance to test this, and it seems to work well overall. However, I would say there might be some confusion with how `timestamp` and `limit` interact when used together.
> 
> Going into the testing, I initially expected the timestamp to work as a starting point from which the limit would be enacted. As it's implemented right now, it still starts counting back from the latest message in the conversation. Maybe it's a matter of choosing a better name for the `timestamp` parameter that better indicates what it's intended for.
> 
> In my previous comment on this issue I suggested the option name to be `{after: xxx}` which I think would clearly indicate that it would start counting from the time specified. I'm not sure what the right name would be for this other method.
> 
> Maybe changing the direction it's counting in could be an option as well? I'm not sure. What do you think?

What about two options: `{after: xxx, before: xxx}`. We could use both, one or none of them. `limit` would still limit the quantity of messages. `before` could defaults to now and after could defaults to undefined, meaning all messages would be fetched, respecting limit

wemersonrv:
Hello guys.

I took time out for myself. I spent some time away to take a vacation and also to recover mentally and emotionally after the recent loss of my dear father.

I hope you forgive me for not having recently interacted here but now I'm back and I will answer everyone.

src/structures/Chat.js
wemersonrv: Hello @pedroslopez, the idea is that the most recent messages are being taken up to a limit, it made sense for me to order them in a descending order.

src/structures/Chat.js
wemersonrv: > Maybe passing in a js `Date` object instead of the unix timestamp would be easier for end users? What do you think?

For me this is not a problem, as long as it's valid. Whatever you find most interesting, it's fine with me!a

pedroslopez:
@wemersonrv So sorry for your loss. No worries at all and take the time you need.

wemersonrv:
> @wemersonrv So sorry for your loss. No worries at all and take the time you need.

Thanks pedro. It's fine now... life goes on!

wemersonrv:
> > I finally had the chance to test this, and it seems to work well overall. However, I would say there might be some confusion with how `timestamp` and `limit` interact when used together.
> > Going into the testing, I initially expected the timestamp to work as a starting point from which the limit would be enacted. As it's implemented right now, it still starts counting back from the latest message in the conversation. Maybe it's a matter of choosing a better name for the `timestamp` parameter that better indicates what it's intended for.
> > In my previous comment on this issue I suggested the option name to be `{after: xxx}` which I think would clearly indicate that it would start counting from the time specified. I'm not sure what the right name would be for this other method.
> > Maybe changing the direction it's counting in could be an option as well? I'm not sure. What do you think?
> 
> What about two options: `{after: xxx, before: xxx}`. We could pass both, one or none of them. `limit` would still limit the quantity of messages. `before` could defaults to now and after could defaults to undefined, meaning all messages would be fetched, respecting limit

I think it is a good idea.

Let's review the points afther our discussion:

- Change optiopn **timestamp** to **_after_**. 
- Add option **_before_**.
- Use **js Date** object instead of the unix **timestamp**. @pedroslopez can you tell us what is better, or if it is just a preference?
- Result reverse order, @pedroslopez you make a point about it. But, In my case, i need the messages in reverse order, and don't think is a good idea to reorder the result after function returns, it's redundant. So, how about add an **_order_**(DESC/ASC)  argument too?

mpirescarvalho:
@wemersonrv I'm sorry for your loss

About your last point (reverse order): IMHO, the first position in the array (`msgs[0]`) should be the last message sent to the chat. How is it working today?

wemersonrv:
> @wemersonrv I'm sorry for your loss
> 
> About your last point (reverse order): IMHO, the first position in the array (`msgs[0]`) should be the last message sent to the chat. How is it working today?

In the current oficial `fetchMessages` funcion, it returns an array ordered by timestamp in ASCENDING ORDER... so the `msgs[0]` has the oldest message on the array... The opposite that i need. So my suggestion is to add an option to choose the order.

esteves67:
> > @wemersonrv I'm sorry for your loss
> > About your last point (reverse order): IMHO, the first position in the array (`msgs[0]`) should be the last message sent to the chat. How is it working today?
> 
> In the current oficial `fetchMessages` funcion, it returns an array ordered by timestamp in ASCENDING ORDER... so the `msgs[0]` has the oldest message on the array... The opposite that i need. So my suggestion is to add an option to choose the order.

The option to reverse order would be much appreciated! 
Because i think most of the times we want to get the last message instead of the earliest.
 
Example if a chat has 50 messages saying "ok", being the first at 1pm and the last at 3pm,
lets say you want to get the last 'ok' (at 3pm),
if you put `const message = (await chat.fetchMessages({ limit: 50 })).find(msg => msg.body == 'ok')` you will get the first ok (at 1pm) not the last (at 3pm).
The approach i did was to reverse the array by using `const reversed = message.reverse();`.

esteves67:
> @wemersonrv I'm sorry for your loss
> 
> About your last point (reverse order): IMHO, the first position in the array (`msgs[0]`) should be the last message sent to the chat. How is it working today?

This msgs[0] to get the first position doesn't always work :cry:

PR #459, opened 2020-11-28
feat: add new typing in chat

caioagiani:
Added typing of group participants in Chat for projects in typescript.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -700,7 +700,11 @@ declare namespace WAWebJS {
         timestamp: number,
         /** Amount of messages unread */
         unreadCount: number,
-
+        /** Indicates the id of all participating users */
+        participants: [{
+            id: ContactId
+        }],
+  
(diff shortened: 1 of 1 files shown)


pedroslopez:
Thanks! It seems like we missed this one.

Could you add all the available properties, not only the id? It seems to be just two booleans: `isAdmin` and `isSuperAdmin`.

pedroslopez (changes requested):
Thanks! It seems like we missed this one.

It looks like we're missing typings for the GroupChat model- as it is now, you would be adding this property to all chats (including private chats), which would not be correct.

Could you add the corresponding GroupChat typings and put the participants property in the right place?

Also, it seems like you're missing two properties from the participants array:  `isAdmin` (bool) and `isSuperAdmin` (bool).

pedroslopez:
Superseded by #485

PR #466, opened 2020-12-02
Get all whatsapp group names

PetrosM263:
Hi guys.

Is it possible to get all group names that i have joined on my whatsapp so that i can send a message to bulk groups at once. I think it will e great if we can a script that compile the group names in a json format way like.

{
whatsapphroups: {
   {
     id: 173860,
     name: Group 1
  },
{
     id: 173862,
     name: Group 2
  },
{
     id: 173863,
     name: Group 3
  },
}
}

if we have something like this we can therefore use foreach function to send messages to all groups available

Thank you


pedroslopez:
This is not a pull request, but rather a question on usage. Please join our discord server and ask there if you're still having issues.

PR #469, opened 2020-12-04
feat: getIdByNumber function added to check the real phone number

felipeDS91:
Here in Brazil we have a problem with some phone numbers that doesn't have the number "9" in front the phone number, because some years ago, all phone numbers of the country received an additional number "9". So when you try to send a message for one of these old numbers with the new pattern (with "9"), whatsapp thinks that's a different number and starts a new chat.
With this function, we can send the number always with the "9" digit and if it is an old number, this function will return the right number (without "9").

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -713,6 +713,18 @@ class Client extends EventEmitter {
     }
 
     /**
+     * Returns a real id registered in whatsapp
+     * @param {string} id the whatsapp user's ID
+     * @returns {Promise<string>}
+     */
+    async getIdByNumber(id) {
+        return await this.pupPage.evaluate(async (id) => {
+            let result = await window.Store.Wap.queryExist(id);
+            return result.status === 200 ? result.jid._serialized : undefined;
+        }, id);
+    }
+
+    /**
      * Create a new group
      * @param {string} name group title
      * @param {Array<Contact|string>} participants an array of Contacts or contact IDs to add to the group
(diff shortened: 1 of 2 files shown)


src/Client.js
```diff
@@ -712,6 +712,18 @@ class Client extends EventEmitter {
...
 
+    /**
+     * Returns a real id registered in whatsapp
+     * @param {string} id the whatsapp user's ID
+     * @returns {Promise<string>}
+     */
+    async getIdByNumber(id) {
+        return await this.pupPage.evaluate(async (id) => {
+            let result = await window.Store.Wap.queryExist(id);
+            return result.status === 200 ? result.jid._serialized : undefined;
```
vitorleonel, line 723: Hey,

Do you think it could be better if it returned null when you didn't find it?

felipeDS91: Yes, of sure. I will change it.

felipeDS91: Thanks by suggestion.

vitorleonel (approved):

lucasrosa90:
Dont need this PR, the lib already support that.

```javascript
try {
  const conversation = await this.client.pupPage.evaluate(
    async (number) => {
      return await window.WWebJS.getNumberId(number);
    },
    id
  );
  console.log(`Registered chat ID: ${conversation._serialized}`);
      return conversationId._serialized;
} catch (err) {
  console.error(`Chat ID not registered.`);
}
```

pedroslopez:
The `getNumberById` function was being used in the past for sending messages to new contacts, but the implementation has changed and it's not being used by the library for anything right now.

If this is the case and it helps with brazilian numbers, maybe instead it would be good to expose the function directly in the client through something like `client.getNumberId()`.

pedroslopez:
Closing in favor of #483

Issue #427, opened 2020-11-11
Revoked Message ID


mehmetonurkizildas:
Hello,

When the message is deleted from the other party, a newly derived ID is received in the incoming log. How can I get the old ID of the message.
I tried to explain in more detail in the pictures.

![Screenshot_1](https://user-images.githubusercontent.com/49166232/98777826-79fb9f00-2402-11eb-9a9c-371ba48f8366.png)

![Screenshot_2](https://user-images.githubusercontent.com/49166232/98777847-808a1680-2402-11eb-825b-02fef36f6883.png)

The ID I need is false_905397291588@c.us_3EB0D5B85E21FA412C16

Because I am doing some operations accordingly.

tudonetwork:
Hello!
You can use the before data.

client.on('message_revoke_everyone', async (after, before) => {
// Fired whenever a message is deleted by anyone (including you)

// message before it was deleted.
 console.log(" - message_revoke_everyone:before", before);

// message after it was deleted.
 console.log(" - message_revoke_everyone:after", after);

});

mehmetonurkizildas:
> Merhaba!
> Önceki verileri kullanabilirsiniz.
> 
> client.on ('message_revoke_everyone', async (sonra, önce) => {
> // Bir ileti herhangi biri (siz de dahil) tarafından silindiğinde tetiklenir
> 
> // silinmeden önceki mesaj.
> console.log ("- message_revoke_everyone: önce", önce);
> 
> // silindikten sonra mesaj.
> console.log ("- message_revoke_everyone: sonra", sonra);
> 
> });

Unfortunately, as I mentioned in the first picture, it generates a new ID and gives it to me. But what I need is the ID in the first message. Because I am taking some actions accordingly.


![Screenshot_3](https://user-images.githubusercontent.com/49166232/99036081-156f4a00-2592-11eb-8dee-3ca13603588a.png)


![Screenshot_4](https://user-images.githubusercontent.com/49166232/99036096-199b6780-2592-11eb-8838-386f710bf34e.png)

mpirescarvalho:
any updates on this?

mehmetonurkizildas:
> any updates on this? have you found any workaround?

Unfortunately. The same situation continues. I couldn't get back

mpirescarvalho:
hey @mehmetonurkizildas check if this PR #470 somehow helps you

mehmetonurkizildas:
[...]

PR #470, opened 2020-12-06
feat: check if the message has been revoked

mpirescarvalho:
Adds the ability to check whether a message has been revoked or not.

This may help those who want to do something with the original ID of a revoked message and can't do it because `message_revoke_everyone` gives a different ID for revoked message other than the original message ID  (#427) .

Exemple of usage: 
```javascript
  if (checkBadWord(msg.body)) {
    await msg.reply(
      'This kind of message is not allowed in this group, please delete it'
    );
  
    setTimeout(async () => {
      const revoked = await msg.checkRevoked();
      if (!revoked) {
          banUser(msg.from, msg.author);
      }
    }, 15000);
  }
});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -504,6 +504,10 @@ declare namespace WAWebJS {
          * Forwards this message to another chat
          */
         forward: (chat: Chat | string) => Promise<void>,
+        /**
+         * Checks if the message has been revoked
+         */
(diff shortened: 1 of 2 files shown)


Matheus-Lima-Moreira (commented):
This works fine but i dont know how this can help with the error of msg revoked with a different id from original

PR #475, opened 2020-12-09
Add method to mark chat as unread

olirock:
I'm adding the method to make it possible to mark a conversation as unread, that is, it is assigned unread = -1 following the whatsapp model


olirock:
@pedroslopez  hello

src/Client.js
```diff
@@ -678,6 +678,17 @@ class Client extends EventEmitter {
...
     }
 
+    /**
+     * Mark Unread the Chat
+     * @param {string} chatId ID of the chat that will be unread
+     */
+    async markUnread(chatId) {
+        await this.pupPage.evaluate(async chatId => {
+            let chat = await window.Store.Chat.get(chatId);
+            await window.Store.Cmd.markChatUnread(chat, -1);
```
pedroslopez, line 688: Where did you get this `-1` value from? In the code as far as I can tell, it seems to send `true` in this param.

src/structures/Chat.js
```diff
@@ -161,6 +161,13 @@ class Chat extends Base {
         return this.client.unmuteChat(this.id._serialized);
     }
 
+    /**
+     * Unread this chat
+     */
+    async markunread(){
```
pedroslopez, line 167:
```suggestion
    async markUnread(){
```

src/Client.js
```diff
@@ -678,6 +678,17 @@ class Client extends EventEmitter {
         }, chatId);
     }
 
+    /**
+     * Mark Unread the Chat
+     * @param {string} chatId ID of the chat that will be unread
+     */
+    async markUnread(chatId) {
```
pedroslopez, line 685:
```suggestion
    async markChatUnread(chatId) {
```

pedroslopez (changes requested):
Thanks for contributing!

Just a couple of notes here. Also, please be sure to add these new functions to the typings.

src/Client.js
olirock: When I run the getchats method I get the value of unread as -1 in conversations that have been marked as unread. Ex: A conversation that has no unread messages and the user marks the conversation as unread, the value that is assigned to the unreadcount is -1, since all messages have already been read, the user that marked as unread.

pedroslopez (approved):

PR #479, opened 2020-12-11
Feat/sticker

### LINKED ISSUE #95: Send Sticker
Now it's finding a way to send customized stickers through images in WEBP format

image

image

PR
https://github.com/pedroslopez/whatsapp-web.js/pull/94

### LINKED ISSUE #358: Sending A Sticker
probably this is duplicate , but i tried to send a sticker in this library 
but i get some problem to implement this code for injected.js
```javascript
window.WWebJS.sendSticker = async (sticker, chatId, metadata, type) =>{
    var chat = await window.Store.Chat.get(chatId);

    if(chat.erro === false || chat.__x_id){
    var ListChat = await Store.Chat.get(chatId),
           stick = new window.Store.Sticker.modelClass();

    stick.__x_clientUrl = sticker.clientUrl;
    stick.__x_filehash = sticker.filehash;
    stick.__x_id = sticker.filehash;
    stick.__x_uploadhash = sticker.uploadhash;
    stick.__x_mediaKey = sticker.mediaKey;
    stick.__x_initialized = false;
    stick.__x_mediaData.mediaStage = 'INIT';
    stick.mimetype = 'image/webp';
    stick.height = metadata && metadata.height ? metadata.height : 512;
    stick.width = metadata && metadata.width ? metadata.width : 512;
    
    await stick.initialize();
    
    var result = await (Promise.all((ListChat)? await stick.sendToChat(chat, {stickerIsFirstParty: false, stickerSendOrigin: 6}) : ""));
    result = result.join('');
    var m = {type: type},
    To = await WWebJS.getChat(chatId);
   
      }else{
          return chat;
      }
}	
```
that's show undefinied in console
or something wrong on that?

i mean like this
jjj

hope this feature will available in this library

mpirescarvalho:
First, good job with the project! 
This PR adds the possibility to send `WEBP` images as stickers.

That's how you would send stickers: 

```javascript
const webpBase64Data = 'UklGRlxWAABXRUJQVlA4WAoAAAAQAAAAgAMAFgIAQUxQSFQ0AAAB...';

    if (msg.body.startsWith('dp!sticker')) {
        const stickerMedia = new MessageMedia('image/webp', webpBase64Data, 'image.webp' );
        client.sendImageAsSticker(msg.from, stickerMedia);
    }
});
```

It's also possible to send using `chat` objects: 

```javascript
const chat = await msg.getChat();
const stickerMedia = new MessageMedia('image/webp', webpBase64Data, 'image.webp' );
chat.sendImageAsSticker(stickerMedia)
```

You could also load the image from file just like normal images:
```javascript
const stickerMedia = MessageMedia.fromFilePath('./image.webp');
```

_**IMAGES HAS TO BE IN `WEBP` FORMAT FOR THE STICKERS TO BE SENT PROPERLY**_

That's the result:


src/util/Injected.js
```diff
@@ -153,6 +155,57 @@ exports.LoadUtils = () => {
         return window.Store.Msg.get(newMsgId._serialized);
     };
 
+    window.WWebJS.sendImageAsSticker = async function (mediaInfo, chatId) {
+        const stickerInfo = await window.WWebJS.processStickerData(mediaInfo);
+        return await window.WWebJS.sendSticker(stickerInfo, chatId);
+    };
+
+    window.WWebJS.sendSticker = async function (stickerInfo, chatId) {
```
Aliyss, line 163: I would rename this to sendStickerRaw, as there still is potential to add a sendSticker via Sticker Packs which is not yet included in this PR.

Aliyss:
@mpirescarvalho 

Also I found some functions that could help retrieve Information on already added Stickers if you want to add that as well:

**Retrieve StickerPacks from phone:**
```js
Store.StickerPack.sync()
```
**Check if StickerPacks are fetched from phone:**
```js
Store.StickerPack.hasFetchedData()
```
**Retrieve Stickers from StickerPack:**
```js
Store.StickerPack.models[0].stickers.fetch()
```
**Sticker Info:**
```js
Store.StickerPack.models[0].stickers.models[0]
```
**Reset StickerPacks**
```js
Store.StickerPack.reset()
```
**Reset Stickers from StickerPack:**
```js
Store.StickerPack.models[0].stickers.reset()
```

mpirescarvalho:
@Aliyss 

Cool, I'd really apreciate if you can add that

`Store.StickerPack.hasFetchedData()` can also be useful to check if stickers have been fetched instead of fetching them every time a sticker is sent

Also, after using it for a while, I think `sendImageAsSticker` should also be renamed, as it is also possible to use it to send gifs or short videos converted to webp format (maybe `sendMediaAsSticker` ?). In fact the only requirements is that media has to be in `WEBP` format and be less than 1MB in size

Another thing is that I think `client.sendImageAsSticker` (or whatever the renamed name will be) should be able to send any image format. For that we would need to convert it internally, but it would require the use of external libs, and I don't know what @pedroslopez thinks about it. If he's positive I can work on it, because I’ve already implemented it in my bot

Aliyss:
@mpirescarvalho 

I'm currently short on time. So I doubt I will be able to implement it soon enough on my own.

Concerning the internal conversion: I see the added value of converting images internally, however I believe it should only be done if it's via WhatsApp. I assume, that WhatsApp has their own conversion tool integrated somewhere within their code. Or at least hope so.

pedroslopez:
Thanks for contributing! 

I haven't yet had the chance to look over this properly, but just seeing it very quickly I see stickers are sent using a new function `client.sendImageAsSticker()`. To keep things consistent, I would prefer to send these messages in the same way other messages are sent, using `client.sendMessage()`. It could possibly be an option when creating the MessageMedia or an option on the `sendMessage()` function, like we did with `sendAudioAsVoice: true` (although, I think maybe MessageMedia would be a better place for it). 

I haven't yet looked into the implementation. Let me know your thoughts.

mpirescarvalho:
@Aliyss,
I don't think whatsapp has their own conversion tool because you can't send new stickers through whatsapp web in normal use, you can only send stickers you have imported in android/ios app first... and it seems that this constraint of media being `WEBP` just happens in stickers.

@pedroslopez,
I would rather prefer using something like `sendMediaAsSticker: true` to keep options consistent with `sendAudioAsVoice`, but sending stickers is done by a different aproach of that used when sending messages. Of course we could just use a `if` and use the send sticker logic when `sendMediaAsSticker === true`, but sending sticker also doesn't return the message that was sent like `client.sendMessage()` does.

Let me know if even then you and me to join this on `sendMessage()`.

What about internal conversion so that it could handle formats other than `webp`?

mpirescarvalho:
@pedroslopez,
I would rather prefer using something like `sendMediaAsSticker: true` to keep options consistent with `sendAudioAsVoice`, but sending stickers is done by a different aproach of that used when sending messages. Of course we could just use a `if` and use the send sticker logic when `sendMediaAsSticker === true`, but sending sticker also doesn't return the message that was sent like `client.sendMessage()` does.

Let me know if even then you and me to join this on `sendMessage()`.

What about internal conversion so that it could handle formats other than `webp`?

esteves67:
@mpirescarvalho Thank you so much for implementing the send sticker function :smile: :+1:

pedroslopez:
> @pedroslopez,
> I would rather prefer using something like `sendMediaAsSticker: true` to keep options consistent with `sendAudioAsVoice`, but sending stickers is done by a different aproach of that used when sending messages. Of course we could just use a `if` inside `sendMessage()` and use the send sticker logic when `sendMediaAsSticker === true`, but sending sticker also doesn't return the message that was sent, like `client.sendMessage()` does.

Yeah, `sendMediaAsSticker` seems good to me- in this case I think it would be ok to internally convert images to their required format automatically as you mentioned.

However, if we're sending messages using the regular `sendMessage` function, it should return the sent message. Is there a way you could manually assign a message ID to the sticker message? What I did to be able to return sent messages was basically rewrite the internal function that created the message and manually generate a new ID for it.

mpirescarvalho:
Ok, so a lots of things have changed here.

### Sending Sticker
First, initial functions `client.sendImageAsSticker(msg.from, stickerMedia);` does not exist anymore.
As @pedroslopez requested, sending stickers is now merged into `client.sendMessage()` functions. That means it's now possible to send stickers from other functions like `msg.reply()` or `chat.sendMessage()`.

Usage example:
```javascript
const webpBase64Data = 'UklGRlxWAABXRUJQVlA4WAoAAAAQAAAAgAMAFgIAQUxQSFQ0AAAB...';

    if (msg.body.startsWith('dp!sticker')) {
        const stickerMedia = new MessageMedia('image/webp', webpBase64Data );
        client.sendMessage(msg.from, stickerMedia, { sendMediaAsSticker: true });
    }
});
```

### Media format
Media format constraint also doesn't exists anymore (images are converted internally), so it's now possible to send any image format like jpeg to the function. You can do:
```javascript
const stickerMedia = MessageMedia.fromFilePath('imagem.jpg');
client.sendMessage(msg.from, stickerMedia, { sendMediaAsSticker: true });
```

### Videos
You can also send any videos format as stickers (though I only tested it with mp4). You should keep in mind that whatsapp stickers max size is 1MB, so anything bigger than that will not be sent, resulting in a error.

**Important:**
If you send a video as sticker that is **not in webp** format, it will try to automaticaly convert it. You need to make sure you have [ffmpeg](https://ffmpeg.org/) installed for this to work. Also, if you need, you can pass the ffmpeg path while instantiating whatsapp-web.js client:

```javascript
const client = new Client({
    ...
    ffmpegPath: 'path/to/ffmpeg'
});
```

pedroslopez:
This is great! That's exactly the way I would have hoped to implement it and it's integrated nicely.
However, the functionality doesn't seem to be working.

Trying to send stickers, it gets stuck pending:
image

I tried with a random image I had, as well as trying to send back a sticker that is already in WhatsApp.

mpirescarvalho:
> This is great! That's exactly the way I would have hoped to implement it and it's integrated nicely.
> However, the functionality doesn't seem to be working.
> 
> Trying to send stickers, it gets stuck pending:
> image
> 
> I tried with a random image I had, as well as trying to send back a sticker that is already in WhatsApp.

Are you getting any error message? It works nice for me:

```javascript
    const webpImage = MessageMedia.fromFilePath('image.webp');
    const jpgImage = MessageMedia.fromFilePath('image.jpg');
    const receivedImage = await msg.downloadMedia();

    client.sendMessage(msg.from, webpImage, { sendMediaAsSticker: true });
    client.sendMessage(msg.from, jpgImage, { sendMediaAsSticker: true });
    client.sendMessage(msg.from, receivedImage, { sendMediaAsSticker: true });
});
```

image

pedroslopez:
Yeah, I'm getting this in the console:
image


That failed request returns this body:
```
{
   "error": {
      "message": "(#403) Invalid timestamp for authentication",
      "type": "OAuthException",
      "code": 403,
      "fbtrace_id": "AV_XXXXXXXXXXXX"
   }
}
```

It only seems to get triggered on the first attempt to send a sticker. Subsequent attempts are also stuck loading, but no more errors show up on the console.

pedroslopez:
Disregard my previous message, the issue I mentioned seems to be unrelated and happens even without sending any stickers.
So no, no issue in the console. Stickers just get stuck in a pending state.

mpirescarvalho:
> Disregard my previous message, the issue I mentioned seems to be unrelated and happens even without sending any stickers with the library.
> So no, no issue in the console. Stickers just get stuck in a pending state.
> 
> Could more data be missing from the message? I know stickers have some information like sticker name and pack name, but I don't see those being set anywhere.

Yes they have aditional information but those seems to be optional. I'm using this branch on my bot (it allows users to send images and the bot creates a sticker from it) from a week and I haven't had any problems at all.

I think we need anyone else to test it, maybe I'm missing something too.

esteves67:
@mpirescarvalho @pedroslopez 
Im having no problem with mpires sticker branch, here it is working perfect.

image
image

@mpirescarvalho again, thanks for implementing the sticker function, it will be so much useful for me, and many others!

pedroslopez:
Alright, the issue I was having was due to the `size` property not being set. This caused an issue on iOS with both sending and receiving stickers sent by the library.

I just need to test out the animated sticker portion, which I'll get to most likely tomorrow.

pedroslopez (approved):
LGTM!

PR #484, opened 2020-12-19
fetching chats by ids & fetching a single message from a chat by id

samer95:


src/structures/Chat.js
```diff
@@ -192,6 +192,22 @@ class Chat extends Base {
...
+    async getMessageById(messageId) {
+        let message = await this.client.pupPage.evaluate(async (chatId, messageId) => {
+            const chat = window.Store.Chat.get(chatId);
+            const msg = chat.msgs.models.find(m => m.id.id === messageId);
+            return window.WWebJS.getMessageModel(msg);
+
+        }, this.id._serialized, messageId);
+
+        return new Message(this.client, message);
+    }
```
mpirescarvalho, line 209: You might want to take a look at `window.Store.Msg.get(msgId)`. It allows you to retrive msg by id and no need to getChat operation

PR #485, opened 2020-12-19
GroupChat types

mpirescarvalho:
adds groupchat types for typescript

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -775,6 +775,12 @@ declare namespace WAWebJS {
 
     }
 
+    export type GroupParticipant = {
+        id: ContactId,
+        isAdmin: boolean
+        isSuperAdmin: boolean
+    }
+
     /** Promotes or demotes participants by IDs to regular users or admins */
     export type ChangeParticipantsPermisions = 
         (participantIds: Array<string>) => Promise<{ status: number }>
@@ -787,7 +793,7 @@ declare namespace WAWebJS {
         /** Group description */
         description: string;
         /** Group participants */
-        participants: Array<any>;
+        participants: Array<GroupParticipant>;
         /** Adds a list of participants by ID to the group */
         addParticipants: (participantIds: Array<string>) => Promise<void>;
         /** Removes a list of participants by ID to the group */
--- a/index.d.ts
+++ b/index.d.ts
@@ -800,7 +800,7 @@ declare namespace WAWebJS {
 
     export interface GroupChat extends Chat {
         /** Group owner */
-        owner: any;
+        owner: ContactId;
         /** Date at which the group was created */
         createdAt: Date;
         /** Group description */
--- a/index.d.ts
+++ b/index.d.ts
@@ -775,6 +775,10 @@ declare namespace WAWebJS {
 
     }
 
+    /** Promotes or demotes participants by IDs to regular users or admins */
+    export type ChangeParticipantsPermisions = 
+        (participantIds: Array<string>) => Promise<{ status: number }>
+
     export interface GroupChat extends Chat {
         /** Group owner */
         owner: any;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -774,6 +774,46 @@ declare namespace WAWebJS {
...
+        /** Group owner */
+        owner: any;
+        /** Date at which the group was created */
+        createdAt: Date;
+        /** Group description */
+        description: string;
+        /** Group participants */
+        participantes: Array<any>;
+        /** Adds a list of participants by ID to the group */
+        addParticipants: (participantIds: Array<string>) => Promise<void>;
```
pedroslopez, line 788: Returns an object, but it's a bit tricky to type because it has keys that correspond to the contact IDs that are trying to be added, indicating a status for each one that gives info about whether the user was able to be added or not.

Example:
```js
await chat.addParticipants(['1234567890@c.us'])

{
   status: 207, // (integer)
    participants: [
        '1234567890@c.us': {
            code: '200' // (string)
        }
    ],
    '1234567890@c.us': 200 // (integer)
}
```

I'm not sure how you would want to type this. Maybe `<any>` is fine for now- it might be worth to process the output at a later point to make it a bit more user-friendly, but that would introduce breaking changes for people that are relying on the current behavior. I'm open to feedback on this.

index.d.ts
```diff
@@ -774,6 +774,46 @@ declare namespace WAWebJS {
...
+        /** Removes a list of participants by ID to the group */
+        removeParticipants: (participantIds: Array<string>) => Promise<void>;
```
pedroslopez, line 790: Same situation as with `addParticipants`. Could return `Promise<any>` for now.

index.d.ts
```diff
@@ -774,6 +774,46 @@ declare namespace WAWebJS {
...
+        /** Promotes participants by IDs to admins */
+        promoteParticipants: (participantIds: Array<string>) => Promise<void>;
```
pedroslopez, line 792: Returns an object with a status code:
```js
{
    status: 200 // (integer)
}
```

index.d.ts
```diff
@@ -774,6 +774,46 @@ declare namespace WAWebJS {
...
+        /** Demotes participants by IDs to regular users */
+        demoteParticipants: (participantIds: Array<string>) => Promise<void>;
```
pedroslopez, line 794: Returns an object with a status code:

{
    status: 200 // (integer)
}

pedroslopez (changes requested):
A couple notes

index.d.ts
mpirescarvalho: I'm not sure if I understood well the type of the key `participants`. Would that be:

```javascript
{
    status: number;
    participants: Array<{
        [key: string]: {
           code: number
        }
    }>
} & {
    [key: string]: number;
}

//This translates to:

{
    status: 207,
    participants: [
        {
            '1111111111@c.us': {
                code: '200' 
            }
        },
        {
            '2222222222@c.us': {
                code: '200' 
            }
        }
    ],
    '1111111111@c.us': 200,
    '2222222222@c.us': 200, 
}
```

What about this one:

```javascript
{
    status: number;
    participants: {
        [key: string]: {
            code: number
        }
    }
} & {
    [key: string]: number;
}

//This translates to:

{
    status: 207,
    participants: {
        '1111111111@c.us': {
            code: '200' 
        },
        '2222222222@c.us': {
            code: '200' 
        }
    },
    '1111111111@c.us': 200,
    '2222222222@c.us': 200, 
}
```

mpirescarvalho: About processing this output, I think `participants` key is unecessary, because we can easily access its value on those properties:
```javascript
{
    ...
    '1111111111@c.us': 200,
    '2222222222@c.us': 200,
}
```

pedroslopez: It would be your first example (the one where participants is an array)

mpirescarvalho: done

index.d.ts
```diff
@@ -774,6 +774,69 @@ declare namespace WAWebJS {
...
+                    code: number
+                }
+            }>
+         } & {
+             [key: string]: number;
+         }>
+
+    export interface GroupChat extends Chat {
+        /** Group owner */
+        owner: any;
```
mpirescarvalho, line 803: @pedroslopez what about this one?

pedroslopez: It's a ContactId

mpirescarvalho: All done

pedroslopez (approved):

PR #494, opened 2020-12-23
feat: star/unstar messages

mpirescarvalho:
adds functions to star/unstar messages to `client`, `chat` and `message`.

```javascript
client.starMsgs([msg1, msg2]);
client.unstarMsgs([msg1, msg2]);

chat.starMsgs([msg1, msg2]);
chat.unstarMsgs([msg1, msg2]);
    
msg.star();
msg.unstar();
```

Usage: 

```javascript
client.on('message', async msg => {
    if (msg.body.startsWith('!star')) {
        const quotedMsg = await msg.getQuotedMessage();
        await quotedMsg.star();
    }

    if (msg.body.startsWith('!unstar')) {
        const quotedMsg = await msg.getQuotedMessage();
        await quotedMsg.unstar();
    }
});
```

Result (bot point of view):
image

Would it be better if `client.starMsgs` and `chat.starMsgs` receives an array of IDs instead of messages?

closes #481

PR #501, opened 2020-12-27
Added star data to message structure

antoineschaller:
Completes #494


pedroslopez:
Did a quick renaming to avoid conflicts with the function defined in #494 
Thanks for this!

PR #527, opened 2021-01-29
Add support for sticker Name and author

purpshell:
This adds sticker name and author support by modifying EXIF metadata of the webp sticker file. I extracted the exif data from a legit sticker, so this should function properly... 
Example image

image


PurpShell:
this also requires the installation of an npm package called "webp-convert"

luizferrbarbosa (approved):

src/Client.js
```diff
@@ -475,6 +479,27 @@ class Client extends EventEmitter {
...
+                const exifPath = 'data.exif';
+                const resultPath = 'sticker.webp';
+                internalOptions.attachment = await Util.formatToWebpSticker(internalOptions.attachment);
+                fs.writeFileSync(resultPath, internalOptions.attachment.data, 'base64');
+                const random_id = Math.floor(Math.random() * (9999999999 - 1000000000) + 1000000000);
+                const stickerpackid = "com.marsvard.stickermakerforwhatsapp.stickercontentprovider "+ random_id;
+                const packname = options.stickerName || "undefined";
+                const author = options.stickerAuthor || "undefined";
+                const googlelink = "https://play.google.com/store/apps/details?id=com.marsvard.stickermakerforwhatsapp";
+                const applelink = "https://itunes.apple.com/app/sticker-maker-studio/id1443326857";
```
mpirescarvalho, line 492: I don't think every sticker should have information of another person's app (stickerpackid and links to google and apple store). Maybe passing theses throught options along with name and author might be a good option. 

If you default them to null author and name still works?

src/Client.js
```diff
@@ -492,6 +517,21 @@ class Client extends EventEmitter {
...
+                window.WWebJS.sendSeen(chatId);
+            }
+
+            const msg = await window.WWebJS.sendMessage(chat, message, options, sendSeen);
+            return msg.serialize();
+        }, chatId, content, internalOptions, sendSeen);
+
+        return new Message(this, newMessage);
+    }
+
```
mpirescarvalho, line 534: Why add this repeated code? is it a mistake?

mpirescarvalho (commented):
First, **all my comments are just suggestions**, I'm not the one who will merge this.

I suggest you to fix linting issues, this will make you PR gets more attention.

Another thing is that I don't think this code should be added in `Client.js` file, maybe `processStickerData` function at `Injected.js` would be a better place to handle this

If I can help you in some way just let me know, I'm looking forward to use this feature.

src/Client.js
PurpShell: WhatsApp provides the following exif data, since sticker makers are only probably available to apps via whatsapp's boilerplate (https://github.com/WhatsApp/stickers)
Not providing will probably chance an account getting banned?

PurpShell:
> If I can help you in some way just let me know, I'm looking forward to use this feature.
Thanks, that would be appreciated as I don't have much experience in puppeteer.

src/Client.js
PurpShell: hmmm

PurpShell: It is

purpshell (commented):
Fixed

src/Client.js
PurpShell: I will have to look more into that, as you are probably most right.. also if sticker author or name is specified then it adds that.

mpirescarvalho: I don't think this will increase the probabilities of getting banned because at the moment we are not providing any of that info

PurpShell: Okay then, removing that.

PurpShell:
Credits go to @mpirescarvalho for refactoring the code

PurpShell:
@pedroslopez

mpirescarvalho:
@PurpShell there's a last PR I pushed, please take a look!

PurpShell:
Okay

mpirescarvalho (approved):

src/util/Util.js
```diff
@@ -5,7 +5,11 @@ const path = require('path');
 const Crypto = require('crypto');
 const { tmpdir } = require('os');
 const ffmpeg = require('fluent-ffmpeg');
+const webp = require('webp-converter');
 const fs = require('fs').promises;
+const exists = require('fs').existsSync;
+const MessageMedia = require('../structures/MessageMedia');
```
pedroslopez, line 11: Move internal dependency down. External dependencies should come first.

src/util/Util.js
```diff
@@ -135,16 +149,56 @@ class Util {
...
+
+
+        if (metadata.name || metadata.author) {
+            const tempPath = os.tmpdir();
+            const hash = this.generateHash(32);
+            const exifPath = `${tempPath}/${hash}.exif`;
+            const resultPath = `${tempPath}/${hash}.webp`;
+            try {
+                await fs.writeFile(resultPath, webpMedia.data, 'base64');
+                const stickerpackid = hash;
```
pedroslopez, line 184: use camelCase
```suggestion
                const stickerPackId = hash;
```

pedroslopez (changes requested):
Functionality everything seems to work ok, just a couple of style things to correct.

However, I would like to keep dependencies down whenever possible. Is `sharp` still needed to convert images to webp? Could we just use this new `webp-convert` library to do what we were doing with `sharp`? Or better yet, is it possible to set exif metadata with `sharp`?

Edit: just saw that the webp-converet lib is **34 MB** unpacked! If we only need it to set the exif data maybe there's a better option. (Doing a quick search https://github.com/hMatoba/piexifjs seems to have much more usage in terms of downloads and is much smaller)

PurpShell:
> Functionality everything seems to work ok, just a couple of style things to correct.
> 
> However, I would like to keep dependencies down whenever possible. Is `sharp` still needed to convert images to webp? Could we just use this new `webp-convert` library to do what we were doing with `sharp`? Or better yet, is it possible to set exif metadata with `sharp`?
> 
> Edit: just saw that the webp-converet lib is **34 MB** unpacked! If we only need it to set the exif data maybe there's a better option. (Doing a quick search https://github.com/hMatoba/piexifjs seems to have much more usage in terms of downloads and is much smaller)

As for the first part: 
https://www.npmjs.com/package/webp-converter can convert different image types to webp. (animated included, but requires GIF )

PurpShell:
As for the lib you recommended, It doesn't seem to support WebP...

PurpShell:
For the requested changes, I will be releasing a commit soon.

PurpShell:
webp-converter has the options to convert images to webp, with gif support, also it has a feature of turning a base64 string of an image into a webp vesion

PurpShell:
@pedroslopez

PurpShell:
> Functionality everything seems to work ok, just a couple of style things to correct.
> 
> However, I would like to keep dependencies down whenever possible. Is `sharp` still needed to convert images to webp? Could we just use this new `webp-convert` library to do what we were doing with `sharp`? Or better yet, is it possible to set exif metadata with `sharp`?
> 
> Edit: just saw that the webp-converet lib is **34 MB** unpacked! If we only need it to set the exif data maybe there's a better option. (Doing a quick search https://github.com/hMatoba/piexifjs seems to have much more usage in terms of downloads and is much smaller)

We can also go the route of stripping the functionality out of the lib

pedroslopez:
> webp-converter has the options to convert images to webp, with gif support, also it has a feature of turning a base64 string of an image into a webp vesion

If we can do everything that we're using `sharp` for with this new `webp-converter` library, maybe we can just go with that and remove the `sharp` dependency completely? Seems to have a lot of the same functionality.

I would have preferred to use `sharp`, but while the feature has been requested and someone made some progress a long time ago (https://github.com/lovell/sharp/issues/650) unfortunately it doesn't seem like anybody's working on adding exif metadata functionality to sharp.

(cc @mpirescarvalho since you initially worked on this)

PurpShell:
I will look into it, I will post an update later today on if it works

mpirescarvalho:
@pedroslopez I think that could be done, but I'm afraid that webp-converter doesn't have this funcionality:
```js
sharpImg = sharpImg.resize(512, 512, {
    fit: 'contain',
    background: { r: 0, g: 0, b: 0, alpha: 0 },
});
```

Without that the image will lose it's aspect ratio, but we could run some tests and see how it goes.

mpirescarvalho:
What about shipping that webpmux binary that we discussed on discord, so we don't end up with 32mb from webp-converter, but rather with something around 900kb?

PurpShell:
> What about shipping that webpmux binary that we discussed on discord, so we don't end up with 32mb from webp-converter, but rather with something around 900kb?

yup!
we can use that
i think I shall start looking deep for other alternatives, or even better, compiling webpmux to suit our needs
i think it is taking this much space, is because it has a lot of features, today i will make sure to send a reply with a solution to this problem

PurpShell:
i think we should just stick to using the webpmux binaries and select them based on OS, and launch them using child_processes

PurpShell:
https://github.com/ApeironTsuka/node-webpmux looks promising too

PurpShell:
Yo!
We managed to talk to @AperionTsuka, and he made some patches to AperionTsuka/node-webpmux!
We can now use it to make our webp dreams come true!

it is only 200kb in raw size
and 
it allows me to get rid of saving the exif straight into a file then loading

PurpShell:
@pedroslopez

PurpShell:
hmm, well node-webpmux is kinda an es6 not a commonjs module, you will need node 13 :hankey:

PurpShell:
It is done! I have successfully added the sticker author and sticker name fields!

PurpShell:
just finished another batch of commits, yeah ik i should have tested locally because all of this is just linting.
Optimizations: Now gotten rid of setting temporary files, just uses built in functions to load the image / animation

PurpShell:
This is the last commit i am gonna publish to this pr, unless anything goes wrong..

PurpShell:
@pedroslopez 
check this one out :)

PurpShell:
Just added sticker Categories, it is an array of emojies that put your stickers in the categories menu of stickers. 

For more info: https://github.com/WhatsApp/stickers/wiki/Tag-your-stickers-with-Emojis

pedroslopez:
Thank you!

pedroslopez (approved):

PR #547, opened 2021-02-15
Fixes star() & unstar() functions

antoineschaller:
Fixes missing "client" in star() and unstar() functions in Message structure.


pedroslopez:
Good catch! It seems the latest refactor should have been tested more.
Thanks!

PR #550, opened 2021-02-16
Add getMessageById

wictornogueira:
Adds getMessageById method to Client.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -56,6 +56,9 @@ declare namespace WAWebJS {
         /** Get all current contact instances */
         getContacts(): Promise<Contact[]>
 
+        /** Get message instance by ID */
+        getMessageById(msgId: string): Promise<Message>
+
         /** Get all current Labels  */
         getLabels(): Promise<Label[]>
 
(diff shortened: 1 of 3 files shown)


pedroslopez:
Thanks for contributing!

The reason why I haven't implemented this is because it only works for messages that have been cached by WhatsApp Web. Older messages will not be included in the cache, so getting messages by ID using this method can produce unreliable results. Getting messages by ID is something that is necessary and would be super useful to have, but this limitation of only being able to retrieve cached messages could create confusion and is not the end goal. I would like to avoid "stopgaps" or temporary solutions whenever possible.

I have some ideas for how getting messages by ID could potentially be done, but I have to dedicate some more time into exploring these and testing to make sure it works.

If people really want to be able to get _cached_ messages, let me know of your use cases and I'll definitely consider it.

wictornogueira:
Oh, I see...
Someone on Discord was trying to do exactly that, so I implemented this method to help them.
Anyway, if I come up with something more useful, I will let you know.

Thank you.

PR #552, opened 2021-02-17
add getMessageById (again)

wictornogueira:


pedroslopez:
This seems like it would work, but I think the issue with this approach would be that it could take quite a long time to get the specific message, since every 30 or 50 messages a fetch occurs to WhatsApp to get the next set of messages sequentially. Someone that's using this function would expect the message retrieval to occur directly.

A way I've been thinking about tackling this would be to check out how individual messages are retrieved when you click on a quoted message on WhatsApp Web. When clicking the message, WhatsApp scrolls up and fetches the messages in that location. This could probably be used to more efficiently get a message by its ID.

PR #553, opened 2021-02-17
add getLastSeen method

### LINKED ISSUE #82: Check user online state. [feature request]
Hi everyone.

Is it possible to get user online state? (like : online, seen today at X, etc)

Thank you!

wictornogueira:
closes #82


pedroslopez:
I was initially trying to implement this back here https://github.com/pedroslopez/whatsapp-web.js/commit/ddfb616d1bb41f52aa194f71150f10a000050095 but I haven't had time to finish it. Thanks for working on this!

The issue I encountered back then was that getting the presence information could be out of date or not accurate when retrieving it if you're not subscribed to their updates. This can be seen with people that are far down in the chat list.

Unfortunately the function used to subscribe to presence updates doesn't update the presence right away, so I was experimenting with how to accurately retrieve the state after subscribing. One could argue a `setTimeout` could be used with a fixed value, but I'd like to avoid this and use something more reliable. Would appreciate if you could look into this!

Invy55:
> I was initially trying to implement this back here [ddfb616](https://github.com/pedroslopez/whatsapp-web.js/commit/ddfb616d1bb41f52aa194f71150f10a000050095) but I haven't had time to finish it. Thanks for working on this!
> 
> The issue I encountered back then was that getting the presence information could be out of date or not accurate when retrieving it if you're not subscribed to their updates. This can be seen with people that are far down in the chat list.
> 
> Unfortunately the function used to subscribe to presence updates doesn't update the presence right away, so I was experimenting with how to accurately retrieve the state after subscribing. One could argue a `setTimeout` could be used with a fixed value, but I'd like to avoid this and use something more reliable. Would appreciate if you could look into this!
> 
> Also, I think having this method on the `Contact` would make more sense.

I was looking into this and I'm having trouble with data disappearing on a normal Firefox instance with manual code execution, any suggestion? I think it has to be something with focusing or the window not being used for some time. (I know this isn't strictly related but I'm hoping in some help)

tuyuribr (changes requested):
`  chat.presence.subscribe()` needs to be reliable

wictornogueira:
Mfw yuri necroposts lol

shirser121:
Is this still relevant?

purpshell (approved):
fix conflicts and merge

PR #586, opened 2021-03-04
add searchMessages method

wictornogueira:
Same thing as the search bar


PurpShell:
I am guessing this is better than queryMessages?

pedroslopez:
Nice!

Issue #383, opened 2020-10-15
Adding options to use proxy


Eonus21:
Hello, I want to use many clients on one server. I made it with Client.initialize() overriding. It is possible to add this feature to api? I can make pull request

guicuton:
It's possible.
You can see → https://github.com/pedroslopez/whatsapp-web.js/issues/45

If you have a different approach, show us :-)

This will use more CPU and memory from server but i do not see any problem with that

IIE4EHbKAKOKC:
Yes, It is possible, but it would be better to use a proxy for each account. I made it with two new lines in Client.initialize()

pedroslopez:
Like @guicuton mentioned, all that's necessary for running more sessions is creating a new Client for each one. I'm not sure what you mean by using a proxy for each account. Please post your solution here.

IIE4EHbKAKOKC:
Yes, I can show

```
async initialize(credentials) {
        const browser = await puppeteer.launch(this.options.puppeteer);
    
        const page = (await browser.pages())[0]
    
        await page.authenticate(credentials);
        page.setUserAgent(UserAgent);

        this.pupBrowser = browser;
        this.pupPage = page;
        .....
}
```

This is example of my overriding to use proxy with auth

parthibd:
You can build a Rest API over WWJs or even as a gRPC service. For multiple clients I'll suggest not modifying the original source code of WWJs. Instead use docker containers for a sandboxed environment for each user. That way you'll have process isolation and will keep your code and architecture simple. Managing multiple clients from a single Nodejs process is not a recommended way because if the nodejs process goes down, so does all the clients. In a sandboxed environment even if one client goes down, but others remain intact. Although the number of clients that you can host on a single machine is constrained by the resources.

If your problem is solved, please close this issue.

jessekalil:
> Sim posso mostrar
> 
> ```
> async initialize(credentials) {
>         const browser = await puppeteer.launch(this.options.puppeteer);
>     
>         const page = (await browser.pages())[0]
>     
>         await page.authenticate(credentials);
>         page.setUserAgent(UserAgent);
> 
>         this.pupBrowser = browser;
>         this.pupPage = page;
>         .....
> }
> ```
> 
> Este é um exemplo da minha substituição para usar proxy com auth. Mas ele precisa de tratamento de erros e será melhor colocar as credenciais em ClientOptions no construtor

I need something like this...

Eonus21:
[...]

PR #594, opened 2021-03-17
Add option to use proxy with auth

Eonus21:
This is my idea to solve #383 . Hope you will advice what to improve in this merge request to be merged


Eonus21:
Maybe it will be better move proxyInfo in constructor options. Hope we will get compromise

pedroslopez:
It seems like there are other, unrelated changes in this PR. Please resubmit with a single PR per change.

PR #600, opened 2021-03-22
add extra option to MessageSendOptions

wictornogueira:
Small tweak for allowing property overriding.


pedroslopez:
What are you needing to override?
Curious to see if maybe there's something the library should support directly instead of this patch

PR #606, opened 2021-03-25
Browser disconnected

tofers:
Reset browser state on browser execution or crash. This is necessary to check whether the browser is running or not.


alirezaebiii (commented):
@Maryam

pedroslopez:
Haven't fully tested this yet, but you can also check this by calling `client.pupBrowser.isConnected()`, which returns a boolean. Is there a specific use case here where you think setting it to null would be more beneficial?

Issue #609, opened 2021-03-28
Disconnected not work ?


IrfanMaulana00:
after I scanned the qr, and logged out the wa web session from the phone, the code disconnected didn't work.

is there something wrong ?

`client.on('disconnected', (reason) => {
    console.log("DC");
});`

adisutomo:
yes me to..
client.on('message', async msg =>  not work

frmontini:
The disconnected event is also not working for me. A few days ago everything was fine. Perhaps an update to the web version may have broken this listener. It is important to say that I did not update my code before it stopped working. I am using the latest version of lib (v1.12.5).

arisros:
I made PR @pedroslopez 

`onAppStateChangedEvent` state now is just `WAState.OPENING`

thanks ✌️

IrfanMaulana00:
> I made PR @pedroslopez
> 
> `onAppStateChangedEvent` state now is just `WAState.OPENING`
> 
> thanks ✌️

thanks, this is working

PurpShell:
Closing this issue, you are welcome to open it again if you think this was done in error. 

Links: 
- [**Starter Guide**](https://waguide.pedroslopez.me)
- [**Documentation**](https://pedroslopez.me/whastapp-web.js)


Best regards, 

[PurpShell](github.com/PurpShell)

mfandre:
At version 1.9.2 is not working

PR #622, opened 2021-04-03
delete opening state from array #609

arisros:
Fix:
- Delete `WAState.OPENING` from `ACCEPTED_STATES` array  #609
- delete reason args, cause is not relevant
- Doc generated


pedroslopez:
This fix has some issues interacting with session takeovers if `takeoverOnConflict` is enabled. The reason why the `disconnected` event broke is because whatsapp web is doing something weird and refreshing the page before the right events are called. I'm detecting this refresh now as of 6cece4a0061fb2a1224b7c71618c31b406f52a94

PR #648, opened 2021-04-23
Add from which device the message was sent.

tuyuribr:
Use the message Id for fingerprint the device that was being use when the message was sent

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -82,6 +82,12 @@ class Message extends Base {
         this.author = (typeof (data.author) === 'object' && data.author !== null) ? data.author._serialized : data.author;
 
         /**
+         * String that represents from which device type the message was sent
+         * @type {string}
+         */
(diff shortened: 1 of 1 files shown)


pedroslopez:
This is interesting! How did you determine these things? Is there any reference that this is how it works anywhere in the code or otherwise?

tuyuribr:
> 
> 
> This is interesting! How did you determine these things? Is there any reference that this is how it works anywhere in the code or otherwise?

This is a pattern I found when searching for vulnerabilities in the code, Messages sent by WebW start with 3E, Messages sent by an Iphone start with 3A, messages sent from an Android device are 32 HEX in size.
Isn't documented anywhere but I got 0 errors from my user base ( 1k + )

PurpShell:
Interesting!

pedroslopez:
Seems to work for me!
thanks

Kouthk:
Have these been added to the features documents?

PR #650, opened 2021-04-25
Format phone numbers

purpshell:
My code adds two functions to wweb.js, 

Client instance
.getFormattedNumber(id) and .getCountryCode(id) 
 
Contact instance
.getFormattedId() and .getCountryCode()


PurpShell:
These functions add the spacing to the numbers
return null if not valid ids, though countrycode function seems to not care and provide a not working number

PurpShell:
Finalized.

Best regards, 

[PurpShell](https://github.com/PurpShell)

src/structures/Contact.js
```diff
@@ -120,6 +120,22 @@ class Contact extends Base {
         return await this.client.getProfilePicUrl(this.id._serialized);
     }
 
+    /**
+     * Returns the contact's formatted phone number, (+1 (234) 5678-901)
+     * @returns {Promise<string>}
+     */
+    async getFormattedId() {
```
pedroslopez, line 127: `getFormattedId` might be misleading since this doesnt quite represent a contact id. Maybe a better name is `getFormattedNumber`

src/Client.js
```diff
@@ -810,6 +810,45 @@ class Client extends EventEmitter {
         }
     }
 
+    /**
+     * Get the formatted number of a WhatsApp ID. 
+     * Will return null if the number is not registered on WhatsApp.
+     * @param {string} number Number or ID ("@c.us" will be automatically appended if not specified)
+     * @returns {Promise<string|null>}
+     */
+    async getFormattedId(number) {
```
pedroslopez, line 819: Same as the other comment, maybe a better name is `getFormattedNumber`

src/Client.js
```diff
@@ -810,6 +810,45 @@ class Client extends EventEmitter {
...
+        if(!number.endsWith('@s.whatsapp.net')) {
+            number = number.replace('c.us', '');
+            number += 's.whatsapp.net';
+        }
+
+        try {
+            return await this.pupPage.evaluate(async numberId => {
+                return window.NumberInfo.formattedPhoneNumber(numberId);
+            }, number);
+        } catch(_) {
```
pedroslopez, line 829: What does this throw that needs to return null? In what cases is this triggered. Maybe we should be more explicit of which error we're catching.

src/Client.js
```diff
@@ -810,6 +810,45 @@ class Client extends EventEmitter {
...
+     * @returns {Promise<string|null>}
+     */
+    async getCountryCode(number) {
+        number = number.includes(' ') ? number.replace(' ', '') : number;
+        number = number.includes('+') ? number.replace('+', '') : number;
+        try {
+            return await this.pupPage.evaluate(async numberId => {
+                return window.NumberInfo.findCC(numberId);
+            }, number);
+        } catch(_) {
```
pedroslopez, line 847: Same comment about errors.

pedroslopez (changes requested):
Just a few comments. Also, please remember to add typings for typescript.

src/Client.js
PurpShell: I don't quite remember, but seems like it doesn't throw an error. I'll remove it.

src/Client.js
PurpShell: Agreed

src/structures/Contact.js
PurpShell: Agreed

src/Client.js
PurpShell: 👍

PurpShell:
Hey! Sorry for the low quality on this PR, I opened it back in a time that coding wasn't my primary focus.. Hopefully these commits could fix it's rush and bad timing. Really feeling bad after I just dubbed it finalized a few days ago. 

Best regards, 

PurpShell

index.d.ts
```diff
@@ -745,9 +751,16 @@ declare namespace WAWebJS {
          * Will return null when getting chat for currently logged in user.
          */
         getChat: () => Promise<Chat>,
-
+        
+        /** Returns the contact's formatted phone number, (12345678901@c.us) => (+1 (234) 5678-901) */
+        getFormattedNumber(): Promise<string>,
+        
+        /** Returns the contact's countrycode, (1541859685@c.us) => (1) */
+        getFormattedNumber(): Promise<string>,
```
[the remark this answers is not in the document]
PurpShell: yes, I am too dumb to even remotely make a PR it seems :C

PR #660, opened 2021-05-03
Add Filesize to MessageMedia

purpshell:
# MessageMedia File size

##  INFO

This pull request is setting a new property to MessageMedia, allowing you to know the file size of the file. 

Accuracy-wise, it seems whatsapp gives the value of the file before compression so it seems a little off than in the preview.

Sometimes i have seen 2.7mb turn into 3 on the preview, it seems whatsapp rounds up on that counter

## DEMO

image

## USAGE

**FORMAT:** BYTES

Pass in the "filesize" property when making a file, it is optional though.
fromFilePath and downloadMedia automatically handle this.


tuyuribr:
This seems to load the IO more  to all iterations with messageMedia, can you make this optional (and null by default)?

aliyss (dismissed):
lgtm

PurpShell:
> This seems to load the IO more to all iterations with messageMedia, can you make this optional (and null by default)?

hmmmmm

PurpShell:
> This seems to load the IO more to all iterations with messageMedia, can you make this optional (and null by default)?

seems like it only stems from fromFilePath

wictornogueira (approved):
Looks good.
However, I would make fileSize default to 0 instead of null 🤔

Aliyss:
> Looks good.
> However, I would make fileSize default to 0 instead of null 🤔

why 0 instead of null?

aliyss (commented):
lgtm

aliyss (approved):

wictornogueira:
> > Looks good.
> > However, I would make fileSize default to 0 instead of null 🤔
> 
> why 0 instead of null?

'cause my guts said so :P
Just kidding... I just think it makes sense for a number to default to 0; probably because I tend to think of numbers as bytes...

Aliyss:
> 'cause my guts said so :P
> Just kidding... I just think it makes sense for a number to default to 0; probably because I tend to think of numbers as bytes...

@wictornogueira 

imo it should be null. Just in case a user really has a file that is 0 bytes. So it isn't confused with not having mentioned the amount of bytes regarding to having actually 0 bytes.

Other than that i don't really care about it too much. It is more a question of design and I'm open for the change.

tuyuribr:
> 
> 
> > 'cause my guts said so :P
> > Just kidding... I just think it makes sense for a number to default to 0; probably because I tend to think of numbers as bytes...
> 
> @wictornogueira
> 
> imo it should be null. Just in case a user really has a file that is 0 bytes. So it isn't confused with not having mentioned the amount of bytes regarding to having actually 0 bytes.
> 
> Other than that i don't really care about it too much. It is more a question of design and I'm open for the change.

I prefer defaulting to 0, even if you have a file that is 0 bytes that file will be invalid on waWeb context.

wictornogueira:
What about using -1?
I'm ok with null though.

src/structures/MessageMedia.js
```diff
@@ -10,40 +10,49 @@ const { URL } = require('url');
  * Media attached to a message
  * @param {string} mimetype MIME type of the attachment
  * @param {string} data Base64-encoded data of the file
- * @param {?string} filename Document file name
+ * @param {?string} filename Document file name. Value can be null
+ * @param {?number} filesize Document file size in bytes. Value can be null
  */
 class MessageMedia {
-    constructor(mimetype, data, filename) {
+    constructor(mimetype, data, filename, filesize = null) {
```
tuyuribr, line 17: Shouldn't this be 0 too ?

aliyss (dismissed):

Aliyss:
approved by mistake. Same as with tuyuribrs review -> ping me after the changes

src/structures/MessageMedia.js
PurpShell: Ouiiii

PurpShell:
> approved by mistake. Same as with tuyuribrs review -> ping me after the changes

changed

PR #676, opened 2021-06-01
Fixed documentation for Node version requirements & Bumped jsdoc from 3.6.6 to 3.6.7

purpshell:


pedroslopez:
Thanks for updating! However, we should wait until the release is out to regenerate documentation. This is mainly because docs are auto-published to the site, and generating it before the release means that people will see things in the docs that aren't available or relevant yet.

PurpShell:
I am a dumb human being at this point, i keep forgetting simple things

PR #677, opened 2021-06-02
acceptGroupV4Invite

Darker935:


index.d.ts
```diff
@@ -507,6 +508,8 @@ declare namespace WAWebJS {
         location: Location,
         /** List of vCards contained in the message */
         vCards: string[],
+        /** Invite v4 info */
+        invite: object
```
pedroslopez, line 512: We should offer better typings for this object and what it contains

src/Client.js
```diff
@@ -584,6 +584,20 @@ class Client extends EventEmitter {
...
     }
 
+    /**
+     * Accepts a private invitation to join a group
+     * @param {object} inviteInfo Invite V4 Info
+     * @returns {Promise<Object>}
+     */
+     async acceptGroupV4Invite(inviteInfo) {
+        if(!inviteInfo.inviteCodeV4) return 'Invalid invite code'
+        if (inviteInfo.inviteCodeExpV4 == 0) return 'Expired invite code'
```
pedroslopez, line 594: We should `throw` here instead of returning to indicate something went wrong

pedroslopez (changes requested):
Cool!

This makes me think of a couple questions:
- Do the old invites still work?
- How do these differ from the old invites?
- Could something be done to make the existing `acceptInvite` function work with the new format?

Also, it seems like typings are missing for the new `acceptGroupV4Invite` function.

JoseHM8A:
> 
> 
> Cool!
> 
> This makes me think of a couple questions:
> 
>     * Do the old invites still work?
> 
>     * How do these differ from the old invites?
> 
>     * Could something be done to make the existing `acceptInvite` function work with the new format?
> 
> 
> Also, it seems like typings are missing for the new `acceptGroupV4Invite` function.

1° - Yes, old invites are links, that anyone can join using them
2° - The InviteV4 is a private invite that only the user can use (it is received when the user does not allow to be added to groups by not contacts)
3° - The new method requires additional parameters, like expiration time, invite code, the user who received and who sent, that are in the message data

index.d.ts
JoseHM8A: Done (i think), this is my first PR 😅

pedroslopez (approved):

pedroslopez:
Thanks! Tested this and it worked great. Just made a few changes to the typings.

PR #683, opened 2021-06-04
Garbage Collector / Memory Optimization

### LINKED ISSUE #821: in multi sessions to many browser open or ram consume
Hi , 

i am using chrome for multi sessions its consume too many ram . can anyone help me for this solution with step by step please

tuyuribr:
Add method to collect garbage from waweb, trying to optimize ram consuptions. This leads to a CPU and Network overhead.
Maybe change .delete() to .deleteMsgsOlderThan(ts)

Pros:
Better Ram Consumption

Cons:
Cpu and Network overhead
Any Select queries will have aditional overhead ( Like fetching messages from a chat )
This enables a race condition ( When you process a message and fires GC at the same time )

It needs futher testing, any comments are welcome.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -201,10 +201,10 @@ class Client extends EventEmitter {
         if (this.options.memoryLogPath) {
             setInterval(async ()=>{
                 let memoryLine = await page.evaluate(() => {
-                    return Math.floor(Date.now() / 1000) + "," + window.performance.memory.totalJSHeapSize + "," + window.performance.memory.usedJSHeapSize + "\n";
+                    return Math.floor(Date.now() / 1000) + ',' + window.performance.memory.totalJSHeapSize + ',' + window.performance.memory.usedJSHeapSize + '\n';
                 });
                 Util.appendToLog(this.options.memoryLogPath,memoryLine); //this needs to be on Util because fs isn't here
-                }, 60000); // lets do each minute
+            }, 60000); // lets do each minute
         }
         
         // Check window.Store Injection
--- a/src/Client.js
+++ b/src/Client.js
@@ -32,6 +32,7 @@ const { ClientInfo, Message, MessageMedia, Contact, Location, GroupNotification 
  * @param {string} options.ffmpegPath - Ffmpeg path to use when formating videos to webp while sending stickers 
  * @param {boolean} options.bypassCSP - Sets bypassing of page's Content-Security-Policy.
  * @param {number} options.memoryOptimizationMs - Interval to run the pseudo garbage collector, 0 to disabled
+ * @param {string} options.memoryLogPath - File to log memory usage
  * 
  * @fires Client#qr
  * @fires Client#authenticated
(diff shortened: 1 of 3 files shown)


victormga:
i'm trying to deal with the excessive memory usage for months now, and there is some points i could input. 

- for what i've perceived, the main reason memory usage are media stuff. 
- by my benchmarks, even thousands of chats and messages uses only some mb of memory, the biggest usage i've came across was about 60mb (15.000+ chats).
- sure when loading more messages this usage will come up, but i don't know if its possible for only json/text/object data to amount to the 1.5gb of ram usage i've came across
- in comparison, sending a 2mb video to a chat, increases the mem usage by about 2mb.

i've tried to block the browser of caching media in ram, but with no success. the memory exploded anyway. i think maybe the blob url's genereated for the media blocks the gc, as it maybe count as a reference to the data...(?)
i'm no expert in chrome. maybe there's a way to disallow memory cache of media (or a way to release this memory/force the gc), or sink this data to disk.

anyway, i'm not sure how much it pays off deleting messages/chats since it increases processing/fetching times... but if it works it can be a solution.
do you have some stats about the efficacy of this solution?

tuyuribr:
> 
> 
> i'm trying to deal with the excessive memory usage for months now, and there is some points i could input.
> 
>     * for what i've perceived, the main reason memory usage are media stuff.
> 
>     * by my benchmarks, even thousands of chats and messages uses only some mb of memory, the biggest usage i've came across was about 60mb (15.000+ chats).
> 
>     * sure when loading more messages this usage will come up, but i don't know if its possible for only json/text/object data to amount to the 1.5gb of ram usage i've came across
> 
>     * in comparison, sending a 2mb video to a chat, increases the mem usage by about 2mb.
> 
> 
> i've tried to block the browser of caching media in ram, but with no success. the memory exploded anyway. i think maybe the blob url's genereated for the media blocks the gc, as it maybe count as a reference to the data...(?)
> i'm no expert in chrome. maybe there's a way to disallow memory cache of media (or a way to release this memory/force the gc), or sink this data to disk.
> 
> anyway, i'm not sure how much it pays off deleting messages/chats since it increases processing/fetching times... but if it works it can be a solution.
> do you have some stats about the efficacy of this solution?

Thanks for the input :D

Yes, media messages take up a lot of memory, the good thing about deleting MSG is that they contain cached media in the body (at least for images) so we have a benefit to delete it ( see bellow img )

image

About stats: I sent 10 100mb videos with and without the "garbage collector" with the gc the ram stabilizes way lower than without it. I'm going to make a commit here to enable "RAM USAGE TRACKING" so we could test this more efficiently

I think this solution isn't for everyone because it's only good if you are only receiving and sending messages ( if you query something you will have the overhead ), this solution seems very good to devices with more processing power than ram ( Like the raspberry pi 4 of 4gb )

tuyuribr:
Preliminary tests indicate that this solution is insufficient, in a short period of time there is a significant gain in memory usage, but after a few hours this function significantly degrades the memory usage.
I will test other solutions, I am open to suggestions

sourabhpatil1992:
Is it any possibility like our phone, don't download media or clear the last 7 days message time to time or clear chat and release memory.

PR #709, opened 2021-06-21
feat(dev): add new event page loaded

ricardoapaes:
Adding a new event when the page is loaded and you can interact with the DOM using the puppeter, below is a practical example:

> Adding a new div to the html for example:
> - Request login to your internal system before starting whatsapp.


PurpShell:
hey! i wouldn't get how this is effective? Does this allow editing of the UI? Maybe something we can accomplish differently. 

Thank you for your contributions. 

Best regards, 

[PurpShell](https://github.com/PurpShell)

JoseHM8A:
wow, i think this solved my problem getting the profile pictures of participants/groups.

Thank so much

ricardoapaes:
> hey! i wouldn't get how this is effective? Does this allow editing of the UI? Maybe something we can accomplish differently.
> 
> Thank you for your contributions.
> 
> Best regards,
> 
> [PurpShell](https://github.com/PurpShell)

@PurpShell 

Since I'm using the puppeteer in "headless: false" mode and I'm integrating with my web software, I'd like before opening whatsapp itself the user would login to my software so that I can get session information to be used later, so I add other html tags that contain the login form.

I couldn't find an easier way to do this since the events provided by the platform are only emitted after whatsapp is loaded.

If you have a better suggestion for my problem I am open to other opinions.

PurpShell:
> > hey! i wouldn't get how this is effective? Does this allow editing of the UI? Maybe something we can accomplish differently.
> > Thank you for your contributions.
> > Best regards,
> > [PurpShell](https://github.com/PurpShell)
> 
> @PurpShell
> 
> Since I'm using the puppeteer in "headless: false" mode and I'm integrating with my web software, I'd like before opening whatsapp itself the user would login to my software so that I can get session information to be used later, so I add other html tags that contain the login form.
> 
> I couldn't find an easier way to do this since the events provided by the platform are only emitted after whatsapp is loaded.
> 
> If you have a better suggestion for my problem I am open to other opinions.

looks good.

PurpShell:
unstale

src/Client.js
```diff
@@ -101,6 +101,8 @@ class Client extends EventEmitter {
             timeout: 0,
         });
 
+        this.emit(Events.PAGE_LOADED, page);
```
PurpShell, line 104:
```suggestion
        await page.waitForSelector('[class=web]');
        this.emit(Events.PAGE_LOADED, page);
```

purpshell (changes requested):
just a slight change

PR #720, opened 2021-06-29
Include Calls

tuyuribr:
Include call class, so we can have an event when called ( if we discover how to auto-reject calls, I will include here too )

Thanks to @QUISHOT (@JoseHM8A)


pedroslopez (changes requested):
Nice! Thanks for adding this.

A few things:
1. We should add the `on('incoming_call')` event to `index.d.ts`.
2. When calling to a group, i'm not seeing `isGroup` become true. How does `isGroup` work?
3. I'm noticing a message of type `call_log`. Maybe we should also handle these or add more attributes to the message object if they're available that have some more information about the call? I haven't looked into this much on whatsapp web itself, but I feel like there's probably some stuff we're missing there that we could include.

tuyuribr:
> 
> 
> Nice! Thanks for adding this.
> 
> A few things:
> 
>     1. We should add the `on('incoming_call')` event to `index.d.ts`.
> 
>     2. When calling to a group, i'm not seeing `isGroup` become true. How does `isGroup` work?
> 
>     3. I'm noticing a message of type `call_log`. Maybe we should also handle these or add more attributes to the message object if they're available that have some more information about the call? I haven't looked into this much on whatsapp web itself, but I feel like there's probably some stuff we're missing there that we could include.

1. Done! thanks for the heads up 
2. Apparently this is broken (WaWeb will return false no matter the conditions), but it should be fixed when they update more things about the call function
3. The call_log will return if is group or not ( this one works hahahha ) and if it's video or not. I think it isn't a good idea to implement the call_log because: It only works when the call is missed ( so I can have a call running for minutes before trigging something in the code ), and if we can implement something with calls ( like auto reject ) maybe will cause confusion having two places about call. Nevertheless if you think it's a good idea I can implement it

pedroslopez:
> 3. The call_log will return if is group or not ( this one works hahahha ) and if it's video or not. I think it isn't a good idea to implement the call_log because: It only works when the call is missed ( so I can have a call running for minutes before trigging something in the code ), and if we can implement something with calls ( like auto reject ) maybe will cause confusion having two places about call. Nevertheless if you think it's a good idea I can implement it

Are there any other properties on `call_log` messages that would be useful to expose in our Message class? Maybe something that at least indicates that it's a missed call?

tuyuribr:
> Are there any other properties on `call_log` messages that would be useful to expose in our Message class? Maybe something that at least indicates that it's a missed call?

If you get a 'call_log' it's because you missed the call, it doesn't appear if you answer the call. call_log expose if isGroup and if isVideo. That's why I think is better to only get onCall.

pedroslopez (approved):

PR #722, opened 2021-06-30
Add contact_change_state event. 

soncemvo:
With this change you can recieve "typing" or "recording" events of your contacts


tuyuribr:
I think that you are missing the "window.Store.Presence"

soncemvo:
> I think that you are missing the "window.Store.Presence"

No, its available in Store

tuyuribr:
> 
> 
> > I think that you are missing the "window.Store.Presence"
> 
> No, its available in Store

My bad, so the only thing to change is doublequotes to single quotes

soncemvo:
> > > I think that you are missing the "window.Store.Presence"
> > 
> > 
> > No, its available in Store
> 
> My bad, so the only thing to change is doublequotes to single quotes

done

purpshell (approved):

PurpShell:
@pedroslopez couldn't this be used to update the presence to be gotten by #553 ? You mentioned stuff about setTimeout, seems like storing the presence in some kind of variable and then accessing them by the function viable.

pedroslopez:
Thanks for contributing! This is something that's definitely been asked for in the past.

My concern with this is that, from previous experience, WhatsApp Web doesn't subscribe to the Presence of everybody, so the results of the events you're getting may be unexpected. You might want to listen for state changes of a particular user, but WhatsApp Web might not be subscribed to their presence.

So, this brings up a few questions for me: Has this changed? If not, could we instead write a function to subscribe to someone's presence? How many contacts does WhatsApp Web subscribe to by default?

fatihmesse:
I use the this function, at startup is it working but later than throw all chat status and stop working. How can ı fix it?
Ekran Resmi 2022-10-30 15 59 12

PR #725, opened 2021-07-05
Add Operating System information for bug reports.

purpshell:
Adds os info to bug reports for better diagnosing.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -33,6 +33,8 @@ If applicable, add code snippets to help explain your problem.
  - NodeJS version (`node -v`):
  - npm or yarn version (`npm -v`):
  - whatsapp-web.js version:
-
```
pedroslopez, line 36: I think you're missing a line break here- if you look at the preview it seems like "Other" is added to the `whatsapp-web.js version` line

pedroslopez (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
PurpShell: seems correct

pedroslopez (approved):

PR #726, opened 2021-07-05
fix "TimeoutError: waiting for selector `canvas` failed:..."

victormga:
fixed "TimeoutError: waiting for selector `canvas` failed: timeout 45000ms exceeded"

explanation: in node, loops as setTimeout or setInterval runs by default on the global scope, so you can't catch exceptions easily.
this catches and ignores the timeout exception

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -120,25 +120,29 @@ class Client extends EventEmitter {
 
         } else {
             const getQrCode = async () => {
-                // Check if retry button is present
-                var QR_RETRY_SELECTOR = 'div[data-ref] > span > button';
-                var qrRetry = await page.$(QR_RETRY_SELECTOR);
-                if (qrRetry) {
-                    await qrRetry.click();
-                }
+                try {
+                    // Check if retry button is present
+                    var QR_RETRY_SELECTOR = 'div[data-ref] > span > button';
+                    var qrRetry = await page.$(QR_RETRY_SELECTOR);
+                    if (qrRetry) {
+                        await qrRetry.click();
+                    }
 
-                // Wait for QR Code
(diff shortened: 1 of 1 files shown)


pedroslopez:
Hey victor, thanks for contributing.

I don't think that this exception should be ignored, and the timeout used is configurable. If this is happening consistently, maybe we should instead consider incrementing the timeout or trying to identify in which cases it takes longer than expected.

victormga:
hi, pedro. thanks for the reply.
let me try to explain further.

the main reason is:
some versions of node crashes and exit your app on a non-handled exceptions. 
in a situation where i'm trying to read the qrcode, but it fails (lets say the phone wasn't ready or the user is afk), i'm just wanna close the pupPage graciously and try again later when the user is ready to read the qrcode. (basically .destroy and later .initialize again).

but the unhandled exception crashes the whole process exiting the application.
in a localhost situation where the user can simply reopen the program, this is hardly a problem, but is a weird situation in a vps or container context.

not totally related: 
this setInterval runs every 20 seconds, and the canvas timeout is by default set to 45 seconds. so this may cause a problem with multiple calls while to old call is still waiting (leaks?). maybe it would be good to set the default timeout to 20 seconds, the same as the setInterval?
also it would be to find a way to handle this exceptions outside the global context, maybe dispatching and event on client (client.on("error"... ?)

i made this changes on my production code, and it works fine. but it would be nice if this qol fixes could be in the main lib.
didn't include on this pr because i'm trying to organize the prs each one with it's own improvement.

victormga:
actually i realised i've made a confusion.
i'll review the code and create another pr if necessary.

PR #727, opened 2021-07-06
Add the option to use "browserWSEndpoint"

### LINKED ISSUE #753: Access a remote instance of chrome
Hi, I need to know how to configure the puppeteer to access a vps where the chromium is.

tegila:
puppeteer using launch is heavy and very inconvenient when used in docker environments.

puppeteer-core connect is much lower profile as the installation of chrome or others will
take place outside the npm scope.

I did this small try with some simple code and it worked fine:

```
const { Client } = require('whatsapp-web.js');

const host = 'localhost';
const port = 3000;

const client = new Client({ 
    puppeteer: {
        browserWSEndpoint: `ws://${host}:${port}`
    }
});

client.on('qr', (qr) => {
    // Generate and scan this code with your phone
    console.log('QR RECEIVED', qr);
});

client.on('ready', () => {
    console.log('Client is ready!');
});

client.on('message', msg => {
    if (msg.body == '!ping') {
        msg.reply('pong');
    }
});

client.initialize();
```
index.js

```
version: '3.5'

services:  
  browserless:
    restart: always
    image: browserless/chrome:1.38-chrome-stable
    environment:
      - PREBOOT_CHROME=true
      - KEEP_ALIVE=true
      - CHROME_REFRESH_TIME=86400000
      - MAX_CONCURRENT_SESSIONS=1
      - MAX_QUEUE_LENGTH=20
      - CONNECTION_TIMEOUT=600000
      - ENABLE_CORS=true
    ports:
      - 3000:3000
```
docker-compose.yaml

This simple setup and you are ready to go.

My pull request is just a PoC


tuyuribr:
This doesn't seem to be the best approach, because puppetter-core doesn't work out of the box ( doesn't install chromium ). Sure it would be nice to downsize everything, but if you are scalling things you now how to make this work and optimize the lib, but if you are new programmer you would have more problems than solutions.

tegila:
Until puppetter-core I thought that was the case. 

But for now it's not convenient to have a browser installed by npm many times in machines that near 100% of the times already have a browser installed.

I think it's only a case of patching the README.md to include the references about how to run
chrome with proper settings to enable puppetter-core connect to it for Example usage and one scenario with docker-compose following the basic instructions.

If you agree I could do it myself and push another request.

tuyuribr:
Push your pach to the README, but I still disagree with this solution, it will break some workflows. I will leave to other people review this.

pedroslopez:
puppeteer-core wasn't available when I started this project, but if it was I would have probably used it instead of puppeteer proper. I've considered switching over to puppeteer-core, but it would be a breaking change for the many people that are using this library. Maybe we can do this when some sort of version 2 of the library is released, but I don't think it's as important to implement at the moment. 

You can skip the browser download when installing puppeteer by setting the environment variable `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to `true`, and it solves the main concern with switching to puppeteer-core.

tegila:
Hi @pedroslopez,

Thank you for all the work you have done in this project.

I already know and used `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` but there is another problem in the source code as it's using puppeteer.launch instead of puppeteer.connect, so this *require* a local install.

Maybe a small patch can satisfy everybody needs like for example using a conditional `if` when `options.puppeteer.browserWSEndpoint` is present and using .connect instead of .launch.

Using this aproach we can help others who want have remote chrome and also those who like to have it at local path (current) and it also will help migrating, in the future, to -core.

remote example: 
```
const { Client } = require('whatsapp-web.js');

const host = 'localhost';
const port = 3000;

const client = new Client({ 
    puppeteer: {
        browserWSEndpoint: `ws://${host}:${port}`
    }
});
```

PurpShell:
Hmm, after thinking about this change it would rather be an optional thing...

tegila:
@PurpShell:

> Maybe stick with that and `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to do your magic.. Doesn't seem too friendly to have you manually install chromium because I'm pretty sure a lot of people don't know it's chromium, and don't care to know... Plus people could end up using old versions and potentially break stuff.. (or the other way around when puppeteer breaks cuz of new chrome updates?)

`PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` was a previous solution before puppeteer-core was released. It was nice, but as you told, after that, with the current source code, you have to use your own chrome locally installed. 

I had used it many times but when we get ready to ***deploy and scale*** we have to fix the code ourselves.
 
> Edit; i just realized this may also raise privacy concerns (devs are not comfortable using chrome/chromium, and now they have to provide their local installations??)

Well, give a look to new readme.md where I suggest the installation of browserless using docker.
It's very straight forward to developer this days and will keep the privacy way better than npm can do.

I suggest the enforcement of this path as it's more adapted to the current developer skill set.

Also if others agree, we can also clean the input json of `new Client` to get straight through **browserWSEndpoint** .

Waiting for replies to see where to move next.

tuyuribr:
Don't forget that chromium is available in all devices ( I'm looking at you arm devices ), and from my experience we do a npm install on a clean instalation and we are ready to go, we don't have any browser installed in our fresh ubuntu setups ( ubuntu server )

PurpShell:
> @PurpShell:
> 
> > Maybe stick with that and `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to do your magic.. Doesn't seem too friendly to have you manually install chromium because I'm pretty sure a lot of people don't know it's chromium, and don't care to know... Plus people could end up using old versions and potentially break stuff.. (or the other way around when puppeteer breaks cuz of new chrome updates?)
> 
> `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` was a previous solution before puppeteer-core was released. It was nice, but as you told, after that, with the current source code, you have to use your own chrome locally installed.
> 
> I had used it many times but when we get ready to _**deploy and scale**_ we have to fix the code ourselves.
> 
> > Edit; i just realized this may also raise privacy concerns (devs are not comfortable using chrome/chromium, and now they have to provide their local installations??)
> 
> Well, give a look to new readme.md where I suggest the installation of browserless using docker.
> It's very straight forward to developer this days and will keep the privacy way better than npm can do.
> 
> I suggest the enforcement of this path as it's more adapted to the current developer skill set.
> 
> Also if others agree, we can also clean the input json of `new Client` to get straight through **browserWSEndpoint** .
> 
> Waiting for replies to see where to move next.

What i meant is that to scale you don't have to use puppeteer core, i think we should push our efforts to pull #590 and this is a safe middle-ground. (skip chromium via npm, use browserWSEndpoint)

pedroslopez:
> Hi @pedroslopez,
> 
> Thank you for all the work you have done in this project.
> 
> I already know and used `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` but there is another problem in the source code as it's using puppeteer.launch instead of puppeteer.connect, so this _require_ a local install.
> 
> Maybe a small patch can satisfy everybody needs like for example using a conditional `if` when `options.puppeteer.browserWSEndpoint` is present and using .connect instead of .launch.
> 
> Using this aproach we can help others who want have remote chrome and also those who like to have it at local path (current) and it also will help migrating, in the future, to -core.
> 
> remote example:
> 
> ```
> const { Client } = require('whatsapp-web.js');
> 
> const host = 'localhost';
> const port = 3000;
> 
> const client = new Client({ 
>     puppeteer: {
>         browserWSEndpoint: `ws://${host}:${port}`
>     }
> });
> ```

@tegila Thanks for providing more context! It's super helpful.

I agree with what's being discussed, and instead of moving to puppeteer-core at this point, support `.connect` if the `browserWSEndpoint` is passed instead of `.launch`. It does seem like there was a PR that did something similar, but the author seems to be inactive. Feel free to takeover and implement it in this PR if you want to.

purpshell (approved):
.

tegila (commented):
I really prefer this way with only 3 file changes than using the puppeteer as every single developer this days should have docker installed.

Anyway I did the requested changes in the oldest commit using the if...else statement to check if this.options.puppeteer.browserWSEndpoint is defined, but I'm not sure if this.options will always be a {}.

Please review both but I left my vote for this review using ***puppeteer-core and browserless***. Last week I held whatsapp web running for 3 days without issues.

tuyuribr:
> as every single developer this days should have docker installed.
Let's not start a war, but this is false.


> Anyway I did the requested changes 
LGTM, But I don't feel confortable _"You need to connect to any previously existing browser instance"_


> Last week I held whatsapp web running for 3 days without issues.
Can you give more details about the memory consumption and your usage ? ( I have isntances that run for weeks without issues because they don't send a lot of msgs ). 

For the memory consumption, it would be nice to have heap consumption and available memory in the system PR #683  has heap consumption

README.md
```diff
@@ -13,11 +13,37 @@ The module is now available on npm! `npm i whatsapp-web.js`
 
 Please note that Node v14+ is required.
 
+## Requirements
```
Aliyss, line 16: Don't get me wrong. I really enjoy the pro-docker and puppeteer-core changes here. I also believe we should move toward puppeteer-core in the future.

However at this point and since puppeteer is a dependency, I think the Requirements should clearly be listed as optional. And moved closer to the end of the setup and example.

I would structure it this way, following the order in complexity

- Normal Usage
  - Example Normal
- Websocket
  - Changes | only add the changes necessary not the full example
- Docker
  - Setup Docker + Changes | only add the changes necessary not the full example

Extra (if you want to):
You can add the browserWSEndpoint option as a comments in the example.js file.

aliyss (changes requested):
Don't get me wrong. I really enjoy the pro-docker and puppeteer-core changes here. I also believe we should move toward puppeteer-core in the future.

However at this point and since puppeteer is a dependency, I think the Requirements should be either structured as written in my comments or listed as optional.

README.md
tegila: It's done as I got your requests but let me keep clear that still prefer the initial aproach directly. It's more simple and straight to the point.

tegila:
> > as every single developer this days should have docker installed.
> 
> Let's not start a war, but this is false.

Ok, peaceful agreement here.

> > Anyway I did the requested changes
> 
> LGTM, But I don't feel confortable with _"You need to connect to any previously existing browser instance"_

Ok, I had changed that, now you don't **NEED**, but could also just run simple code and u're ready. But you'll have to wait a long time until your npm resolve and npm is also known to be very slow at regular use, and installing a browser go way beyond the `regular use`.

> > Last week I held whatsapp web running for 3 days without issues.
> 
> Can you give more details about the memory consumption and your usage ? ( I have isntances that run for weeks without issues because they don't send a lot of msgs ).

Yes, sure. They are not sending a lot of msgs, just passing the messages and pictures through to MQTT where the action happens. And sure it's very power hungry but browserless has many methods to even restart the instance if the memory consumption goes out-of-control.

> For the memory consumption, it would be nice to have heap consumption and available memory in the system PR #683 has heap consumption

I don't have the tools installed at the machine to make this charts as it's only running for testing purposes yet.

purpshell (approved):
LGTM

PurpShell:
changed the title since the contents of the PR changed

aliyss (approved):

PR #737, opened 2021-07-09
Fix downloading media for WhatsApp Web v2.2126.10

tuyuribr:


pedroslopez:
The function shouldn't have been async- fixed in 1.13.2

PR #762, opened 2021-07-15
Update node-webpmux 2.x syntax to 3.x (support upstream breaking change)

ApeironTsuka:
dependabot bumped you to 3.0.0 apparently ignoring that major version bumps _may_ contain breaking changes. In this case, it did. This _should_ be the last breaking change I make to node-webpmux for quite some time.

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -183,9 +183,9 @@ class Util {
             let jsonBuffer = Buffer.from(JSON.stringify(json), 'utf8');
             let exif = Buffer.concat([exifAttr, jsonBuffer]);
             exif.writeUIntLE(jsonBuffer.length, 14, 4);
-            await img.loadBuffer(Buffer.from(webpMedia.data, 'base64'));
+            await img.load(Buffer.from(webpMedia.data, 'base64'));
             img.exif = exif;
-            webpMedia.data = (await img.saveBuffer()).toString('base64');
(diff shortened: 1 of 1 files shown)


PurpShell:
Hey Apeiron,  while you're here, can I ask about private methods?? How are they used in node-webpmux, and are they required for   functionality? Thanks for your library, underrated blessing

PurpShell:
I asked for this since there has been an explosion of people that don't have node 14+, and are questioning the issue. Node 12 is also in LTS, and is preferred by package managers for Linux distros, and Node 13+ require Windows 8+ to install.

ApeironTsuka:
They're more of a convention. An easy way to make sure that internal functions _stay_ internal. The library would work fine if you made all of them public... I don't think I have any naming conflicts with them. I wasn't aware 13+ needed Win8+ .. that's rather silly. I'll drop their usage for a 3.1.0 in a day or two, as I had figured with 14 being LTS would mean a wider adoption, but you're right - Debian is also still on 12.21.0. Honestly hadn't noticed since I don't use the system version. And 13+ needing Win8+ is also a bit of a problem.

PurpShell:
good news! Thanks again for your time and effort, hope you get recognized more!

pedroslopez:
Thanks for coming in here and fixing the compatibility issues!

PR #769, opened 2021-07-17
Add fromUrl method

### LINKED ISSUE #705: Send media with sendMessage by media URL

### LINKED ISSUE #771: cannot send media
Hello,

I just tried your api, this could be send message as well (according on example.js), but how to send media, because when I tried to send media like this : 

```
client.on('ready', () => {
	const media = new MessageMedia(
        'image/png', 
        'https://homepages.cae.wisc.edu/~ece533/images/boat.png'
    );

    client.sendMessage("614xxxxxxxxx@c.us",media).then((r) => {
        console.log("sendMessage",r)
    })
});
```

it always return error :  _Evaluation failed: DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded._

Please help me

Thanks

wictornogueira:
Does basically the same thing as fromFilePath but takes a url instead

### DIFF
--- a/src/structures/MessageMedia.js
+++ b/src/structures/MessageMedia.js
@@ -82,9 +82,7 @@ class MessageMedia {
                 });
                 data = btoa(data);
             }
-
-            console.log(mime);
-
+            
             return { data, mime };
         }
 
--- a/src/structures/MessageMedia.js
+++ b/src/structures/MessageMedia.js
@@ -49,30 +49,53 @@ class MessageMedia {
     /**
      * Creates a MessageMedia instance from a URL
      * @param {string} url
-     * @param {number} [sizeLimit=0]
+     * @param {Object} [options]
+     * @param {number} [options.unsafeMime=false]
+     * @param {object} [options.client]
+     * @param {object} [options.reqOptions]
+     * @param {number} [options.reqOptions.size=0]
      * @returns {Promise<MessageMedia>}
      */
-    static fromUrl(url, sizeLimit = 0) {
-        return new Promise((resolve, reject) => {
-            const pUrl = new URL(url);
-            const mimetype = mime.getType(pUrl.pathname);
+    static async fromUrl(url, options = {}) {
+        let mimetype;
 
-            if (!mimetype) {
-                return reject(new Error('Unable to determine MIME type'));
+        if (!options.unsafeMime) {
+            const pUrl = new URL(url);
+            mimetype = mime.getType(pUrl.pathname);
+
+            if (!mimetype)
+                throw new Error('Unable to determine MIME type');
+        }
+
+        async function fetchData (url, options) {
+            const reqOptions = Object.assign({ headers: { accept: 'image/* video/* text/* audio/*' } }, options);
+            const response = await fetch(url, reqOptions);
+            const mime = response.headers.get('Content-Type');
+            let data = '';
+
+            if (response.buffer) {
+                data = (await response.buffer()).toString('base64');
+            } else {
+                const bArray = new Uint8Array(await response.arrayBuffer());
+                bArray.forEach((b) => {
+                    data += String.fromCharCode(b);
+                });
+                data = btoa(data);
             }
 
-            fetch(url, { size: sizeLimit, headers: { accept: 'image/* video/* text/* audio/*' } })
-                .then(async res => {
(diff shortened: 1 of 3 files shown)


Gimenz (approved):

PurpShell:
Couldn't we use `http` instead of `node-fetch` here? Also you haven't updated `package.json`

wictornogueira:
> Couldn't we use `http` instead of `node-fetch` here? Also you haven't updated `package.json`
> 
> also why are you wrapping the entire function in a `Promise`? Wouldn't most of that job be done if you used `async` and `await fetch()` instead of `fetch().then`
> 
> Also you are determining the mime-type via the URL itself, instead of using `mime`, which i think is less relevant in today's standards, maybe get the data via `node-fetch`'s response "content-type" header

1. I would have to write a wrapper in order to use http, which would be a bit overkill
2. I haven't added any dependency to the dependency tree; `node-fetch` is a depency of `puppeteer`
3. Using Promise makes things easir, beucase It allows me to easly resolve it by calling resolve
4. Promises also allow you to raise an exception from anywhere inside its scope
5.
  5.1. Some server still respond with `application/octet-stream`, which is not exactly ideal
  5.2 Altho I've specified the `Content-type` I would accept, it's up to the server to respect it or not
  5.3 Not even browsers trust `Content-type`
  5.4 Saves some time and bandwidth by only downloading something that is more likely to work
  5.5 And I know all this comes with the tradeo-off of not being able to use weird urls

Gimenz:
how about request config options?

wictornogueira:
> how about request config options?

What do you mean?

Gimenz:
> > how about request config options?
> 
> What do you mean?

request options..
like headers, user-agent, cookie.. etc

wictornogueira:
> > > how about request config options?
> > 
> > 
> > What do you mean?
> 
> request options..
> like headers, user-agent, cookie.. etc

How does it help?

Gimenz:
> > > > how about request config options?
> > > 
> > > 
> > > What do you mean?
> > 
> > 
> > request options..
> > like headers, user-agent, cookie.. etc
> 
> How does it help?

maybe, it will help if some cases need headers

for example :

```javascript
MessageMedia.fromUrl('url', 0, { 
    headers: {
        Cookie: 'some cookie',
        'User-Agent': 'ua'
    }
})
```

like this [getDUrl](https://github.com/open-wa/wa-automate-nodejs/blob/5b7f2406f7d52a538ffd85c3513a24c666103d45/src/utils/tools.ts#L63)

wictornogueira:
Ohhh... yeah, indeed.

src/structures/MessageMedia.js
```diff
@@ -43,6 +45,35 @@ class MessageMedia {
...
+     * @param {number} [sizeLimit=0]
+     * @returns {Promise<MessageMedia>}
+     */
+    static fromUrl(url, sizeLimit = 0) {
+        return new Promise((resolve, reject) => {
+            const pUrl = new URL(url);
+            const mimetype = mime.getType(pUrl.pathname);
+
+            if (!mimetype) {
+                return reject(new Error('Unable to determine MIME type'));
```
tuyuribr, line 61: I think that rejecting when you are unable to get he mime type would be an issue, how about a param "ignoreMime" ?

PurpShell:
> Ohhh... yeah, indeed.

yeah also some users are using a proxy keep that in mind

wictornogueira:
You can now skip the initial mime checking thing, use proxy, override request options and choose context

purpshell (approved):
LGTM

aliyss (approved):
lgtm

iriobf:
How to set the filename in the fromUrl method?
Media being sent with filename null.
```javascript
const media = await MessageMedia.fromUrl(fileUrl)
const msgMedia = await sessions[from].sendMessage(numberTo, media)
```

wictornogueira:
```js
media.filename = 'whatever'
```

d0v3riz:
> ```js
> media.filename = 'whatever'
> ```

Can I get the original filename from that url?

iriobf:
> > ```js
> > media.filename = 'whatever'
> > ```
> 
> Can I get the original filename from that url?

media.filename = new URL(url).pathname.split('/').pop();

PR #770, opened 2021-07-18
Add getBlockedContacts method

Gimenz:
Simple method to get blocked ids

```javascript
const blockedIds = await client.getBlockedIds();
//['62813xxx@c.us', '628xxx@c.us']
```
it will return empty array if no one contact is blocked


src/Client.js
```diff
@@ -931,6 +931,18 @@ class Client extends EventEmitter {
...
+    /**
+     * Get all blocked ids by Host account
+     * @returns {Promise<Array>}
+     */
+    async getBlockedList() {
+        let chatIds = await this.pupPage.evaluate(() => {
+            return Object.keys(window.Store.Blocklist._index);
+        });
+
+        return chatIds
```
tuyuribr, line 944: Missing semicolon

pedroslopez (commented):
Nice! Thanks for contributing.

I would suggest being a bit clearer on what kind of ID you're returning, for example `getBlockedContactIds`.

Better yet, it may be more useful to return contacts themselves instead of just their IDs (`getBlockedContacts()`).

purpshell (approved):
fixed a bunch of stuff for you, hope you don't mind. It was simple changes, and I didn't it to be delayed. anyways, looks good to me.

Gimenz:
> fixed a bunch of stuff for you, hope you don't mind. It was simple changes, and I didn't it to be delayed. anyways, looks good to me.

thank you, I'm not very careful

pedroslopez (approved):
Added some tests and fixed it up a bit, thanks for contributing!

PR #775, opened 2021-07-18
Fix: Contact Block Return boolean (Issue: #767)

### LINKED ISSUE #767: Contact.block() does not return boolean value.
### Bug description

Contact.block() does not return boolean value if contact is not blocked. However, it does return a boolean if the contact is already blocked.

### Reproduction steps

1. Get any contact by using client.contactById(number) method. 
2. run the block method on the contact instance.
3. logging the result of block would not return a boolean value.

### Expected behavior 

Expected to get a boolean from block method when contact is not blocked.

### Relevant code
```
const contact = await client.getContactById(<contact number>)
console.log(await contact.block());
```

### Environment (please complete the following information):
**WhatsApp**
 - Account type [Standard / Business]: Standard
 - Device OS [iOS / Android]: Android
 - WhatsApp Web version [run `await client.getWWebVersion()`]: 2.2126.11

**Library**
 - Browser [Chrome / Chromium]: Chromium
 - NodeJS version (`node -v`): 14.16.1
 - npm or yarn version (`npm -v`): 6.14.12
 - whatsapp-web.js version: 1.13.2

felixic:
Added boolean to return if contact is blocked

### DIFF
--- a/src/structures/Contact.js
+++ b/src/structures/Contact.js
@@ -143,7 +143,7 @@ class Contact extends Base {
             await window.Store.BlockContact.blockContact(contact);
         }, this.id._serialized);
 
-        return this.isBlocked ? true : false;
+        return this.isBlocked;
     }
 
     /**
--- a/src/structures/Contact.js
+++ b/src/structures/Contact.js
@@ -140,7 +140,7 @@ class Contact extends Base {
 
         await this.client.pupPage.evaluate(async (contactId) => {
             const contact = window.Store.Contact.get(contactId);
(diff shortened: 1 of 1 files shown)


tuyuribr (commented):
Commented about this in the #767 

* The logic is invalid
* the code should be cleaner
* will break workflows because someone will try to block groups

felixic:
Update Pull Request #775 pushed

tuyuribr:
I think we shouldn't patch, we should expect waweb fix this bug

PR #784, opened 2021-07-20
add more tests

pedroslopez:
Adds more testing around the client and Chat model.

### DIFF
--- a/tests/client.js
+++ b/tests/client.js
@@ -194,13 +194,16 @@ describe('Client', function() {
             this.timeout(35000);
             client = helper.createClient({withSession: true});
             await client.initialize();
-
-            const version = await client.getWWebVersion();
-            console.log(`WA Version: ${version}`);
         });
 
         after(async function () {
             await client.destroy();
+        });
+
+        it('can get current WhatsApp Web version', async function () {
+            const version = await client.getWWebVersion();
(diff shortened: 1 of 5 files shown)


tests/client.js
```diff
@@ -413,5 +416,48 @@ END:VCARD`;
...
+
+        describe('Status/About', function () {
+            it('can set the status text', async function () {
+                this.timeout(5000);
+
+                await client.setStatus('My shiny new status');
+
+                const me = await client.getContactById(client.info.wid._serialized);
+                const status = await me.getAbout();
+                expect(status).to.eql('My shiny new status');
```
PurpShell, line 449: set the status to what it was before the test?

tests/client.js
```diff
@@ -413,5 +416,48 @@ END:VCARD`;
...
+            });
+
+            it('can set the status text to something else', async function () {
+                this.timeout(5000);
+
+                await client.setStatus('Busy');
+                
+                const me = await client.getContactById(client.info.wid._serialized);
+                const status = await me.getAbout();
+                expect(status).to.eql('Busy');
```
PurpShell, line 459: same as above

purpshell (changes requested):
some changes

PR #804, opened 2021-07-31
replaced sharp with webp converter because sharp doesent work on aarch64

Glowman554:
This pull request enabels aarch64 / arm64 users to use this libary because sharp faile's to install on those systems with the following error:

```
npm ERR! In file included from ../src/common.cc:24:
npm ERR! /usr/include/vips/vips8:35:10: fatal error: glib-object.h: No such file or directory
npm ERR!  #include <glib-object.h>
npm ERR!           ^~~~~~~~~~~~~~~
npm ERR! compilation terminated.
npm ERR! make: *** [sharp.target.mk:139: Release/obj.target/sharp/src/common.o] Error 1
```

webp converter also doesen't work right from the begining but it is way easier to fix with the following script:

```bash
wget https://storage.googleapis.com/downloads.webmproject.org/releases/webp/libwebp-1.2.0.tar.gz
tar -xvf libwebp-1.2.0.tar.gz
cd libwebp-1.2.0
./configure --prefix=$PWD/../node_modules/webp-converter/bin/libwebp_linux/ --disable-shared
make -j 4
make install
cd ..
rm -rfv libwebp-1.2.0*
```


tuyuribr:
So, I'm running on raspberry pi 4 (arm64), and don't have any issue with any dependencies, is just npm --install and we are good to go. I have information that people are running this on M1 chips too, so it seems arm64 is good.
Could you provide more information? A new sharp update broke something?

Glowman554:
I'm running it on a raspberry pi 4 too with the official raspberry pi os 64 bit and I can't get the problem fixed in any way I tried

Glowman554:
I talked about the issue with some people and they said it's basicly impossible to use sharp on arm64

purpshell (changes requested):
Hey dude! I have tried to push for  `webp-converter` in the past, as seen in #527, but simply put:
1. 33MB raw file size, really bad
2. uses binaries, not reliable.

Glowman554:
If you know a better libary i would love to hear about it my only problem is i cant get sharp to work

Glowman554:
I think this pull request is kinda useless because I'm the only person having this issue so I'm going to close it

Issue #808, opened 2021-08-04
File name is null when using method MessageMedia.fromUrl()


d0v3riz:
I'm trying a new feature, when using function `MessageMedia.fromUrl()` successfully sends a media message, but the name for the document file is always **null**. Is this a bug ?

![photo_2021-08-04_08-19-05](https://user-images.githubusercontent.com/23495597/128106741-1386bbf8-91ba-4715-b63b-f3afe189d51f.jpg)

My code is

```
sendMediaFromUrl: async (phone, path) => {
	let number = convertNumber(phone)
	const media = await MessageMedia.fromUrl(path)
	client.sendMessage(number, media)
	console.log('+ Media dari url berhasil dikirim ke ' + number)
	return media
}
```

PurpShell:
@wictornogueira

PurpShell:
Hey! I've taken a look at the PR that added `MessageMedia.fromUrl()` (#769) and you can see that it ignores the file name: 
![image](https://user-images.githubusercontent.com/52720753/128206329-3dfe48aa-3401-4f7e-9ada-c97d722ea5a8.png)

@wictornogueira i hope you can rectify this, lol

PR #811, opened 2021-08-05
update MessageMedia.fromUrl

wictornogueira:
- fromUrl will now set filename based on the url (#808)
- Content-type is now used as fallback


purpshell (approved):
lgtm

tuyuribr (approved):
lgtm

tuyuribr:
Security warning:
`"http://url/file/..\\..\\..\\..\\..\\Windows\\write.exe".split('/').pop();`
Evals to "..\\..\\..\\..\\..\\Windows\\write.exe"
If someone uses MessaMedia.fileName on windows it can break something
But the risk is close to none

tuyuribr:
Warning: if you download something from an empty path ( https://image.somedomain.com/ ) fileName will be empty. but the chances of this happening seems very low

aliyss (approved):

PR #819, opened 2021-08-07
Update Message Media.

codizen:
- set filename using option or from url path.

### DIFF
--- a/src/structures/MessageMedia.js
+++ b/src/structures/MessageMedia.js
@@ -59,13 +59,10 @@ class MessageMedia {
      */
     static async fromUrl(url, options = {}) {
         let mimetype;
-        let filename = null
 
-        if (!options.filename) {
-            filename = new URL(decodeURIComponent(url)).pathname.split('/').pop();
-        } else {
-            filename = options.filename
-        }
+        const filename = options.filename?
+            options.filename
+            : new URL(decodeURIComponent(url)).pathname.split('/').pop();
 
         if (!options.unsafeMime) {
             const pUrl = new URL(url);
--- a/src/structures/MessageMedia.js
+++ b/src/structures/MessageMedia.js
@@ -50,6 +50,7 @@ class MessageMedia {
      * Creates a MessageMedia instance from a URL
      * @param {string} url
      * @param {Object} [options]
+     * @param {string} [options.filename]
      * @param {number} [options.unsafeMime=false]
      * @param {object} [options.client]
      * @param {object} [options.reqOptions]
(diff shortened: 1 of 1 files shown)


src/structures/MessageMedia.js
```diff
@@ -58,6 +59,13 @@ class MessageMedia {
      */
     static async fromUrl(url, options = {}) {
         let mimetype;
+        let filename = null
```
tuyuribr, line 62: Missing semicolon

src/structures/MessageMedia.js
```diff
@@ -58,6 +59,13 @@ class MessageMedia {
...
+
+        if (!options.filename) {
+            filename = new URL(decodeURIComponent(url)).pathname.split('/').pop();
+        } else {
+            filename = options.filename
```
tuyuribr, line 67: Missing semicolon

src/structures/MessageMedia.js
tuyuribr: Maybe doesnt need to be set to null

src/structures/MessageMedia.js
```diff
@@ -58,6 +59,13 @@ class MessageMedia {
      */
     static async fromUrl(url, options = {}) {
         let mimetype;
+        let filename = null
+
+        if (!options.filename) {
```
tuyuribr, line 64: this can be simplified

src/structures/MessageMedia.js
codizen: Ok, thanks to this check.

src/structures/MessageMedia.js
codizen: Ok, thanks to this check.

tuyuribr:
Thanks for your contribution, but I think #811 resolve this issue, right?

src/structures/MessageMedia.js
codizen: Do you want to simplify in one line?

codizen:
> Thanks for your contribution, but I think #811 resolve this issue, right?

Yes similar, but have a small difference.
In this code, we can pass filename in options and clean filename if get from url path.

PurpShell:
> > Thanks for your contribution, but I think #811 resolve this issue, right?
> 
> Yes similar, but have a small difference.
> In this code, we can pass filename in options and clean filename if get from url path.

Hey! I appreciate you coming here and taking the effort to open a new PR. Please coordinate your efforts, this is not a marketplace. You can open a PR in a fork, so maybe port your changes there? As for setting a filename by an option, you can simply just change the property on the MessageMedia object.

```js
let media = await MessageMedia.fromURL("https://abcd.xyz");
media.filename = "test";
```

PurpShell:
Update: Seems like #811 has implemented the options you were looking for. I'll close this PR, hopefully you continue contributing there!

PR #820, opened 2021-08-07
[REFERENCE] Fix doc getChat

### LINKED ISSUE #818: Docs for the return of the promise of the getChat function is incorrect
The documentation for the return of the promise of the getChat function is incorrect, because the function can return a Promise<Chat | GroupChat | PrivateChat>. In the documentation informs you that promise only returns Promise<Chat>. This can give typescript typescript errors when trying to access them.

image

image

image

image

image

tuyuribr:
For reference #818


tuyuribr:
Bad Pr

PR #857, opened 2021-09-01
Fix error Evaluation failed: TypeError: this._find is not a function

ahmdaka06:


pedroslopez:
Fixed in latest release. I was going to merge this one in, but it looks like there are old commits on here. Thanks though!

PR #866, opened 2021-09-07
Added MessageMedia.toFilePath function

purpshell:
In this pull request I add a function to MessageMedia to save files, since this has apparently became a headache, and I though having a function that does this by default is fitting. 

Example usage:
```js
const media = await msg.downloadMedia();
await media.toFilePath('/home/purpshell/Documents/message') // the code adds the extension for you (if you enter a directory we will use the filename property (if it's null then we'll throw an error))
```


PurpShell:
.

PR #855, opened 2021-08-31
Type file improvements

purpshell:
Updated the types so:
  - Buttons and Lists don't throw errors when sending a message (MessageContent)
  - puppeteer: args and puppeteer: headless don't throw an error

PR #868, opened 2021-09-08
fix: Use correct type for puppeteer launch options

andreroggeri:
On typescript is not possible to pass `puppeteer: { headless: false }` on the `Client`constructor


PurpShell:
Hey! #855 has been opened, we could add this there. Thanks for your contributions!

Issue #867, opened 2021-09-07
Messages are not received


korobaka:
### Bug description
When I send messages to myself, I get no logs.

### Reproduction steps
1. Initiate an instance of whatsapp-web.js
2. Connect
3. Try to send a message to yourself

### Expected behavior 
The message should be logged into the console.

### Relevant code
```
const qrcode = require('qrcode-terminal');

const { Client } = require('whatsapp-web.js');
const client = new Client();

const boxen = require('boxen');

client.on('qr', qr => {
    qrcode.generate(qr, {small: true});
});

client.on('ready', async () => {
    console.log('Client is ready!');
    const { info } = client;
    const battery = await info.getBatteryStatus();
    console.log(boxen(
        "Client phone: "+info.wid.user+"\n"+
        "Phone battery: "+battery.battery+"%\n"+
        "Phone blugged: "+battery.plugged+"\n"+
        "Pushname: "+info.pushname
    , { padding: 3, margin: 1 }));
});

client.on('message', message => {
    console.log(message.body);
});

client.initialize();
```

### Environment (please complete the following information):
**WhatsApp**
 - Account type :Business
 - Device OS :Android
 - WhatsApp Web version : 2.2134.10

**Library**
 - Browser: Chrome
 - NodeJS version (`node -v`): 16.8.0
 - npm or yarn version (`npm -v`): 6.14.8
 - whatsapp-web.js version: 1.14.1

**Other**
 - Operating system (the one running node) Windows
 - Operating system version (ex. Windows 10):  Win 10
### Additional context
-

korobaka:
after a few research, just found out ! client.js, line 244
it would be cool to add an option to determine if you want to fire the event if the message is from yourself.

PR #870, opened 2021-09-08
Add allowFromMe

justekoro:
Related to issue 867 (https://github.com/pedroslopez/whatsapp-web.js/issues/867), I added an option to choose if you want to trigger the event when the message is from yourself. This is useful for exploring wweb.js.


pedroslopez:
Hi @korobaka , you can use the `message_create` event to listen for all messages, including those you've sent. I believe using this event should be enough to accomplish this.

korobaka:
okie! closing pull request then. thank you for your answer!

PR #882, opened 2021-09-15
Feat: disable Logs & telemetry

tuyuribr:
Disable log and telemetry.
telemetry is disabled by default if you have adblock


tuyuribr:
I'm doing an /*eslint-disable */ because the fallthrough is intentional, maybe someone nows how to just disable fallthrough

PR #761, opened 2021-07-15
Multi-device Support [TESTERS NEEDED!]

purpshell:
**NOTICE:** Merge is waiting for multi-device's release.  

**ANOTHER NOTICE:** #889 has been opened since I changed plans (instead of having a fork, i'll have a multidevice branch and merge all the enhancements we can add later)

Suggested changes to bring multi-device beta support. 

Milestones:
- [x] Add envisioned file management system
- [ ] Test all functions using tests

Known issues - Things to fix: 
- [x] Send Messages
- [ ] Random "BUNDLE" errors when getting data is expected. We are working hard to identify the source of this issue, though I think it might have to do with serialization of data to be sent between Puppeteer and the browser environment.  
- [ ] #847 compatibility.
 
 
Known issues - not fixable: 
- Battery percentage information & events are gone
- Phone information gone too (phone, os, os version, wa version)
- Also you are now required to use "clientId" in the Client constructor's options parameter for more than one session. Look at [this](https://github.com/PurpShell/whatsapp-web.js/blob/330b33ed03722a2b674eec47756f44b7fddc72cf/src/Client.js#L32)
- Browserless / Remote Access (browserWSEndpoint) might not work as well (saving sessions) (re-writing soon to support)
- WhatsApp is now storing session data, contacts, chats, messages all in the browser, with no clear way to fetch the data, meaning we are resorting to using the fileSystem, and the files need to be updates in MS intervals if you are getting a lot of messages. Please find a way to use the filesystem. 

<a name="test">Testers</a>: 

Hey, this PR needs people to perform tests, mainly people with multidevice-enabled devices. Hopefully you can test and upload your logs, and if there is any issue hope that we can fix it and make it better. Thank you for your contributions so far. 
 - Steps: 
    - Clone this branch of repo (https://github.com/pedroslopez/whatsapp-web.js/tree/multidevice) or (`npm i github:pedroslopez/whatsapp-web.js#multidevice`)
    - Perform tests (/tests/ in the directory (please look at the readme before running anything))
    - Errors? Success? Send your results here, use a [pastebin](https://pastebin.com) to submit your logs, though.

PR #893, opened 2021-09-18
refactor qr code, refresh qr and ready state

rabrux:
hi!

i was testing this branch and it applies exclusively for beta login and i don't know why the _QR_RECEIVED_ event didn't shows up.

the code applies to both login methods (beta and non beta). maybe it's not the best implementation, i have been a long time without coding.

wish it helps.


rabrux:
it has problems on refresh qr, it _logouts_ when the page is loading

PurpShell:
why???? #889 #761

PR #906, opened 2021-09-27
Check if there is a previously active session

braymarco:
Using a file inside the browser's data folder could help validate if a session exists, temporarily helping the use of headless: true


PurpShell:
implemented changes, thanks for your contribution.

PR #946, opened 2021-10-31
fix: Fix inactivity issue

### LINKED ISSUE #377: WhatsApp Web stop reponding when app stay too long without activity
When whatsapp web stays for too long without activity, the client stop responding.

Try this:

1. Read the QR code and leave the app during the night;
2. When the app stay for too long with no messages received, the battery status and all activity stops and the client freezes;

I have created logs to monitor the activity and detect this behavior in same cases. I have to kill the node process and start it again to come back alive.

We need some auto test state to keep whatsapp web awake and listening. Maybe its a bug in whatsapp web version, but we need to find a solution that work this around. I will try to create a process monitoring for activity to kill and start the process automatic, but this is not an ideal solution.

purpshell:
This PR ensures that the session won't get logged out after 30 minutes of inactivity.

MD and non-MD compliant

 close #377


aliyss (approved):
lgtm

src/Client.js
```diff
@@ -170,7 +170,12 @@ class Client extends EventEmitter {
         }
 
         await page.evaluate(ExposeStore, moduleRaid.toString());
-
+        
+        // remember me
+        await page.evaluateOnNewDocument(() => {
```
pedroslopez, line 175: I'm not sure this would work here- `evaluateOnNewDocument` will run when there's some kind of navigation, which isn't the case. Should this be set before navigating to `WhatsWebURL`?

pedroslopez (changes requested):

src/Client.js
PurpShell: > I'm not sure this would work here- `evaluateOnNewDocument` will run when there's some kind of navigation, which isn't the case. Should this be set before navigating to `WhatsWebURL`?

we do it the same way we do the legacy session

PurpShell: You are right

PurpShell: fixed

pedroslopez (approved):
From what I can see, this is the default behavior (`remember-me` is `true` on clean startup without this fix), but I guess this will make sure it's set?

PurpShell:
> From what I can see, this is the default behavior (`remember-me` is `true` on clean startup without this fix), but I guess this will make sure it's set?

Yeah, since I found no references at all to enabling that option from our end

deba82:
You can use this code 

client.pupPage.click("#pane-side");

to perform a click on chrome window.
This will reconnect whatsapp mobile to web. 
Connected will come in mobile notification bar.
Perform operation after 5 to 6 seconds

PR #956, opened 2021-11-06
qrcode now uses observers instead of timeout

victormga:
removed 'jsqr' dependency.
removed 'qrRefreshIntervalMs' option (not needed anymore).
not more setTimeout, no more memleaks, no more 'Target Closed'.
tested with MD and non-MD.


victormga:
creating new pull request with the automatic detection of authentication

PR #963, opened 2021-11-08
Add isGif to message._patch

shirser121:
Add isGif to the message data.
When There is filed with isGif, set isGif to true


purpshell (approved):
lgtm

pedroslopez (approved):
Thanks for contributing! Added the typings for you :)

PR #985, opened 2021-11-22
fix: linkPreview returning BUNDLE error by Wap link preview generator

jaovitubr:
fix: Overriding Store.Wap.queryLinkPreview function that when set linkPreview true it returns BUNDLE error

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,9 +30,11 @@
   "homepage": "https://wwebjs.dev/",
   "dependencies": {
     "@pedroslopez/moduleraid": "^5.0.2",
+    "cheerio": "*",
     "fluent-ffmpeg": "^2.1.2",
     "jsqr": "^1.3.1",
     "mime": "^2.4.5",
+    "node-fetch": "^2.6.1",
     "node-webpmux": "^3.1.0",
     "puppeteer": "^10.1.0",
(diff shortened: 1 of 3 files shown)


tuyuribr:
Hey, thanks for contributing! So somethings about this:
If wa can't make a preview of a link we should not do anything
or
if we should do something, we should keep the new dependencies at a minimum. We can use pup to do what you are doing so maybe we should just keep using pup.

And we should default linkPreview to false

PurpShell:
> Hey, thanks for contributing! So somethings about this: If wa can't make a preview of a link we should not do anything or if we should do something, we should keep the new dependencies at a minimum. We can use pup to do what you are doing so maybe we should just keep using pup.
> 
> And we should default linkPreview to false

THIS. WhatsApp even said in their official notes about MD that link previews may not function as normal. 

I don't think this project is made to replicate any behavior other than that

PurpShell:
I don't even know how to review this

PurpShell:
Oh, and the reason the link preview generation is broken is because of Wap itself, it's deprecated in MD. I am positive that there is a module for this in MD, so try to find it

PurpShell:
reopen a new pr since changes

PR #990, opened 2021-11-24
Fix WA 2.2146.9

tuyuribr:
WA is serving two versions to me,  so we need to check window.mR.findModule('Conn').length for a while


pedroslopez (approved):
Thanks for this! I simplified the `Conn` stuff a little - I believe you were seeing two results because we were setting things on the default `Store` module directly.

PR #1004, opened 2021-11-24
Fix: multiple list sections

tuyuribr:


purpshell (approved):
LGTM

src/structures/List.js
```diff
@@ -57,7 +57,7 @@ class List {
      */
     _format(sections){
         if(!sections.length){throw '[LT02] List without sections';}
-        if(sections.length > 1){throw '[LT05] Lists with more than one section are having problems';}
+        if(sections.length > 1 && sections.find(s => typeof s.title == 'undefined')){throw '[LT05] You can\'t have empty titles if you have more than one section';}
```
PurpShell, line 60:
```suggestion
        if(sections.length > 1 && sections.filter(s => typeof s.title == 'undefined').length > 1){throw '[LT05] You can\'t have more than one empty title.';}
```

PurpShell: list messages can have only 1 empty title, which is at the beginning.

aliyss (approved):

Issue #1014, opened 2021-11-27
There is no option to get chat default messages due to static 50 limit in fetchMessages()


AxeemHaider:
**Is your feature request related to a problem? Please describe.**
There is no way to get the default messages, Docs say you can fetch **50** messages older than these will come in cost. But
this is not `true` because mostly there are messages in smaller amount you can't apply static limit to all chats. 

**Describe the solution you'd like**
Remove the static limit of 50 in `fetchMessages()`

PurpShell:
Please explain this use case better, i didn't understand how useful this is.

PR #1015, opened 2021-11-27
Get Default messages of chats

AxeemHaider:
This is related to #1014 

Removing static limit from `fetchMessages()` and it solve things. Now if you don't provide any `limit` then it will get default messages.


src/structures/Chat.js
```diff
@@ -171,30 +171,30 @@ class Chat extends Base {
     /**
      * Loads chat messages, sorted from earliest to latest.
      * @param {Object} searchOptions Options for searching messages. Right now only limit is supported.
-     * @param {Number} [searchOptions.limit=50] The amount of messages to return. Note that the actual number of returned messages may be smaller if there aren't enough messages in the conversation. Set this to Infinity to load all messages.
+     * @param {Number} [searchOptions.limit] The amount of messages to return. Note that the actual number of returned messages may be smaller if there aren't enough messages in the conversation. Set this to Infinity to load all messages.
```
pedroslopez, line 174:
```suggestion
     * @param {Number} [searchOptions.limit] The amount of messages to return. If no limit is specified, the available messages will be returned. Note that the actual number of returned messages may be smaller if there aren't enough messages in the conversation. Set this to Infinity to load all messages.
```

pedroslopez (changes requested):
This looks ok to me, though noting that this would change the behavior to not fetch anything and only return the available messages my default, unless a limit is specified. I think this makes sense and is in line with what most people would want.

Let's make sure that the docs reflect the new behavior accurately by updating the jsdoc strings and typings. We'll need to make relevant changes to `MessageSearchOptions` here:

https://github.com/pedroslopez/whatsapp-web.js/blob/46ba9c5e9aa0bc4de5eba1a561e7f4ef8050084a/index.d.ts#L910-L918

purpshell (approved):

pedroslopez (approved):

PR #1049, opened 2021-12-11
fix getLabels error

zulfio:
fix error **Error: Evaluation failed: TypeError: cannot read property 'models' of undefined"** when calling getLabels() method


purpshell (approved):
LGTM

pedroslopez (approved):
Thanks!

PR #1091, opened 2022-01-02
[MD] Added: function getPresences

viniciusgdr:
added a method to get the presence of chat users.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,7 +21,7 @@ declare namespace WAWebJS {
         acceptInvite(inviteCode: string): Promise<string>
 
         /** Accepts a private invitation to join a group (v4 invite) */
-        acceptGroupV4Invite: (inviteV4: InviteV4Data) => Promise<{status: number}>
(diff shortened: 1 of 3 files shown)


PurpShell:
Hey! Please check on the previous PRs, to know what you are missing..

TL:DR; you need to make a async requestPresenceUpdate(jid: string): Promise<void>

Instead of a store based system..

This is to make sure you were actually subbed to a presence before you try to get nothing and get thrown an error.. 

Also make it an event instead.

viniciusgdr:
> Hey! Please check on the previous PRs, to know what you are missing..
> 
> TL:DR; you need to make a async requestPresenceUpdate(jid: string): Promise
> 
> Instead of a store based system..
> 
> This is to make sure you were actually subbed to a presence before you try to get nothing and get thrown an error..
> 
> Also make it an event instead.

I don't understand about the store-based system, could you correct or let me know what it is?

AKwoKWH:
Bring up this MR request
I have a try to run it on my branch and it can get the last seen, but the isOnline is always `false`?
Anything missing here??

PR #1103, opened 2022-01-12
Fix the option to use special options

shirser121:
When we want to use some extra function we need to pass it into the options object, but because it inserted in the first line, the auto options reset some options (like 'type').
This is minor change that can help to add functionallity


tuyuribr:
We should be careful about this. We changed this behavior for a reason I'm going to try to find the reason

PurpShell:
> We should be careful about this. We changed this behavior for a reason I'm going to try to find the reason

no real actual reason

shirser121:
> > We should be careful about this. We changed this behavior for a reason I'm going to try to find the reason
> 
> no real actual reason

So we should change it.
This is the only way to use mediaInfo from received message.
This is the way to use more options

purpshell (approved):
Lgtm

pedroslopez (approved):
Thanks! Made a few changes so that only the options passed via the `extra` option are used to override others to be a bit safer.

PR #1121, opened 2022-01-24
Added getCommonGroups to branch main and fixed it.

Schwartzblat:


shirser121:
What is the difference between `window.mR.findModule('findCommonGroups')[0].findCommonGroups` to `window.Store.Wap.commonGroupsFind`?

Schwartzblat:
I will check.

Schwartzblat:
Ok you are right. I will fix it.

shirser121:
Perfect!
Also, maybe we should return chatGroup instead of chat id

Schwartzblat:
> Perfect! Also, maybe we should return chatGroup instead of chat id

I don't think it's a good idea because sometimes you want to just get the number of the chats or use the ids.
What are you saying about adding options parameter so you can choose what output you want?

shirser121:
> I don't think it's a good idea because sometimes you want to just get the number of the chats or use the ids. What are you saying about adding options parameter so you can choose what output you want?

I think it's unnecessary, because as you say, we can get the chats by the ida

Schwartzblat:
> > I don't think it's a good idea because sometimes you want to just get the number of the chats or use the ids. What are you saying about adding options parameter so you can choose what output you want?
> 
> I think it's unnecessary, because as you say, we can get the chats by the ida

So leave it as it is?

shirser121:
> > > I don't think it's a good idea because sometimes you want to just get the number of the chats or use the ids. What are you saying about adding options parameter so you can choose what output you want?
> > 
> > 
> > I think it's unnecessary, because as you say, we can get the chats by the ida
> 
> So leave it as it is?

Yes.
In my project i did exactly the same.
I planned to crate a PR but you did it perfectly

shirser121 (approved):
LGTM

Schwartzblat:
> > > > I don't think it's a good idea because sometimes you want to just get the number of the chats or use the ids. What are you saying about adding options parameter so you can choose what output you want?
> > > 
> > > 
> > > I think it's unnecessary, because as you say, we can get the chats by the ida
> > 
> > 
> > So leave it as it is?
> 
> Yes. In my project i did exactly the same. I planned to crate a PR but you did it perfectly

Lol thanks. In my project I am using the multidevice branch so the solution that I uploaded first is compatible to both.

purpshell (approved):
lgtm

Schwartzblat:
@shirser121 can you review again?

purpshell (approved):
Again, lgtm

shirser121 (approved):
Hope this will get merge

pedroslopez:
Closing as this was included as part of the MD merge #889

PR #1144, opened 2022-01-27
Add function getMessageInfo

shirser121:
Get the message info, delivery, reads etc...

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -649,22 +649,22 @@ class Client extends EventEmitter {
     async getMessageInfo(messageId) {
         return await this.pupPage.evaluate(async messageId => {
             const msgExist = await window.Store.MsgInfo._find(messageId);
-            if (typeof msgExist._value === "string") {
+            if (typeof msgExist._value === 'string') {
                 throw msgExist._value;
             }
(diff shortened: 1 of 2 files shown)


PurpShell:
did you know this actually exists?

https://docs.wwebjs.dev/Message.html#getInfo

src/Client.js
```diff
@@ -641,6 +641,36 @@ class Client extends EventEmitter {
...
     }
 
+    /**
+     * Returns an object with information about the message
+     * @param {string} messageId
+     * @returns {Promise<object>} message information
+     */
+    async getMessageInfo(messageId) {
+        return await this.pupPage.evaluate(async messageId => {
+            const msgExist = await window.Store.MsgInfo._find(messageId);
```
PurpShell, line 651: you seem to be missing some stuff in Injected.js, or is this something that we already put to Store?

shirser121: Please explain the question.
MsgInfo is already exist

shirser121:
> did you know this actually exists?
> 
> https://docs.wwebjs.dev/Message.html#getInfo

I just want to improve that and add it to client

src/Client.js
ceringo: is this method working?

PR #1150, opened 2022-01-28
Improved performance 😊

FUZIK:


PurpShell:
it doesn't make a big difference

tuyuribr:
Yeah, to make a difference you should be sending more than one message per second, so it wont affect anything in the real world

PR #1151, opened 2022-01-28
Fix getProfilePicUrl

FUZIK:
😊


Schwartzblat:
You need to add to injected.js this line:
```
window.Store.ProfilePicThumb = window.mR.findModule("ProfilePicThumb")[0].default.ProfilePicThumb;
```

FUZIK:
The method works only with cached images

PR #1166, opened 2022-02-02
Fix chat.clearMessages()

shirser121:
Instead of `window.Store.SendClear.sendClear(chat, false)` that return an error and not work, use `chat.deleteMsgs();`

### DIFF
--- a/src/util/Injected.js
+++ b/src/util/Injected.js
@@ -446,7 +446,7 @@ exports.LoadUtils = () => {
     window.WWebJS.sendClearChat = async (chatId) => {
         let chat = window.Store.Chat.get(chatId);
         if (chat !== undefined) {
-            await window.Store.SendClear.sendClear(chat, false);
+            await chat.deleteMsgs();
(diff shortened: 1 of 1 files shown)


pedroslopez:
I can't seem to reproduce an issue with the current implementation 🤔

Calling `await chat.clearMessages()` seems to work as expected for me. Can you share more about when this issue occurs, what error message comes up, and how to reproduce it?

shirser121:
> I can't seem to reproduce an issue with the current implementation 🤔
> 
> Calling `await chat.clearMessages()` seems to work as expected for me. Can you share more about when this issue occurs, what error message comes up, and how to reproduce it?

fix in new PR (https://github.com/pedroslopez/whatsapp-web.js/pull/1213)
this is not throw an error, its just not delete the chat messages.
in the new PR I just change from `await window.Store.SendClear.sendClear(chat, false);` to `await window.Store.SendClear.sendClear(chat, true);`

PR #1181, opened 2022-02-08
fix: dont cause redirect loop due to session restore

### LINKED ISSUE #1179: Error when user disconnect on APP
### Bug description
Since the last update when the user quit of paired session the client crashes

### Reproduction steps
1. Sync  on main branch with a non-md account
2. After ready, logout directly on app

### Expected behavior 
Close client without exception and destroy client (browser) but the browser still open and get back to qrcode scan area
ref. PR #1169 

### Relevant code
```console
(node:19712) UnhandledPromiseRejectionWarning: Error: Evaluation failed: TypeError: Cannot read properties of undefined (reading 'AppState')
    at __puppeteer_evaluation_script__:2:33
    at ExecutionContext._evaluateInternal (C:\wamp64\www\whatsapp-api\node_modules\puppeteer\lib\cjs\puppeteer\common\ExecutionContext.js:221:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async ExecutionContext.evaluate (C:\wamp64\www\whatsapp-api\node_modules\puppeteer\lib\cjs\puppeteer\common\ExecutionContext.js:110:16)
    at async Client.getState (C:\wamp64\www\whatsapp-api\node_modules\whatsapp-web.js\src\Client.js:729:16)
    at async C:\wamp64\www\whatsapp-api\node_modules\whatsapp-web.js\src\Client.js:454:30

(Use `node --trace-warnings ...` to show where the warning was created)
(node:19712) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled pr
omise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:19712) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### Environment (please complete the following information):
**WhatsApp**
 - Account type Business and Standard
 - Device Android
 - WhatsApp Web version: 2.2202.12

**Library**
 - Browser Chrome
 - NodeJS version (`node -v`): v14.17.3
 - npm or yarn version (`npm -v`): >7
 - whatsapp-web.js version: latest

**Other**
 - Operating system version (ex. Windows 10):

pedroslopez:
The redirect loop was caused because WhatsApp Web was clearing the session from local storage, but we were always adding it back with `evaluateOnNewDocument`. This ensures it only happens the first time.

The `UNPAIRED` state being emitted when logging out seems to be inconsistent, so checking for frame navigation still seems to be the most reliable way to detect a logout. This also should fix #1179 by returning `null` when checking state if the Store isn't available.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -165,8 +165,8 @@ declare namespace WAWebJS {
 
         /** Emitted when the client has been disconnected */
         on(event: 'disconnected', listener: (
-            /** state that caused the disconnect */
-            reason: WAState
+            /** reason that caused the disconnect */
+            reason: WAState | "NAVIGATION"
         ) => void): this
 
         /** Emitted when a user joins the chat via invite link or is added by an admin */
--- a/index.d.ts
+++ b/index.d.ts
@@ -166,7 +166,7 @@ declare namespace WAWebJS {
         /** Emitted when the client has been disconnected */
         on(event: 'disconnected', listener: (
             /** reason that caused the disconnect */
-            reason: WAState | "NAVIGATED"
+            reason: WAState
         ) => void): this
 
         /** Emitted when a user joins the chat via invite link or is added by an admin */
--- a/index.d.ts
+++ b/index.d.ts
@@ -165,7 +165,7 @@ declare namespace WAWebJS {
 
         /** Emitted when the client has been disconnected */
         on(event: 'disconnected', listener: (
-            /** reason that caused the disconnect */
+            /** state that caused the disconnect */
             reason: WAState
         ) => void): this
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -114,7 +114,7 @@ class Client extends EventEmitter {
         if (this.options.session) {
             // Check if session restore was successful
             try {
-                await page.waitForSelector(KEEP_PHONE_CONNECTED_IMG_SELECTOR, { timeout: 0 });
+                await page.waitForSelector(KEEP_PHONE_CONNECTED_IMG_SELECTOR, { timeout: this.options.authTimeoutMs });
             } catch (err) {
                 if (err.name === 'TimeoutError') {
                     /**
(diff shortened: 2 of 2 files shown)


guicuton (approved):
Changes works like a charm!

## Events of my APP (Windows)
image
- line 1 and 2
  > Initialize the client object with `client.initialize()`
- line 3
  > Waiting qrcode on event `qr`
- line 4 and 5
  > Event `authenticated` log and save the session
- line 6 and 7
  > Event `ready` log uses the `client.getWWebVersion()`
- line 8
  > After logout on app we received the `OPENING` message on event `change_state`
- line 9
 > Browser closes without any exception or warnings and received the `NAVIGATION` message on event `disconnected`

## Events of my App (Linux)
image
The events above are only the phases when user quit off session directly on app.
- Lines 1, 2 and 3
  > Messages received on event `change_state`
- Line 4
  > Message received on event `diconnected`

## Client option - `takeoverOnConflict`
With these changes the option is working all good without any exceptions or warnings.

Despite of the message events seems to be in different order when user logout on app according the OS where the lib is located, it's all working properly.

pedroslopez:
@guicuton nice! thanks for trying it out.

> Despite of the message events seems to be in different order when user logout on app according the OS where the lib is located, it's all working properly.

Yeah, I also saw some inconsistent behavior. It seems like the issue is due to a race condition on WhatsApp Web's side, so not much we can do there. Ideally it would always go to the `UNPAIRED` state, but sometimes the nav happens before that and it never gets to `UNPAIRED`. This is why we have to detect / emit the disconnected event on nav still.

PR #1185, opened 2022-02-09
Update ChatFactory.js

hayzedDev:


PurpShell:
This is useless unless you are using the project raw.. please learn to use NPM and use it as a dependency

PR #1187, opened 2022-02-10
Add raw data to Message

shirser121:
Sometimes we need more information about the message, so we should save all the message data in Message.raw


tuyuribr:
Sounds like a good Idea, but is missing docs (.ts)

pedroslopez:
This has been included as part of the MD merge (#889)

PR #1196, opened 2022-02-14
Getting PID of running browser

inceabdullah:
in some cases, PID is useful. Setting priority, core affinity, an so on. I just have put an event just after browser initializing.

This event return PID or error if could not get PID with the function `browser.process()`

I use this event, like:
```JavaScript
client.on("browser_init", (pid, err) =>{
    if (err) {
        console.error("Process could not be running", e);
        return;
    }
    console.log("browser is running with the pid:", pid);
});
```

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -75,6 +75,14 @@ class Client extends EventEmitter {
             browser = await puppeteer.launch(this.options.puppeteer);
             page = (await browser.pages())[0];
         }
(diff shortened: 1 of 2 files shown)


tuyuribr:
Sounds like a good thing to have, but is missing some docs also we have to put a warning to this: "Returns `null` if the browser instance was created with `puppeteer.connect` method "

PR #1197, opened 2022-02-14
Fix syntax error at line 472

siddiquiaffan:


tuyuribr:
Please re-submit with only your changes so we can review it, if it's something to do with MULTIDEVICE please open a PR to the correct branch

PR #1213, opened 2022-02-16
Fix clear chat

shirser121:
change from `await window.Store.SendClear.sendClear(chat, false);` to `await window.Store.SendClear.sendClear(chat, true);`


pedroslopez:
Hmm, I still can't reproduce any issue with the current implementation. All chat messages are cleared just fine in my tests.
Do you have more info on what this parameter means and why we should change it to `true`?

Could you explain more what issue you're seeing with the current implementation and maybe more information about your setup so I can reproduce it?

setsmart:
Please can someone help me, how do I return a link after the person has clicked on the button?
Thanks all.

PR #1227, opened 2022-02-21
Auth Strategies

pedroslopez:
Cleans up the new local file storage and legacy auth into separate Auth Strategies.

This concept could be used later to add something like `S3Auth`, for example. It also means users can develop their own auth strategies.

### `LocalAuth` - Filesystem-based auth that uses puppeteer's userDataDir to persist browser data
Used by default if no `session` option is passed to the client.
```js
const client = new Client();
```

To customize clientId or dataPath, pass as options to the auth strategy:
```js
const client = new Client({
    authStrategy: new LocalAuth({
        clientId: "example",
        dataPath: "./some/path"
    })
});
```

### `LegacySessionAuth` - traditional method that pulls tokens from local storage
Not available for MD.
Used by default if a `session` option is passed to the client, but this is only for backwards compatibility.

```js
const client = new Client({
    session: { } // session data here from wherever you get it from
});
```

Should be used explicitly to be able to get the payload in the auth event without passing a session:
```js
const client = new Client({
    authStrategy: new LegacySessionAuth({
        session: { } // session data here from wherever you get it from
    })
});
```

### `NoAuth` - does not provide any auth saving
```js
const client = new Client({
    authStrategy: new NoAuth()
});

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,7 +1,7 @@
-const { Client, Location, List, Buttons } = require('./index');
+const { Client, Location, List, Buttons, LocalAuth } = require('./index');
 
 const client = new Client({ 
-    clientId: 'example',
+    authStrategy: new LocalAuth({ clientId: 'example' }),
     puppeteer: { headless: false }
 });
 
--- a/index.js
+++ b/index.js
@@ -21,5 +21,10 @@ module.exports = {
     ProductMetadata: require('./src/structures/ProductMetadata'),
     List: require('./src/structures/List'),
     Buttons: require('./src/structures/Buttons'),
+    
+    // Auth Strategies
+    LocalAuth: require('./src/authStrategies/LocalAuth'),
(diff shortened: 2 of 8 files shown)


jtourisNS:
We need to have a good handling of the 'auth_failure' event. 

This should fire if we are not providing a valid path to the MD strategy. This is currently not triggering if an invalid WWebJS session is provided to the authentication. The unhappy path is not easy to handle, the flow ends on a dead application where the message given by the wpClient is:

**Connecting to WhatsApp**
_Make sure your computer has an active Internet connection._

jtourisNS:
Screen Shot 2022-02-25 at 7 10 51 PM

pedroslopez:
@jtourisNS Do you have a reproducible way to get this issue? Is it because of something happening to the WWEBJS folder?

I'm a bit weary of just looking for this popup because it seems like it could come up in regular circumstances 🤔

pedroslopez:
Superseded by #1257

PR #1230, opened 2022-02-22
Added message duration in seconds

Yarondr:
I added a duration variable in the message, so it will be possible to get the message duration in seconds if the message is audio/video/etc...


purpshell (approved):
looks good to me

pedroslopez (approved):

Yarondr:
Great.
Can we add this to the multidevice branch too?

Issue #1235, opened 2022-02-24
Error on create store object - Uncaught TypeError: Cannot read properties of undefined (reading 'Socket')


chirag90851:
window.Store.AppState = window.mR.findModule('STREAM')[0].Socket;

Uncaught TypeError: Cannot read properties of undefined (reading 'Socket')

linconrezende:
In that specifica case, it could not find the 'STREAM', it is undefined and it tries to get by the index.
It could be a timming problem, maybe the whatsapp web page took too long to load stuff and puppeteer couldn`t load it.
If you post the stack trace, maybe I can try to help

chirag90851:
After new business WhatsApp update it's not working. I am just open web.whatsapp.com in the browser and execute create store object js file.
WhatsApp Business Version : 2.22.4.74  , Web Whatsapp Version : 2.2204.13

hamzaFpco:
I have the same error when I launch the code and start to do something

chirag90851:
For reproduce issue -> Open web.whatsapp.com in browser and login using QR Code -> Refresh Browser -> Now execute store object create js file in console.

kikemarto:
Here with the same situation but only one whatsapp of many.

fm1randa:
I have the same error.

eduardoa1ves:
same error

fm1randa:
> Temporary solution, works to me:
> 
> * Remove line window.Store.AppState = window.mR.findModule('STREAM')[0].Socket; from **Injected.js**
> * Remove line window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); }); from **Client.js**

Thanks! It worked for me.

dexterpedro:
> Temporary solution, works to me:
> 
> * Remove line window.Store.AppState = window.mR.findModule('STREAM')[0].Socket; from **Injected.js**
> * Remove line window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); }); from **Client.js**

Thanks! Works here.

xanaxxx:
> Temporary solution, works to me:
> 
>     * Remove line window.Store.AppState = window.mR.findModule('STREAM')[0].Socket; from **Injected.js**
> 
>     * Remove line window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); }); from **Client.js**

[...]

PR #1238, opened 2022-02-24
Fix: Validation not exist socket in stream module ( store.AppState )

pablo2p:
small validation in the stream, where the valid socket attribute was added, if it doesn't exist, it continues running the authentication in the normal way.


KeruMx:
that modification dont fixes anything in issue #1235.
You need to find out which module was changed.
There is a comment that found the module, but I have not tried it because I do not have access to the multidevice activated on my phone

PR #1239, opened 2022-02-24
Feat: Check is MultiDevice

pablo2p:
The purpose of this function is simply to return whether the version is beta or not, in this way, it is possible to manage the best path to the final consumer.


KeruMx:
that function is already exist.

KeruMx:
> that function is already exist.

idk why in that branch not does exists

PR #1259, opened 2022-02-28
fix: Cannot destructure property 'session' of 'undefined' as it is un…

inceabdullah:
at this example, app breaks. Because options is empty: `undefined`

```JavaScript
const client = new Client({
    authStrategy: new LegacySessionAuth()
});
```

Error: `Cannot destructure property 'session' of 'undefined' as it is undefined.`

so, I have just added `{}` in constructor args
```JavaScript
constructor({ session, restartOnAuthFail }={}) {
```


pedroslopez (approved):
Thanks!

PR #1262, opened 2022-02-28
add typescript lint for LocalAuth attributes

fm1randa:
Adds 
```ts
public clientId?: string;
public dataPath?: string;
```


pedroslopez (approved):
Thanks!

PR #1267, opened 2022-03-01
Fix logout method

diasfs:
The logout method was trying to use a function call passing the "dataDir" as the "this" reference


pedroslopez (approved):
Thanks!

PR #1271, opened 2022-03-01
feat: Save MessageMedia to a file

purpshell:
In this pull request I add a function to MessageMedia to save files, since this has apparently became a headache, and I though having a function that does this by default is fitting. 

Example usage:
```js
const media = await msg.downloadMedia();
await media.toFilePath('/home/purpshell/Documents/message') // the code adds the extension for you (if you enter a directory we will use the filename property (if it's null then we'll throw an error))
```


src/structures/MessageMedia.js
```diff
@@ -98,6 +98,32 @@ class MessageMedia {
...
+     * media.toFilePath('/home/purpshell/Documents/message') // the code adds the extension for you (if you enter a directory we will use the filename property (if it's null then we'll throw an error))
+     * @returns {null} no need to return anything
+     */
+    async toFilePath(filePath) {
+        const ext = mime.getExtension(this.mimetype);
+        if (fs.existsSync(filePath)) {
+            const stat = await fs.promises.stat(filePath);
+            if (stat.isDirectory()) {
+                if (this.filename){
+                    filePath += this.filename;
```
RakSrinaNa, line 116: Wouldn't this break if the input doesn't end with a folder separator ?

Like with input `/path/to/folder` it'll save to `/path/to/folderfilename.ext`

src/structures/MessageMedia.js
```diff
@@ -98,6 +98,32 @@ class MessageMedia {
...
+                } else {
+                    throw Error('You passed in a directory but the filename is empty');
+                }
+            }
+        }
+        if (!filePath.includes('.')) {
```
RakSrinaNa, line 122: What if the provided input folder is `./path/to/myfile` or `/path/to.my.folder/myfile`, the extension wouldn't be added.

PurpShell: you're right, ill make some changes

src/structures/MessageMedia.js
PurpShell: aight

Rakambda (approved):

tuyuribr (approved):
LGTM

PR #1273, opened 2022-03-01
feat: Adding file size by bytes to MessageMedia

purpshell:
# MessageMedia File size

##  INFO

This pull request is setting a new property to MessageMedia, allowing you to know the file size of the file. 

Accuracy-wise, it seems whatsapp gives the value of the file before compression so it seems a little off than in the preview.

Sometimes i have seen 2.7mb turn into 3 on the preview, it seems whatsapp rounds up on that counter

## DEMO

image

## USAGE

**FORMAT:** BYTES

Pass in the "filesize" property when making a file, it is optional though.
fromUrl and downloadMedia automatically handle this.


sergiooak (approved):
Tested on a fork, and works great, I think it should be merged

tuyuribr (approved):
LGTM!

aliyss (approved):
LGTM

PR #1300, opened 2022-03-05
feat: RemoteAuth Strategy - Initial Draft

jtouris:
Initial draft of remote-based authentication


src/authStrategies/RemoteAuth.js
```diff
@@ -0,0 +1,254 @@
...
+                    }))
+                    .on('error', err => reject(err))
+                    .on('finish', () => resolve());
+            });
+        } catch (error) {
+            console.log('RemoteAuth: unzipSession => ', error);
+        }
+    }
+
+    static async deleteMetadata() {
```
pedroslopez, line 203: I'm a little concerned that deleting the other files on a running instance might conflict with other things. Maybe we could use the options available in `tar` to filter and only include the files we need in the archive, or copy them to a temp directory.

src/authStrategies/RemoteAuth.js
```diff
@@ -0,0 +1,254 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const unzipper = require('unzipper');
+const archiver = require('archiver');
+const fsPromise = require('fs').promises;
+const schedule = require('node-schedule');
```
pedroslopez, line 8: This dependency is also missing. What's the benefit of using this versus a plain `setInterval`?

pedroslopez (commented):
Nice start! This is going to be really helpful.

I'm interested to know if doing a backup every X amount of time will be enough or if we also need to trigger a save when certain things happen. For example, when you start chatting with a new number I believe that generates an encryption key with that person, so I wonder what happens if after restoring you haven't backed that key up. We can test this later though.

General question: what do you think the interface will be like with RemoteAuth and the different stores? Will we pass them the archive and expect them to return an archive as well, or something else?

PR #1327, opened 2022-03-12
feat: logout on conflict

jeanvcastro:
I added the options that change the destroy method. They have the opposite effect of the _takeoverOnConflict_ option.

If the **logoutOnConflict** option is passed as true to the client and another session is detected (another browser), it will be logged out of the current session.

The **restartOnConflict** option will cause the client to be restarted if it is turned off after a conflict. If used together with the _logoutOnConflict_ option, the user will be logged out and the client will generate qrcodes again.


src/util/Constants.js
```diff
@@ -11,6 +11,8 @@ exports.DefaultOptions = {
     qrMaxRetries: 0,
     takeoverOnConflict: false,
     takeoverTimeoutMs: 0,
+    logoutOnConflict: false,
+    restartTimeoutMs: false,
```
smuzaffar, line 15:
```suggestion
    restartOnConflict: false,
```

PR #1354, opened 2022-03-26
Add ConnectOptions typings

ahmedfouzan:
Add ConnectOptions type for puppeteer to support this [PR](https://github.com/pedroslopez/whatsapp-web.js/pull/727)


pedroslopez (approved):
Thanks! Simplified the type a bit by using the PuppeteerNodeLaunchOptions type defined here https://github.com/puppeteer/puppeteer/blob/08e9978ce662cf1c2c2708f7e6098cbea328cbe2/src/node/LaunchOptions.ts#L142-L144

PR #1362, opened 2022-03-29
Fix: Cannot read properties of undefined (reading 'getProfilePicFull')

jtouris:
Module Fix applied for error:

`Error: Evaluation failed: TypeError: Cannot read properties of undefined (reading 'getProfilePicFull')`

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -877,19 +877,15 @@ class Client extends EventEmitter {
      */
     async getProfilePicUrl(contactId) {
         const profilePic = await this.pupPage.evaluate(async contactId => {
-            let asyncPic;
-            if (window.Store.Features.features.MD_BACKEND) {
+            try {
                 const chatWid = window.Store.WidFactory.createWid(contactId);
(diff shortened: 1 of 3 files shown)


LuisWichoGt (approved):
Si se corrigio el error

dxversity:
It works, thanks

zulfio (approved):
thanks, it's work!

selamethanafi:
thanks, it works!

aliefdany (approved):
idk but it works!

pedroslopez (approved):
Thanks! Made a few changes to ensure we get the right module (it could potentially find `Wap` instead in some cases) and simplified the `getProfilePicUrl` function a bit since the called function is now MD-safe.

jtourisNS:
@pedroslopez Awesome! Shouldn't the mentioned issues be associated to the PR so they can be closed?

Ir001 (commented):
Thanks, its work

PR #1371, opened 2022-04-05
fix:  fetchMessages infinite loop & memory leak

jtouris:
`await chat.loadEarlierMsgs()` used on line#186 of `Chat.js` may return values `undefined` or `"0"`.  

Both scenarios should break the while loop due that it causes memory leak on server/cloud applicaitons.
Memory leak issue & infinite loop have disappeared on my cloud server after this.


jtourisNS:
This provokes one more additional iteration on the loop when loadedMessages is 0 instead of undefined, but this was actually not the root cause of the problem infinite loop problem. 

`msg.DownloadMedia()` was the one causing the infinite loop on my application so I decided to close this and fix it on my end.

PR #1407, opened 2022-04-21
Fix: Cannot read properties of undefined (reading 'features')

shirser121:
Find MD backend without features

PR #1410, opened 2022-04-21
Fix: Cannot read properties of undefined (reading 'features') // original: (#1407)

purpshell:
Credits: @shirser121 

apply this patch by doing: 
```
npm install github:pedroslopez/whatsapp-web.js#features_fix
```


shirser121 (approved):
Great!
We should have a way to merge to main in this cases that pedro is not available

jtouris (approved):

absortian (approved):
It's OK!

PR #1437, opened 2022-05-07
add afterBrowserInitialized hook

MinosIE:


PurpShell:
why did you add prettier? anyways this is not needed.

PR #1443, opened 2022-05-08
Add canSend to groupMetadata

shirser121:
chat.readOnly not works for me (always false) so we need to know if we can send a message to the group


PurpShell:
can't you just use groupMetadata.announce?

shirser121:
@PurpShell I think this is necessary, groupMetadata.announce not doing the work

PR #1480, opened 2022-06-01
Fix of fetchMessages

jurajmatus:
`chat.loadEarlierMsgs()` returns an empty array instead of null|undefined. Because of this it lead to an infinite cycle that never ended.


su-chang (approved):
LGTM

shirser121 (approved):
LGTM

mojila (approved):

src/structures/Chat.js
```diff
@@ -184,7 +184,7 @@ class Chat extends Base {
             if (searchOptions && searchOptions.limit > 0) {
                 while (msgs.length < searchOptions.limit) {
                     const loadedMessages = await chat.loadEarlierMsgs();
-                    if (!loadedMessages) break;
+                    if (!loadedMessages?.length) break;
```
pedroslopez, line 187: Please avoid using features that are not available in node v12 since we still support this version (optional chaining is not available)

pedroslopez (changes requested):
Thanks for the fix! Makes sense to me, just one note on proper compatibility with older node versions.

pedroslopez (approved):

PR #1481, opened 2022-06-01
Async buffer->BASE64 conversion

### LINKED ISSUE #1464: Download media is a blocking call

### Describe the bug

I'm running the library in AWS lambda environment which has some time limitations. Due to that it's massively important to be able to chunk the work and to be able to cut it short if timeout is nearing.

However, I've had cases when downloading media took over 300 seconds and as I found out, it's a blocking call, so I can't even use timeout to stop it because js's event loop won't leave the function.

### Expected behavior

All long-running operations as non-blocking

Possibly provide an option to return `ArrayBuffer` in `downloadMedia`

1. Run `message.downloadMedia()` for some big video/file

### Relevant Code

https://github.com/pedroslopez/whatsapp-web.js/blob/main/src/structures/Message.js#L399
https://github.com/pedroslopez/whatsapp-web.js/blob/b16e1cdf83c7fd901a6fd198a4d746f639757bd6/src/util/Injected.js#L471 -> this function should work asynchronously (it would be good for others too, although I'm not sure where exactly they are used)

### Browser Type

Chromium

### WhatsApp Account Type

Standard

Yes, I am using Multi Device

### Environment

* OS: AWS Lambda internal environment
* Phone OS: Android
* whatsapp-web.js version: 1.16.6
* WhatsApp Web version: the newest
* Node.js Version: 14

### Additional context

_No response_

jurajmatus:
Solution for #1464


jurajmatus:
Rebased

purpshell (approved):
looks good to me

tuyuribr (approved):
LGTM! tested against 2.2228.14 and worked with small and big files

PurpShell:
Update branch.

jurajmatus:
Updated

PR #1499, opened 2022-06-08
Add search option on chat.fetchMessages by min/max timestamp

### LINKED ISSUE #1547: Add option to fetchMessages | Time period

_No response_

fetchMessages using GET MEthod with Time limit

_No response_

### Additional context

_No response_

LucianoHanna:


lfdelphino:
It works here, might need a few exceptions like when max is lower than min or if there is no limit set but seems to be working fine.

jurajmatus:
I would suggest using `break` in case there is a message older than `minTimestamp` (that means that all the messages loaded after will be even older):

```diff
+ if (searchOptions && searchOptions.minTimestamp && loadedMessages.some(msg => msg.t < searchOptions.minTimestamp)) {
+    break
+}
msgs = [...loadedMessages.filter(msgFilter), ...msgs];
```

purpshell (changes requested):
👍 but fix ESLint, will approve after that

tuyuribr:
don't forget index.d.ts too

PR #1533, opened 2022-06-21
feat: Loading screen listener with percent and message

tonbotfy:
It will help a lot of people who need this whatsapp screen status.

When the screen is loading, the 'loading_screen' event is called with two parameters:
percent
message

image


tonbotfy:
> this is great, that's exactly the feature I mentioned on #1556! On a side note, you made a lot of style changes to the code that are not related to the feature, removing these changes would make the PR easier to review,

Actually, my code editor auto save was on.

I'll try to recreate the feature with the project's code styling settings and upload a new PR.

PR #1606, opened 2022-07-26
lastMessage in chat

tofers:
I need to get the last message from the dialog. How is it done in the official WhatsApp Web.


PurpShell:
you can use fetchMessage for that (it is sorted by the newest 50 messages and back)

tuyuribr (changes requested):
index.d.ts docs

index.d.ts
```diff
@@ -1018,6 +1018,8 @@ declare namespace WAWebJS {
         timestamp: number,
         /** Amount of messages unread */
         unreadCount: number,
+        /** Last message fo chat */
```
tuyuribr, line 1021: typo

tuyuribr (changes requested):
LGTM

tofers (commented):
What is wrong in the description?

src/util/Injected.js
```diff
@@ -400,7 +400,15 @@ exports.LoadUtils = () => {
             await window.Store.GroupMetadata.update(chatWid);
             res.groupMetadata = chat.groupMetadata.serialize();
         }
-
+        
+        res.lastMessage = null;
+        if (res.msgs && res.msgs.length) {
+            const lastMessage = window.Store.Msg.get(chat.lastReceivedKey._serialized);
```
PurpShell, line 406:
```suggestion
            const lastMessage = window.Store.Msg.find(a => a.id._serialized == chat.lastReceivedKey._serialized);
```

PurpShell: .get is sometimes unreliable

purpshell (changes requested):

src/util/Injected.js
tofers: this doesn't work for me, causes message serialization error

PurpShell: Now?

tofers: it works stably

tuyuribr (approved):

purpshell (approved):
Looks good

Issue #728, opened 2021-07-07
E: forward a received message


glbatistabh:
### Bug description

When forwarding a received message to another WhatsApp number it is giving the message **"Error: Evaluation failed: TypeError: Cannot read property 'forwardMessages' of undefined"**

> (node:121495) UnhandledPromiseRejectionWarning: Error: Evaluation failed: TypeError: Cannot read property 'forwardMessages' of undefined
>     at __puppeteer_evaluation_script__:5:31
>     at ExecutionContext._evaluateInternal (/home/baptist/devel/whats/whats-api/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:217:19)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:95:5)
>     at async ExecutionContext.evaluate (/home/baptista/devel/whats/whats-api/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:16)
>     at async Message.forward (/home/baptista/devel/whats/whats-api/node_modules/whatsapp-web.js/src/structures/Message.js:226:9)
>     at async Client.<anonymous> (/home/baptista/devel/whats/whats-api/core/server/config-events.js:160:5)

### Reproduction steps
1. create an observer "client.on('message' )'"
```
 client.on('message', async (message) => {
    let forwardTo = '55319239XXXX@c.us'; //part of the number was purposely hidden
    await message.forward(forwardTo);
  });
```
2. send a message from another WhatsApp to WhatsApp connected to whatsapp-web.js;

### Expected behavior
Expected the message to be forwarded to the other whastApp number (numberTo)

### Environment (please complete the following information):
**Whatsapp**
 - Account type: Standard:
 - Device OS: Android:
 - WhatsApp Web version: 2.2123.8

**Library**
 - Browser : chromium:
 - NodeJS: v14.17.2
 - npm or yarn version: 6.14.13
 - whatsapp-web.js version: "^1.12.6"

**PrintScreen**
Code:
![image](https://user-images.githubusercontent.com/8257020/124767591-5398ef80-df0e-11eb-9f23-69c5b6d83bb1.png)

Console:
![image](https://user-images.githubusercontent.com/8257020/124766974-c6559b00-df0d-11eb-8bb1-6a9e26169e38.png)

glbatistabh:
Nobody? `message.forward(chat)` works for you"?

[...]

PR #1624, opened 2022-08-04
updating forward documentation.

NowDev:
This PR addresses issue https://github.com/pedroslopez/whatsapp-web.js/issues/728
Basicaly `Message.forward` method will fail if you never talked with the contact that you will forward the message before.
I don't think that this is a issue to be adressed by wa-web.js since it's probably some anti-spam security from WhatsApp but at least it  _should be documented._

### DIFF
--- a/docs/Message.html
+++ b/docs/Message.html
@@ -420,7 +420,7 @@
               </dl>
               <div class="symbol-detail-labels"><span class="label label-async">async</span></div>
               <h3 id="forward"><span class="symbol-name">forward</span><span class="signature"><span class="signature-params">(chat)</span>&nbsp;&rarr; <span class="signature-returns"> Promise</span></span></h3>
-              <p>Forwards this message to another chat</p>
+              <p>Forwards this message to another chat (that you chatted before, otherwise it will fail)</p>
               <section>
                 <h4>Parameter</h4>
                 <table class="jsdoc-details-table">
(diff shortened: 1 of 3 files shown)


wictornogueira:
Do not update documentation directly.
This is probably what you are looking for
https://github.com/pedroslopez/whatsapp-web.js/blob/main/src/structures/Message.js#L359

PurpShell:
> This PR addresses issue #728 Basicaly `Message.forward` method will fail if you never talked with the contact that you will forward the message before. I don't think that this is a issue to be adressed by wa-web.js since it's probably some anti-spam security from WhatsApp but at least it _should be documented._

As @wictornogueira said, the docs file isn't to be tampered with directly since it is automatically generated on build.. What you should be editing is the JSDoc/typescript declaration

docs/Message.html
```diff
@@ -420,7 +420,7 @@ <h3 id="downloadMedia"><span class="symbol-name">downloadMedia</span><span class
               </dl>
               <div class="symbol-detail-labels"><span class="label label-async">async</span></div>
               <h3 id="forward"><span class="symbol-name">forward</span><span class="signature"><span class="signature-params">(chat)</span>&nbsp;&rarr; <span class="signature-returns"> Promise</span></span></h3>
-              <p>Forwards this message to another chat</p>
+              <p>Forwards this message to another chat (that you chatted before, otherwise it will fail)</p>
```
PurpShell, line 423:
```suggestion
              <p>Forwards this message to another chat</p>
```

purpshell (approved):

tuyuribr (approved):

rizki5411:
How to forward Massage Quote to another number?

PR #1641, opened 2022-08-10
Adjusting dependency validations

jtouris:
If any of the optional dependencies is missing, then it should throw the error, not if only all of them are missing.


tuyuribr:
If any dependency fails it will set all to undefined:

```

try {
    var fs = require('fs-extra');
    var unzipper = require('unzipper');
    var archiver = require('archiver');
} catch {
    fs = undefined;
    unzipper = undefined;
    archiver = undefined;
}
```

PR #1643, opened 2022-08-11
Add "sendMessageByNumber" function

JBunCE:
I did this little thing, I think that is much easy to use 👍🏻


shirser121:
Why don't you use `client.sendMessage`?

tuyuribr:
You should pig back your function in sendmessage function, but I don't find this function useful rn

SrSalchicha:
> You should pig back your function in sendmessage function, but I don't find this function useful rn

Idk, what if you dont want respond a message and you want sent the message directly? In that case you only have the phone number but not the id, and that can sound like a spam but has a múltiple aplications

PR #1651, opened 2022-08-15
Use quick reply

shirser121:


src/util/Injected.js
```diff
@@ -328,40 +348,64 @@ exports.LoadUtils = () => {
         if (!buttonsOptions.buttons) {
             return returnObject;
         }
-
+        
+        if (typeof buttonsOptions.useTemplateButtons === 'undefined' || buttonsOptions.useTemplateButtons === null) {
+            buttonsOptions.useTemplateButtons = buttonsOptions.buttons.some((button) => {
+                return 'callButton' in button || 'url' in button;
```
PurpShell, line 354:
```suggestion
                return 'callButton' in button || 'urlButton' in button;
```

purpshell (changes requested):
This and also add the ReplyButtonModel and ButtonCollection to store (I removed since they weren't being used

PR #1659, opened 2022-08-17
fix: EPERM: operation not permitted, rmdir => Windows & RemoteAuth compatibility

jtouris:
For Windows users, RemoteAuth was occasionally throwing error `EPERM: operation not permitted, rmdir` on `deleteMetadata` when executing backup Syncs to the database.

Adding these catch blocks will avoid the process to crash and the session will be synced correctly without issues.


shirser121 (approved):
LGTM.
Thanks a lot

tuyuribr (approved):
LGTM

purpshell (approved):
looks good to me

PR #1687, opened 2022-09-09
Update package.json

justoverclockl:
This package need to be update due to some error with chromium.


tuyuribr:
Updating pupetter is a bit of issue because it will require greater node versions ( currently we work with node 12 )

PurpShell:
as said by @tuyuribr, we don't have interest in that right now

PR #1688, opened 2022-09-10
Allow the deletion of others' messages

### LINKED ISSUE #1683: [SUGGESTION] Delete others messages

WhatsApp has added the feature to delete others messages as an admin, and using message.delete(true) does not seem to work for others messages

Making message.delete work for other members as long as you have permission to

_No response_

### Additional context

_No response_

soudblox:
Removing the fromMe check in the `message#delete` method does the job

I have tested it and it does indeed work

Does not show in WhatsApp desktop UWP though, because it's not supported there yet

fixes #1683

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -430,14 +430,14 @@ class Message extends Base {
 
     /**
      * Deletes a message from the chat
-     * @param {?boolean} everyone If true and the message is sent by the current user, will delete it for everyone in the chat.
+     * @param {?boolean} everyone If true and the message is sent by the current user or the user is an admin, will delete it for everyone in the chat.
      */
     async delete(everyone) {
         await this.client.pupPage.evaluate((msgId, everyone) => {
             let msg = window.Store.Msg.get(msgId);
 
             if (everyone && msg._canRevoke()) {
-                return window.Store.Cmd.sendRevokeMsgs(msg.chat, [msg], {type: 'Sender'});
+                return window.Store.Cmd.sendRevokeMsgs(msg.chat, [msg], { type: msg.id.fromMe ? 'Sender' : 'Admin' });
             }
 
             return window.Store.Cmd.sendDeleteMsgs(msg.chat, [msg], true);
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -436,7 +436,7 @@ class Message extends Base {
         await this.client.pupPage.evaluate((msgId, everyone) => {
             let msg = window.Store.Msg.get(msgId);
 
-            if (everyone && msg.id.fromMe && msg._canRevoke()) {
+            if (everyone && msg._canRevoke()) {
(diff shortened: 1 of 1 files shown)


sergiooak:
Have you tested? To me it does't work.

src/structures/Message.js
```diff
@@ -436,7 +436,7 @@ class Message extends Base {
         await this.client.pupPage.evaluate((msgId, everyone) => {
             let msg = window.Store.Msg.get(msgId);
 
-            if (everyone && msg.id.fromMe && msg._canRevoke()) {
+            if (everyone && msg._canRevoke()) {
                 return window.Store.Cmd.sendRevokeMsgs(msg.chat, [msg], {type: 'Sender'});
```
sergiooak, line 440: Change the type between "Sender" and "Admin" depending is `msg.id.fromMe`
`return window.Store.Cmd.sendRevokeMsgs(msg.chat, [msg], {type: msg.id.fromMe ? 'Sender' : 'Admin'});`

sergiooak (changes requested):

sergiooak:
Tested with this recommended change, and now it works
image

src/structures/Message.js
ZeroDark1712: Captura de pantalla 2022-09-11 145755
Works!

ZeroDark1712 (approved):

soudblox:
> Have you tested? To me it does't work.

Weird, for me it worked

src/structures/Message.js
soudblox: > Change the type between "Sender" and "Admin" depending is `msg.id.fromMe` `return window.Store.Cmd.sendRevokeMsgs(msg.chat, [msg], {type: msg.id.fromMe ? 'Sender' : 'Admin'});`

Alright

soudblox:
> Tested with this recommended change, and now it works, still need to implement a validation is chat adm image

I should also add a check if it's a group chat or not

purpshell (approved):

tuyuribr (approved):

PurpShell:
> > Tested with this recommended change, and now it works, still need to implement a validation is chat adm image
> 
> I should also add a check if it's a group chat or not
> That would require me to fetch the group first, then the participants to check if the user is a super admin or not

You can simply get the chat (group) and loop through the participants in the browser plane.

sergiooak:
The _canRemoveke do t exclude the need to check if it is a group?

On private the _canRevoke will be false for non sender message anyways.

I suppose that all use cases are already covered with the current if statement

soudblox:
> The _canRemoveke do t exclude the need to check if it is a group?
> 
> On private the _canRevoke will be false for non sender message anyways.
> 
> I suppose that all use cases are already covered with the current if statement

I'm not particularly familiar with the WhatsaApp Web API so I don't really know a lot of private functions that exist

soudblox:
@PurpShell Is there anything else that I have to do?

shirser121:
@PurpShell I think we can merge it

PR #1689, opened 2022-09-10
Battery status for Multi-Devices

RC047:
I tried using the navigator successfully for multi devices, but what is sent is the battery status of the device used to run the browser (puppeteer).


tuyuribr (changes requested):
We need a catch for browsers that don't support getBattery(), and we need to describe this function in index.d.ts

PurpShell:
> 

yeah so pretty useless here, access it through client.pupPage though

PR #1706, opened 2022-09-25
checking if there is an exist wa web page while connecting to an existing chrome instance

musafiroon:
If I have created an instance of chrome that has WhatsApp web completely loaded, then it shouldn't load a new page


purpshell (approved):
👍

purpshell (changes requested):
ESLint

PR #1712, opened 2022-09-27
fix: check if there is an existing wa web page while connecting to an existing chrome instance

numandev1:


PurpShell:
duplicate

numandev1:
@PurpShell other one is for older version

PR #1731, opened 2022-10-16
group_admin event is added

alechkos:


tuyuribr:
How about, one event "group_admin_change"?

alechkos:
> How about, one event "group_admin_change"?

I agree with you, made the necessary changes.

alechkos:
The event was updated.

PR #1743, opened 2022-10-20
fix: message.getLinks is not a function & chat.isEphemeralSettingOn is not a function

smilingOrange:


cbeltrangomez84 (approved):
Not sure about getRawLinks(), but Ephemeral Settings validations are ok

jalamprea (approved):
Both changes getRawLinks and ephemeralSettings are working properly

pedroslopez:
Thanks! Worked on an updated fix here to properly send ephemeral messages #1752

PR #1744, opened 2022-10-20
fix: message.getLinks is not a function & chat.isEphemeralSettingOn is not a function

Eonus21:

PR #1752, opened 2022-10-21
fix: set ephemeral fields and get links from messages

### LINKED ISSUE #1748: TypeError: chat.isEphemeralSettingOn is not a function

### Describe the bug

i have and error when i try to send a message, i've been working with this library like a 15 days, and I had not seen an error like this

### Expected behavior

i expect to send a message an send a media file

1. run application
2. wait to scan the qr
3. on ready event, try to send a message

### Relevant Code

client.on("ready", async () => {
	console.log("connected");
	const media = MessageMedia.fromFilePath("./build/CALENDARIO.xlsx");
	await client
		.sendMessage(
			`${process.env.COUNTRY_CODE}${process.env.NUMBER4}@c.us`,
			media
		)
		.catch((error) => console.log(error));
});

### Browser Type

Google Chrome

### WhatsApp Account Type

Standard

Yes, I am using Multi Device

### Environment

node.js

### Additional context

_No response_

pedroslopez:
## Description

Fixes compatibility with WhatsApp Web v2.2240.7

Takes some inspiration from #1744 #1743, but these were not compatible with older versions of NodeJS.
Instead, use WhatsApp's built-in method for getting ephemeral fields.

## Related Issue

close #1748

### DIFF
--- a/src/util/Injected.js
+++ b/src/util/Injected.js
@@ -363,7 +363,7 @@ exports.LoadUtils = () => {
 
         msg.isEphemeral = message.isEphemeral;
         msg.isStatusV3 = message.isStatusV3;
-        msg.links = (message.getLinks()).map(link => ({
+        msg.links = (message.getRawLinks()).map(link => ({
             link: link.href,
             isSuspicious: Boolean(link.suspiciousCharacters && link.suspiciousCharacters.size)
         }));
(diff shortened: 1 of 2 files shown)


pedropuppim:
thankssssssssssssssssssssssssssss @pedroslopez

PR #1847, opened 2022-11-26
feat(reliability): WhatsApp Web version caching

pedroslopez:
# Description

Enables loading a specific whatsapp web version. This would mean we can target a specific whatsapp web version with a specific library version, solving our issues with having to always play catch up with WhatsApp whenever they make breaking changes.

This works by retrieving the index.html from a remote cache (which references a specific entrypoint). WhatsApp keeps old cached JS files in their CDN for a long time, so replacing the index works even after they've released new versions.

## How to use

This introduces two new client options:

- `webVersion`, which indicates which version of WhatsApp Web should be retrieved
- `webVersionCache`, which specifies options for how the cache is set up / where to retrieve the version

There are three implementations of web version caches included:

### (Default) Local

Stores the version in the local file system the first time its seen. Next time when there's an update, we'll retrieve it from the filesystem instead of using the latest whatsapp web version.

This is currently set by default so you don't need to do this, but you can explicitly configure the local cache like so:

```
webVersionCache: {
    type: 'local'
}
```

By default this is saved to the `./.wwebjs_cache` dir, but you can change this with the `path` option.

The downside to this one is that you need to have run the client at least once for the cache to be populated (or manually grab it from somewhere else and put it in the right directory). This still means if someone happens to download an older library version that has a breaking change, or if the client isn't running on a persistent file system, we'll still fall back to the latest version and have issues.

### Remote

Retrieves the version from some remote web server. This is pretty flexible and _does not include a default  host configured_. We might revisit this later on and make this the default since it solves the issues mentioned with the Local cache. 

Here's an example configuration using the [wa-version archive](https://github.com/wppconnect-team/wa-version/):

```
webVersionCache: {
    type: 'remote',
    remotePath: 'https://raw.githubusercontent.com/wppconnect-team/wa-version/main/html/{version}.html',
},
```

Since this is so generic, `remotePath` lets you configure how the url is structured. `{version}` will be replaced at runtime with the requested version (the one passed in `options.webVersion`).

### None

Disables any caching functionality. It's as if this feature never existed.

Usage:

```
webVersionCache: {
    type: 'none'
}
```

## Strict mode

Both the `local` and `remote` caches have a `strict: true` option available that will throw an exception if the requested version can't be found in the cache. By default this is false and will fall back to the latest version.


matricce:
When the session opened in the latest version and I switch to an older version (e.g. 2.2245.9 to  2.2244.6) a database error occurs and it goes to the QR screen, sometimes when I clear the cache before downgrading it seems to work

pedroslopez:
> 
> 
> When the session opened in the latest version and I switch to an older version (e.g. 2.2245.9 to  2.2244.6) a database error occurs and it goes to the QR screen, sometimes when I clear the cache before downgrading it seems to work

Yep, I've seen that as well. Some updates make changes to the database that can't be reversed, so downgrading will most likely not be explicitly supported. This would mostly be useful for deciding when to update and making sure that everything is functional and compatible with the new version before everyone updating.

tryvin:
@pedroslopez do you need any help here?

matricce:
Using this implementation:
https://github.com/wppconnect-team/wppconnect/blob/33135c09251f2789911321652aec130be306b7d9/src/controllers/browser.ts#L65
https://github.com/wppconnect-team/wa-version/blob/966c554b705d01e8b6b7ec7de834721088a74cb7/src/getPageContent.ts#LL29
...
https://github.com/matricce/whatsapp-web.js/commit/2b6afeded4e25d2d86347e0f5829bacfa7ddf158

pedroslopez:
@matricce thanks for letting me know about wa-version! Added support for it - check out the PR description for how to hook it up 😄

PR #1882, opened 2022-12-22
feat: Implement Call Rejection + Introduction of Socket API

purpshell:
# PR Details

This PR implements the internal functions to send messages to the socket & as an example of how this tech can be used, I've implemented call rejection via the socket (the same exact message sent from the Desktop app when rejecting a call). 

## Description

I've added the Socket & Wap (not to be confused with the old Wap) to the Store.
I've added a reject method to the call class.

## Related Issue
None.

## Motivation and Context

It allows for the versatility of having both high-level WhatsApp code access and the underlying Socket protocol. This allows us to do things the Web version of the app couldn't, for example things that the Desktop app can.

## How Has This Been Tested

I've spent a few hours testing the stability of everything and it is OK.

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1326,6 +1326,9 @@ declare namespace WAWebJS {
         webClientShouldHandle: boolean,
         /** Object with participants */
         participants: object
+
+        /** Reject the call */
+        reject: () => Promise<void>
     }
 
     /** Message type List */
--- a/src/structures/Call.js
+++ b/src/structures/Call.js
@@ -62,7 +62,15 @@ class Call extends Base {
         
         return super._patch(data);
     }
-    
+
+    /**
+     * Reject the call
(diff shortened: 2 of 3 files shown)


Aliyss:
Not sure, but the link preview stuff should probably be in another pull request

PurpShell:
> Not sure, but the link preview stuff should probably be in another pull request

The old Wap and the new Wap are very different. The new one is just a helper utility for Jids & producing whatsapp xml messages in a proper format. The old one was just a general utility and is now gone

PurpShell:
i think I'll just distinguish them

aliyss (approved):
lgtm

tuyuribr (approved):

PR #1907, opened 2023-01-07
Automaticaly format audio to ogg/opus

Helvio88:
# PR Details

Sending audio through wwebjs is not very reliable, as WhatsApp needs a very specific format

## Description

Added ffmpeg conversion in the Util class and automatic detection of mime for efficient conversion in sendMessage()

## Related Issue

https://github.com/pedroslopez/whatsapp-web.js/issues/1625
https://github.com/pedroslopez/whatsapp-web.js/issues/1584
https://github.com/pedroslopez/whatsapp-web.js/issues/1813

## Motivation and Context

I wanted to seamlessly send an audio via wwebjs like it happens with Stickers.
I am using ffmpeg to convert an mp3 file into ogg/opus and it works.
Wanted everyone to benefit from it, too.

## How Has This Been Tested

I have not tested the code I sent you, but I have done the conversion in a similar fashion in my code, essentially proxying all the Audio and PTT messages through it. It works

## Types of changes

- [ ] Dependency change
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly (index.d.ts).

Issue #1813, opened 2022-11-16
An error will occur when sending an audio message to the webpage version of whatsapp


hejie615:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The audio message can be sent successfully, but the webpage whatsapp cannot be played and the person who sent the audio message cannot be displayed after refreshing the page

### Expected behavior

It is expected that the audio can be sent successfully and played

### Steps to Reproduce the Bug or Issue

After the webpage version whatsapp is refreshed, the following error will be reported

vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27 
TypeError: Cannot read properties of undefined (reading 'toString')
    at t.default (bootstrap_main.a4fb49fc7d9de22241e5.js:2:534066)
    at ia (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:59075)
    at $s (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:111314)
    at Tu (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:98520)
    at ku (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:98448)
    at Eu (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:98311)
    at vu (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:95298)
    at vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:44970
    at t.unstable_runWithPriority (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:51:3768)
    at Yi (vendor1~bootstrap_qr.a5a17f05e19d85ba8f6c.js:27:44747)

vendor1~bootstrap_qr.ef018a7bfbd9257ea51d.js:27 
TypeError: Cannot read properties of undefined (reading 'length')
    at Object.matchLargeEmojiPattern (bootstrap_qr.05be0932a7796d566db1.js:75:155613)
    at t.default (bootstrap_main.aff103e5f4f324a15f2c.js:3:207799)
    at ia (vendor1~bootstrap_qr.ef018a7bfbd9257ea51d.js:27:59075)
    at Ws (vendor1~bootstrap_qr.ef018a7bfbd9257ea51d.js:27:111314)
[...]

Issue #1625, opened 2022-08-05
Error when sending mp3 files


Aromakelapa:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Describe the bug

Hello sir, i got an error when trying to send mp3
### Error :
```
✓ Connected!
Incoming Messages : 
💬 . : !ytmp3 https://www.youtube.com/watch?v=RyZSKarK78U
Sending RyZSKarK78U.mp4a
/home/lmint1404/tester/romBot-Whatsapp/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221
            throw new Error('Evaluation failed: ' + helper_js_1.helper.getExceptionMessage(exceptionDetails));
                  ^

Error: Evaluation failed: d
    at ExecutionContext._evaluateInternal (/home/lmint1404/tester/romBot-Whatsapp/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async ExecutionContext.evaluate (/home/lmint1404/tester/romBot-Whatsapp/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:110:16)
    at async Client.sendMessage (/home/lmint1404/tester/romBot-Whatsapp/node_modules/whatsapp-web.js/src/Client.js:599:28)
[nodemon] app crashed - waiting for file changes before starting...
```

### Expected behavior

got an error when trying to send mp3

### Steps to Reproduce the Bug or Issue

1. Getting link
2. Getting info
3. Download mp3
4. Send mp3
5. See error

### My sc

```js
else if (msg.body.startsWith('!ytmp3 ')) {
    msg.reply('_Woke diproses!_')
    const link = msg.body.slice(7)
    const id = link.split('=')[1]

    async function ytmp3(link){
	const getIngfo = ytdl.getBasicInfo(link).then(info => {
	    return {
	        thumb: info.videoDetails.thumbnails[3]['url'],
	        author: info.videoDetails.ownerChannelName,
	        view: info.videoDetails.viewCount,
	        title: info.videoDetails.title,
	        likes: info.videoDetails.likes
	    }
        })

	function downDir() {
            if (!fs.existsSync('./downloads/')){
                fs.mkdirSync('./downloads')
            } else { }
        } downDir()

	const ingfo = await getIngfo

[...]

Issue #1584, opened 2022-07-12
Send Audio file not working


Guilherme-j10:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Evaluation failed: d

### Expected behavior

That the voice message be send

### Steps to Reproduce the Bug or Issue

run code

### Relevant Code

![image](https://user-images.githubusercontent.com/55926786/178547328-6843ef56-b2b2-4ef1-918a-5c0d98119256.png)

### Browser Type

Chromium

### WhatsApp Account Type

WhatsApp Business

### Does your WhatsApp account have multidevice enabled?

Yes, I am using Multi Device

### Environment

whatsapp-web.js version: 1.17.1

stefanfuchs:
Duplicated: #1579

PR #1956, opened 2023-01-26
Automaticaly format audio to ogg/opus

Helvio88:
# PR Details

Re-creation of PR #1907
Sending audio through wwebjs is not very reliable, as WhatsApp needs a very specific format

## Description

Added ffmpeg conversion in the Util class and automatic detection of mime for efficient conversion in sendMessage()

## Related Issue

#1625 
#1584 
#1813 

## Motivation and Context

I wanted to seamlessly send an audio via wwebjs like it happens with Stickers.
I am using ffmpeg to convert an mp3 file into ogg/opus and it works.
Wanted everyone to benefit from it, too.

## How Has This Been Tested

I have not tested the code I sent you, but I have done the conversion in a similar fashion in my code, essentially proxying all the Audio and PTT messages through it. It works.

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -683,7 +683,7 @@ class Client extends EventEmitter {
             );
         }
 
-        if (internalOptions.attachment.mimetype.startsWith('audio')) {
+        if (internalOptions.sendAudioAsVoice && internalOptions.attachment.mimetype.startsWith('audio')) {
             internalOptions.attachment = await Util.formatToOpusAudio(internalOptions.attachment);
         }
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -683,6 +683,10 @@ class Client extends EventEmitter {
             );
         }
 
+        if (internalOptions.attachment.mimetype.startsWith('audio')) {
+            internalOptions.attachment = await Util.formatToOpusAudio(internalOptions.attachment)
+        }
+
         const newMessage = await this.pupPage.evaluate(async (chatId, message, options, sendSeen) => {
             const chatWid = window.Store.WidFactory.createWid(chatId);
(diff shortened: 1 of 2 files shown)


Helvio88:
@tuyuribr - I have re-created this PR, because I pushed commits I didn't want to in the other one. I closed #1907.

FYI, all the `ffmpeg` code follows the *exact* guidelines as lines 93-120, including temp file generation, reject and resolve. I would think it's best to keep it this way, for standardization. If changes are required, I'd suggest they be made after this PR is merged, and in all the similar functions in this file.

tofers:
I wouldn't put that into processing, sometimes you just need to send mp3. 

As a help for voice, fine, but I would rather recommend processing audio/video before sending to put less stress on your container load and memory consumption

Helvio88:
I think the approach should be transparent - therefore I added a check to see if `sendAudioAsVoice` is set and convert accordingly. See latest commit.

purpshell (commented):
I suggest more testing is done here, many audio formats can't be moved into opus/ogg with ffmpeg, I've hit such a roadblock before. I suggest you keep on using ffmpeg though, as it has a uniform and expected response.

PurpShell:
> I think the approach should be transparent - therefore I added a check to see if `sendAudioAsVoice` is set and convert accordingly. See latest commit.

Please add more documentation regarding its behavior, readers of the documentation should know what is happening here

tofers:
I apologize of course, maybe I seem rude. But even so, I would have to rewrite my application and write a crutch to get around this conversion.

For two years I have been sending voice messages in mp3 format

PR #1905, opened 2023-01-06
fix: Fix client-side message rendering issue when sending attachements through the library

purpshell:
# PR Details

This PR aims to fix the issues with rendering an attachement sent through WWebJS in the GUI after the reload/restart of the client.

image

## Description

It seems the issue was that the mediaData prop was not getting properly serialized and added to the message object, which has been reflected & changed in MD. Now in MD, (in the code) the prop is also serialized and appended to the msg object in WhatsApp Web code. I have tested all types of attachments and this seems to work properly

## Motivation and Context

This fix will make the UI more enjoyable and appropriate. The issue was in how the message type `"chat"` was not affected after the concatenation of the attachment options, and so on reload the body (the thumbnail) was eventually just rendered as text rather than an image. This change was needed to make the library a better experience overall to people who use the GUI in co-ordinance with the library. This also fixes other issues like the instability of message type after sending a message with attachments.

## How Has This Been Tested

Fully tested on a session with normal WhatsApp on Android. Please try this pull request and help us assure it is fully functioning, it should be though as it now mimics 1:1 how it is sent in WA.

## Types of changes

- [ ] Dependency change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly (index.d.ts).

Issue #1977, opened 2023-02-06
Sending Video mp4 with Caption made the fetchMessages result has no mimetype and type is "chat"


JenJenFenFen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

i tried to send video mp4 with the caption is "test", it works. But when i tried to fetchMessage for seeing the result, i got this

![image](https://user-images.githubusercontent.com/114470554/216880943-521ddbd1-b987-4f1c-83cf-cb0cb073c69b.png)

And the mimetype property is nothing

![image](https://user-images.githubusercontent.com/114470554/216881249-c5316c14-437e-4941-957d-eb277a6bb75d.png)

When i tried to send video mp4 without the caption, the mimetype is shown and the type is "video"

![image](https://user-images.githubusercontent.com/114470554/216881705-25d4ed8b-e2ce-4f97-9a22-752f98aa7e25.png)
![image](https://user-images.githubusercontent.com/114470554/216881740-5f97987a-360c-4960-992f-f83a37b60ac7.png)

### Expected behavior

i need the type changes to "video" and the mimetype is show to "video/mp4

### Steps to Reproduce the Bug or Issue

i have no idea what i should do

### Relevant Code

This is my code

```
const client = new Client({
    authStrategy: new LocalAuth({
        clientId: "client-one"
    }),
    puppeteer: {
        headless: true,
        executablePath: 'C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe' //depends on your chrome app location
    }
})

app.post('/sendmedia', upload.single('file'), async (req, res) => {
    try {
        let number = req.body.number
        let file = req.file

        // validation
        if (number == '' || number == null || file == null) {
            return res.status(400).json({
                'status': 'Gagal',
                'error_message': 'Please check your input'
            })
        }

        let numberPhone = numberPhoneFormat(number)
        const isRegisteredNumber = await checkRegisteredNumber(numberPhone)

        if(!isRegisteredNumber) {
            return res.status(400).json({
                'status': 'Gagal',
                'error_message': 'The number phone is not registered'
            })
        }

        let captionFile = req.body.caption
        let file_name = file.originalname
        let file_mimetype = file.mimetype
        let file_buffer = file.buffer.toString('base64')
        const fileSend = new MessageMedia(file_mimetype, file_buffer, file_name)

        client.sendMessage(numberPhone, fileSend, {caption: captionFile})

        res.status(200).json({
[...]

PR #1981, opened 2023-02-07
Fix media renders

purpshell:
#1905


tuyuribr (approved):

src/util/Injected.js
```diff
@@ -271,6 +271,7 @@ exports.LoadUtils = () => {
             ...ephemeralFields,
             ...locationOptions,
             ...attOptions,
```
StiflerPasha, line 273: And why we need first ...attOptions ?

PurpShell: because the first one includes the original mirrored content and the other has the property that will override the type. Overriding the type is how you fix this bug.

StiflerPasha: I'm not sure, but it looks like necessary. I've tried without '...attOptions' and it's fine. But I don't use stickers. It seems that toJson() or serialize() converts object's fields (e.g. __x_type=> type) and this will not override

matricce:
attOptions will be always true, no?, {} is truthy :thinking: 
obs: optional chaining not supported by node v12

tofers:
I rolled out this fix on my clients

node v 16, but I don't use stickers until I've found a use for them.
If I notice any bugs I will let you know.

tofers:
#1977

src/util/Injected.js
```diff
@@ -271,6 +271,7 @@ exports.LoadUtils = () => {
...
+            ...(attOptions?.toJSON ? attOptions.toJSON() : {}),
```
PurpShell, line 274:
```suggestion
            ...(attOptions.toJSON ? attOptions.toJSON() : {}),
```

shirser121:
Should we merge it? @pedroslopez

PurpShell:
@Aliyss @pedroslopez

aliyss (approved):
lgtm

PR #2098, opened 2023-03-25
Get stories

Gimenz:
a simple feature to get contact stories

Tanpa Judul


src/Client.js
```diff
@@ -1178,6 +1178,19 @@ class Client extends EventEmitter {
...
         return blockedContacts.map(contact => ContactFactory.create(this.client, contact));
     }
+
+    /**
+     * Get stories
+     * @returns {Promise<Message[]>}
+     */
+    async getStories() {
+        const messages = await this.pupPage.evaluate(async () => {
+            const models = window.Store.Msg._models.filter(x => x.__x_id.remote.server == 'broadcast');
```
tuyuribr, line 1188: Can we filter per contact?

example.js
```diff
@@ -40,6 +40,18 @@ client.on('message', async msg => {
...
+    } else if (msg.body.startsWith('!getstory ')) {
+        // let number = await msg.getMentions()[0]
+        const stories = await client.getStories();
+        let number = msg.body.split(' ')[1];
+        number = number.includes('@c.us') ? number : `${number}@c.us`;
+        let filtered = stories.filter(v => v.author == number);
+        msg.reply(`found ${filtered.length} stories`);
+        
+        for (let i of filtered) {
+            await sleep(2000);
```
tuyuribr, line 52: Change this

tuyuribr (changes requested):
index.d.ts is missing

example.js
Gimenz: omgg, i forgot to delete sleep func

did you mean filter per contact by adding this example into getStories function?

tuyuribr: Can we use something like getStories('551199991234@c.us') ?

purpshell (changes requested):
use the StatusV3 module to do all this, you aren't forced to use Msg

Koceng404:
bro kalo biar chat Groupnya Ke filter Gimana ya? jadi semua Fitur hanya bisa dipake kalo ga di chat di group?

purpshell (changes requested):
You can still just get all models without specifying an id though right? just `.map()` on `_models` and you should be good

example.js
PurpShell: yes, and no number returns everything.

PR #2109, opened 2023-03-31
FIX #2101 - only process supported images

### LINKED ISSUE #2101: Fail when sending .dwg files (don't know of others)

### Describe the bug

When sending files from Autodesk Autocad (dwg extension) whatsapp-web.js fails with the following error:

```
Error: Evaluation failed: l
    at ExecutionContext._evaluateInternal (/home/allgood/Projetos/whatsapp-web.js/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (/home/allgood/Projetos/whatsapp-web.js/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:110:16)
    at async Client.sendMessage (/home/allgood/Projetos/whatsapp-web.js/src/Client.js:686:28)
```

### Expected behavior

The file being sent

To replicate the error I've cloned the current repository and executed the following commands:

```
npm install
node exec example.js
```

After authorizing the whatsapp access on the device, from another whatsapp number I've sent a .dwg file (tested with the file available on https://www.cadforum.cz/catalog_en/block.asp?blk=7894 ) to the number controlled by whatsapp web and then replied to my own message with the `!resendmedia` command, after this the error showed on the console screen as seen on the following image:

image

### Relevant Code

_No response_

### Browser Type

Chromium

### WhatsApp Account Type

Standard

Yes, I am using Multi Device

### Environment

tested on Debian Linux 11

### Additional context

_No response_

allgood:
# PR Details

When sending files with an `image/*` mime type, treat it as an image only if it have supported image format.

## Description

Added a small function and supported formats list and changed to treat any unsupported image as a document.

## Related Issue

This fixes #2101 issue

## Motivation and Context

Problem when sending files such as autodesk autocad (dwg) the file doesn't get sent.

## How Has This Been Tested

Tested using the same approach to replicate the problem detailed in the issue.

## Types of changes

- [ ] Dependency change

## Checklist

(really didn't updated index.d.ts because I think it is unnecessary, please inform if I'm wrong)

### DIFF
--- a/src/util/Injected.js
+++ b/src/util/Injected.js
@@ -324,10 +324,20 @@ exports.LoadUtils = () => {
         return stickerInfo;
     };
 
+    window.WWebJS.supportedImages = [
+        'image/jpeg',
+        'image/x-png',
+        'image/gif'
+    ];
+
+    window.WWebJS.isUnsupportedImage = function (mimeType) {
+        return mimeType.startsWith('image/') && !window.WWebJS.supportedImages.includes(mimeType);
+    }
+
     window.WWebJS.processMediaData = async (mediaInfo, { forceVoice, forceDocument, forceGif }) => {
         const file = window.WWebJS.mediaInfoToFile(mediaInfo);
         const mData = await window.Store.OpaqueData.createFromData(file, file.type);
-        const mediaPrep = window.Store.MediaPrep.prepRawMedia(mData, { asDocument: forceDocument });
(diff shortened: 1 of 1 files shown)


PurpShell:
instead of having such a locked approach - there are many more types of images that work just fine - you should use the sendMediaAsDocument option and pass true or false to that based on your code

allgood:
Hi @PurpShell ,

where can I find the the supported images list? Would not be possible to force sendMediaAsDocument on whatsapp-web.js if an image type isn't supported?

PR #2166, opened 2023-04-22
Improvements in mentionedJidList and content, referring to sendMessage

rafael-ofc:
All the changes made do not oblige users to modify their codes, but this adds new options when sending a message, simplifying some tasks.

## mentionedJidList
When mentioning a participant, you don't necessarily need to send an array of objects, but you can also send an array of strings, which makes it easier to mention specific users, see the example below:
```js
//Sending messages with mentions
client.sendMessage(msg.from, `Hello @5516****8888`, {mentions: ["5516****8888@c.us"]})
```


NuevaGeneracionALB:
Perfeito 👌

src/structures/MessageMedia.js
shirser121, line 1: Why add fromAll instead of do the check in the sendMessage and call fromUrl or fromFilePath?

rafael-ofc: That's an idea too, but when it comes to software architecture, shouldn't each class have its unique responsibilities?

 Anything else I can do is change here.

shirser121: @pedroslopez @PurpShell what do you think?

PurpShell: Greatly disagree with the use of fromAll. This may seem like a good idea, but it'll bring a lot of exploits that allow RCE and file extortion from the filesystem.

rafael-ofc: > Discordo totalmente do uso de fromAll. Isso pode parecer uma boa ideia, mas trará muitos exploits que permitem RCE e extorsão de arquivos do sistema de arquivos.

Correct, but what about using (fs.existsSync()), do you think it's a good idea to use it?  To know if the path is a url or a folder

rafael-ofc: image
I was thinking of using it like this, directly in sendMessage, what do you think?

If it finds the key url, it executes this code, otherwise it continues normally, being able to use other methods.

PurpShell: Use it in your own code

purpshell (changes requested):
Split into more than one pull request

rafael-ofc:
> Dividido em mais de um pull request

What is to be changed?

src/structures/MessageMedia.js
NuevaGeneracionALB: > Use-o em seu próprio código

Difficult the situation of the project, I am thinking of using it for a while and analyzing the response data of the developers of whatsapp-web.js, but do not apply the idea of almost any new developer, and when they launch the idea already ready, they discard the revolution of the project and ask for use in the code itself, and the idea is for the advancement of the project and not itself,  Everything that is created is to maintain stability and also advancement.

src/Client.js
```diff
@@ -2,6 +2,7 @@
 
 const EventEmitter = require('events');
 const puppeteer = require('puppeteer');
+const fs = require('fs');
```
PurpShell, line 5:
```suggestion
```

src/Client.js
```diff
@@ -751,12 +752,17 @@ class Client extends EventEmitter {
...
             extraOptions: options.extra
         };
 
         const sendSeen = typeof options.sendSeen === 'undefined' ? true : options.sendSeen;
 
+        if (content.url) {
+            let { url, ...contentOptions } = content;
+            content = fs.existsSync(url) ? await MessageMedia.fromFilePath(url) : await MessageMedia.fromUrl(url, contentOptions);
+        }
+
```
PurpShell, line 765:
```suggestion
```

purpshell (changes requested):
Just revert the fromAll functionality I think. Too many security exploits will be caused by this

purpshell (approved):

Aliyss:
If we do this it may make sense, that we let the user handle the map serialized and only allow for serialized ids, without the contact object?

@PurpShell

PurpShell:
> If we do this it may make sense, that we let the user handle the map serialized and only allow for serialized ids, without the contact object?
> 
> @PurpShell

Yes, i think so, to improve code quality and allow for more options and less errors.. Getting the contacts is mostly a hindrance and not an enjoyable time

PurpShell:
@Aliyss updated

aliyss (approved):
lgtm

Aliyss:
>  If we decided to switch to TS at the time of the TS declarations being added, that would've been much easier to maintain than to have a TS declarations file and an outdated JS config.

 +1 to everything. Lets switch to TS!

loic-hamdi:
`Mentions with an array of Contact are now deprecated. See more at https://github.com/pedroslopez/whatsapp-web.js/pull/2166.`


`index.d.ts` has not been updated, it still ask for `array of Contact`

```ts
/** Contacts that are being mentioned in the message */
        mentions?: Contact[]
```

amirave:
bumping @loic-hamdi.

PR #2190, opened 2023-05-09
using puppeteer with stealth plugin to avoid detection

maikelohcfg:
# PR Details

Introduce using of Stealth Plugin with puppeteer to minimize chances of being detected

## Description

Puppeteer's plugins provide a lot of functionality to improve puppeteer, one of them is the Stealth, here we use it to make harder to detect the automated browser

## Motivation and Context

This change will reduce the risk of detection, which should decrease the number of account banned. This is not a magic solution, sending a lot of messages is still something against WhatsApp policies, so use it wisely

## How Has This Been Tested
Just instantiate the browser, the WhatsApp client as usually. The change is just related to how we launch puppeteer using the proper plugin

## Types of changes
- [x] Dependency change

## Checklist

- [x ] I have updated the documentation accordingly (index.d.ts).


tofers:
Think of the solution as a separate module so you don't depend on it. 

Among those who need it will use

Ermesoml:
Is that possible to use it as a separated module on the wwebjs?
I'm also interested but I agreed that it could be better as a separated module.

PurpShell:
Here is what I suggest you do, keep the change you need to make (require puppeteer-extra) and simply require it from your script and add any extensions you need. This way it is not a forced decision.

src/Client.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 const EventEmitter = require('events');
-const puppeteer = require('puppeteer');
+const puppeteer = require('puppeteer-extra');
+const StealthPlugin = require('puppeteer-extra-plugin-stealth');
```
PurpShell, line 5:
```suggestion
```
add this on your side

src/Client.js
```diff
@@ -89,6 +90,7 @@ class Client extends EventEmitter {
      * Sets up events and requirements, kicks off authentication request
      */
     async initialize() {
+        puppeteer.use(StealthPlugin());
```
PurpShell, line 93:
```suggestion
```
add this on your side as well

purpshell (changes requested):

kaiser620:
I got this error while trying to implement this Puppeteer stealth using this way and making the changes. Error: Navigating frame was detached.

PR #2250, opened 2023-06-03
Add search option on chat.fetchMessages by min/max timestamp

LucianoHanna:


LucianoHanna:
Reopening #1499 with ESLint fixed and index.d.ts doc added

tolotrasamuel (commented):
MessageSearchOptions type definition is not updated

index.d.ts
```diff
@@ -1113,7 +1113,15 @@ declare namespace WAWebJS {
         clearState: () => Promise<boolean>,
         /** Deletes the chat */
         delete: () => Promise<boolean>,
-        /** Loads chat messages, sorted from earliest to latest. */
+        /**
+         * Loads chat messages, sorted from earliest to latest.
+         * @param {Object} searchOptions Options for searching messages. Right now only limit is supported.
+         * @param {Number} [searchOptions.limit] The amount of messages to return. If no limit is specified, the available messages will be returned. Note that the actual number of returned messages may be smaller if there aren't enough messages in the conversation. Set this to Infinity to load all messages.
+         * @param {Number} [searchOptions.minTimestamp] If specified, filter message by minimum timestamp.
```
tolotrasamuel, line 1120: Probably minimum can be rephrased to oldest and maximum earliest

tolotrasamuel: also, millis or unix

src/structures/Chat.js
```diff
@@ -190,19 +192,33 @@ class Chat extends Base {
...
                     if (!loadedMessages || !loadedMessages.length) break;
+
+                    if (searchOptions.minTimestamp && loadedMessages.some(msg => msg.t < searchOptions.minTimestamp))
+                        break;
+
                     msgs = [...loadedMessages.filter(msgFilter), ...msgs];
                 }
-                
+
                 if (msgs.length > searchOptions.limit) {
```
tolotrasamuel, line 222: Extraneous chats with timestamp beyond range are handled with splice  ?

LucianoHanna: why would this be a problem? can you elaborate?

index.d.ts
LucianoHanna: i personally prefer "minimum timestamp" but I can change it...

src/structures/Chat.js
tolotrasamuel: So the loop breaks before we insert the loadedMessages that was filtered, some messages might be within the range but not returned, maybe line 219 should go first . Unless I missed something

LucianoHanna: You are right. I will fix it.

shirser121: @LucianoHanna where we standing with this?

ivanseidel:
I just spent the last 4 hours debugging an odd behavior, that limits fetching new messages from the device and forces the user to "click" the button to fetch new messages from device.

What I can resume from my findings:

## There are 3 Layers of data:
- In memory messages (internally used as `chat.msgs.getModelsArray`)
- In local db messages (internally used as `Store.ConversationMsgs.loadEarlierMsgs`)
- **External messages (internally called `history sync on demmand`)**: This is where the problem resides

## Looking into what actually happens is as follows:
- When the interface loads up a specific conversation, just 20 messages are pulled from local db and loaded into memory.
- When user scrolls, and reaches the end, more messages are pulled from local db to memory.
- When user reaches the LIMIT, no data is returned from `loadEarlierMsgs` and a button appears so the user can force a `history sync` from the device.

The thing is, this method is not whell documented, but I have figured out how to call it and force it to load more messages, and even, be able to change "how many messages" to fetch from device at once.

## The steps are (assuming `moduleRaid` exists):

**Step 0:** (optional) set how many messages to force a `history sync`
```javascript
mR.findModule('bulkCreateOrReplaceABPropConfigs')[0].bulkCreateOrReplaceABPropConfigs([{
    "configCode": 3811,
    "configValue": 1000, // Default value is 50
    "configExpoKey": null,
    "hasAccessed": false
}])
```

**Step 1:** Trigger a History sync on demand
```javascript
const HISTORY_SYNC_ON_DEMAND = 3 // Operation code declared a few times in source code
await mR.findModule('sendPeerDataOperationRequest')[0].sendPeerDataOperationRequest(HISTORY_SYNC_ON_DEMAND, {
    "chatId": mR.findModule('createWid')[0].createWid('5511999999999@c.us') // The WID to Sync on demand goes here
})
```

**Step 2:** Use loadEarlierMsgs after finished (notice that this function is async, but doesn't wait until completion to return, need to investigate further)

**Step 3:** Read from getModelsArray as usual

This should be enough to force a on demand sync of a specific chat.

Probably, we should consider this logic into the fetchMessages, otherwise, new messages will never get read from remote device

flawnn:
Why is this not merged yet?

tuyuribr:
Stale

PR #2254, opened 2023-06-06
feat: Edit message

tofers:
# PR Details

Added message editing from version "2.2323.4"

## Description

Currently, message editing is only possible for text type messages. But there are other types in Whatsapp code that will be available later.

Settings are put in a separate type, we will update it as the functionality fills up.

Instructions Install the branch:
```
NPM: npm i github:tofers/whatsapp-web.js#editMsg
Yarn: yarn add github:tofers/whatsapp-web.js#editMsg
```
Message change event
```
client.on('message_edit', async (msg, newBody, prevBody) => {
    console.log('message', msg );
    console.log('message text', prevBody, newBody);
});
```

Example usage
```
client.on('message', async msg => {
    console.log('MESSAGE RECEIVED', msg);
    if (msg.body === '!edit') {
        if (msg.hasQuotedMsg) {
            const quotedMsg = await msg.getQuotedMessage();
            if (quotedMsg.fromMe) {
                quotedMsg.edit(msg.body.replace('!edit', ''))
            } else {
                msg.reply('I can only delete my own messages');
            }
        }
    }
});
```

## Related question

Haven't seen anyone ask that yet. But I need this functionality.

## Motivation and Context

To be able to update a message 

## How has this been tested

I tested it on linux systems

## Types of changes

- [ ] Change dependencies
- [ ] Bug fix (non-breaking change that fixes the problem)
- [x] New feature (non-breaching change that adds functionality)
- [x] Violating change (a fix or feature that will change existing functionality)

## Checklist

- [x] My code matches the code style of this project.
- [x] I have updated the documentation accordingly (index.d.ts).

PR #2258, opened 2023-06-07
feat: Edit messages (original) + Fix link previews

purpshell:
# PR Details

you can edit messages, very easily
## Types of changes

- [ ] Dependency change

## Checklist


PurpShell:
> #2254

That one uses a different implementation and rewrites sendMessage for no reason

package.json
```diff
@@ -35,7 +35,8 @@
     "mime": "^3.0.0",
     "node-fetch": "^2.6.5",
     "node-webpmux": "^3.1.0",
-    "puppeteer": "^13.0.0"
+    "puppeteer": "^13.0.0",
+    "link-preview-js": "^3.0.4"
```
tuyuribr, line 39: We shouldn't use this.
1 - it is fairly easy to do something basic.
2 - We should use the waweb way of doing
3 - The lib is no longer maintained

PurpShell: 1. It is the way used by Baileys
2. WhatsApp Way of doing is a React component and some weird Web APIs, and images/info is gotten from the socket using Sync States (you send a sync request for the server to scrape the page for you), trust me I tried for 2 hours
3. That doesn't really matter as the standards that the library uses aren't about to change anytime soon, if so, the API opens access to headers and other configurations

PR #2129, opened 2023-04-09
Upgrade node version

tuyuribr:
# PR Details
It's time to upgade the node version

## Description

The latest ubuntu version, installs node v18. So we can upgrade node and the other dependencies 

[https://packages.ubuntu.com/kinetic/nodejs]

## Warnings
Win7 don't support node 18 ... (only up to 13)

## Types of changes

- [x] Dependency change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## TO DO
- [ ] Quick steps for upgrade
- [ ] Other dependencies update
- [ ] User base test

## Tested
Windows. Node v18.16.0. WaWeb 2.2326.10. [2023-06-24]

PR #2259, opened 2023-06-07
Updated Puppeteer version to avoid warn from npm while installation

M-Umais-Hassan:
# PR Details

Having a warning from npm while installing the library and after that issue, the installations just kept stuck.

### Problem: 
So, when I was trying to download the library I was having a warning from npm, and after having the warning the installation hang up and nothing was happening at all further. 
Here is the warning message that I am having,
<strong>"npm info run puppeteer@13.7.0 install node_modules/whatsapp-web.js/node_modules/puppeteer node install.js"</strong>

### Solution: 
After some debugging, I just found that inside devDependencies we have an old version of puppeteer so I just updated it from 13.6.0 to 20.5.0 and it worked fine after the updation.

## How Has This Been Tested

So, I have tested is by creating an instance of whatsapp web and its all running perfectly fine.

## Types of changes

- [ ] Dependency change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "umais-whatsapp-web.js",
+    "name": "whatsapp-web.js",
     "version": "1.21.0",
     "description": "Library for interacting with the WhatsApp Web API ",
     "main": "./index.js",
--- a/package.json
+++ b/package.json
@@ -1,60 +1,60 @@
 {
-  "name": "whatsapp-web.js",
-  "version": "1.21.0",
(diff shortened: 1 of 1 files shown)


tuyuribr:
#2129

M-Umais-Hassan:
Seems like an acceptable solution, but what I we have a project with an old version setup and don't want to upgrade the node for that. Just thinking a scenario because I just had a similar one, so isn't it a better solution to upgrade the puppeteer and it will work fine hopefully for all the versions.

PR #2289, opened 2023-06-18
adding type to mentions to match with PR 2166

diegofemello:
# PR Details
Adding type mentions to match with #2166

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -913,7 +913,7 @@ declare namespace WAWebJS {
         /** Id of the message that is being quoted (or replied to) */
         quotedMessageId?: string
         /** Contacts that are being mentioned in the message */
-        mentions?: Contact[]
+        mentions?: Contact[] | string[]
         /** Send 'seen' status */
         sendSeen?: boolean
         /** Media to be sent */


codebossdev (approved):

aliyss (changes requested):
@PurpShell I think we only decided on string[] and no contact is that correct?

Edit:

[Deprecation Warning](https://github.com/pedroslopez/whatsapp-web.js/blob/3d7ef5531651f4ace0a4116b9547865b99e0b8d4/src/Client.js#L746)

So according to this, we can actually remove the need for the Contact[] type. Since it's deprecated.

shirser121:
> @PurpShell I think we only decided on string[] and no contact is that correct?
> 
> Edit:
> 
> [Deprecation Warning](https://github.com/pedroslopez/whatsapp-web.js/blob/3d7ef5531651f4ace0a4116b9547865b99e0b8d4/src/Client.js#L746)
> 
> So according to this, we can actually remove the need for the Contact[] type. Since it's deprecated.

So we need to remove that or not?

PR #2295, opened 2023-06-22
feat:  control  auto-loading of files + fixed method setDisplayName

tofers:
# PR Details

Settings management

## Description

Fixed setDisplayName function as well as added several functions that control auto-loading of files

```

"whatsapp-web.js": "github:tofers/whatsapp-web.js#settings"
```

## Motivation and Context

The media_uploaded event no longer works and I have to look for another solution for the files. 

The settings have auto-uploaded media and it is turned off for videos and documents. 
I would like to manage this

image

## How Has This Been Tested

Linux, Centsos. 
When I connect Whatsapp after READY I turn on the options I want

## Types of changes

- [ ] Dependency change

## Checklist


tuyuribr:
Shouldn't we set everything to true by default on startup? And make options instead of functions (is there a scenario the user would change this settings on fly? If there is, shouldn't we do both? By default follow options and them if the user want to change he triggers the function)

tofers:
> Shouldn't we set everything to true by default on startup? And make options instead of functions (is there a scenario the user would change this settings on fly? If there is, shouldn't we do both? By default follow options and them if the user want to change he triggers the function)

Yes, I think we can call them all on at startup. 

Okay, I'll do that.

tuyuribr:
Other question can we do `await message.downloadMedia();` without enabling this settings? if we can, I think the default should be disable the auto download

tofers:
Now I encountered a problem when the `message` event happens and I upload a picture file (it succeeds), but an empty result is returned for the documents (if I repeat the upload later, the file will be uploaded. But it's not for everyone, I haven't caught the dependencies ).

I'll check on my clients that complain about uploading files and report the results. 

There is also an event `change:mediaData.mediaStage
As in telegram which says that the file is ready.  While I'm looking for a solution to upload 100% of the files inbound and outbound

image

mpirescarvalho:
I have tested `msg.downloadMedia()` with this settings set to false and it still works, so IMO this should be set to false by default only on the first startup, leaving the option (the functions) so that developers can enable them if needed

tuyuribr (approved):

shirser121:
Ping me tomorrow i will check this

tofers:
[opssemnik](https://github.com/opssemnik)   I ran into a problem where when sending files to WWeb and receiving the message_create event, the file was simply downloaded using the msg.downloadMedia method. Also incoming files were not downloaded either ( randomly )

My clients were suffering from this and there were many complaints. So I was looking for every possible solution to 100% download files using the downloadMedia method.  I found a solution (I will make a new PR, maybe someone will find it useful) and yes indeed this PR has a small functionality that does not affect the main functionality.

src/Client.js
```diff
@@ -1028,8 +1028,8 @@ class Client extends EventEmitter {
             if(!window.Store.Conn.canSetMyPushname()) return false;
 
             if(window.Store.MDBackend) {
-                // TODO
-                return false;
+                await window.Store.Settings.setPushname(displayName);
```
alechkos, line 1031: seems like that function `window.Store.Settings.setPushname` changes the pushname only locally
i tested the bellow method and works great

```suggestion
                await window.mR.modules[858640].default(displayName);
```

shirser121:
We should resolve the conflicts

tofers:
> We should resolve the conflicts

Thanks for your help. Conflict resolved

alechkos (approved):
Tested on the latest wweb v2.2346.52 and works great!

@pedroslopez @PurpShell @aliyss @shirser121

shirser121 (approved):
Thanks!
Work great

PR #2310, opened 2023-06-29
indicates that the message is sent via wa web automatically

RC047:
# PR Details
Added the option that the message is sent via whatsapp web

## Description
Some of these possibilities can be useful to be able to distinguish which messages are sent via whatsApp web automatically and which are not

## Motivation and Context
So that now it can detect which messages are sent via wa web automatically and which are not

## How Has This Been Tested
I've tested it with my own Bot on windows platform and it works great!

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -130,6 +130,12 @@ class Message extends Base {
         this.fromMe = data.id.fromMe;
 
         /**
+         * Indicates if the message was sent via whatsapp web
+         * @type {boolean}
+         */
+        this.isWWebJS = data.id.id.startsWith('3EB0') && data.id.id.length === 12;
+
+        /**
          * Indicates if the message was sent as a reply to another message.
          * @type {boolean}
          */
--- a/index.d.ts
+++ b/index.d.ts
@@ -749,7 +749,7 @@ declare namespace WAWebJS {
         from: string,
         /** Indicates if the message was sent by the current user */
         fromMe: boolean,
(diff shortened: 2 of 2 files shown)


src/structures/Message.js
```diff
@@ -129,6 +129,12 @@ class Message extends Base {
          */
         this.fromMe = data.id.fromMe;
 
+        /**
+         * Indicates if the message was sent via whatsapp web
+         * @type {boolean}
+         */
+        this.isWWebJS = data.id.id.startsWith('3EB0') && data.id.id.length === 12;
```
alechkos, line 136: What about [`this.deviceType`](https://github.com/pedroslopez/whatsapp-web.js/blob/8f5d9ad6fce6a8e8746176892adc0b796d96e1d6/src/structures/Message.js#L93)?

RC047: `this.deviceType` is only to indicate the type of device that is connected to whatsapp web, this one's function is to distinguish which messages are sent automatically and which messages are sent manually in whatsapp web

tuyuribr:
Totally a bad Idea

renzhxr:
is this a solution so that bots respond to commands sent from wa web?

RC047:
> is this a solution so that bots respond to commands sent from wa web?

yes

aalwhaib:
On event "message_created" just check if `message.ack === 0 `
this mean the message send automatically

PR #2444, opened 2023-08-26
feat: new chat/group settings + methods + minor updates

### LINKED ISSUE #1559: set ephemeral protocol

_No response_

an option to set the ephemeral setting of the chat

_No response_

### Additional context

_No response_

alechkos:
ᅠ


tuyuribr:
Good work, just a note `group_join` triggering when someone need admin approve to join is a breaking change.
Someone might be using this to tag users in a welcome message and it will fail for this scenario.
Anyway, This scenario should trigger `group_join`, I don't think we need a new event just for approval, What do you think?

alechkos:
@tuyuribr 
Made `group_join` event docs clarifications

matheus-de-araujo (approved):
Very document code, congratulations

Halphas:
Did this also fix set group description and set group profile pic? If not, I can provide it if u like.

rizwan1239:
It stays stuck at "Authenticated Successfully" and no event seems to be fired. Any idea?

rizwan1239:
After installing this PR, it stays stuck at "Authenticated Successfully" and no event seems to be fired. Any idea?

purpshell (changes requested):
Mandatory block

PR #2446, opened 2023-08-26
add to MessageMedia class fromVars property

Itay-Sharoni:
# PR Details
MessageMedia class support only fromFile and fromUrl properties
i needed a way to pass the data as variables to MessageMedia from other application that run alongside
so i added fromVars property, it will accept mimetype, b64data, filename vars

example (not real):

```
// send/ptt
app.post('/send/ptt', async (req, res) => {
  const { phone, message } = req.body;
  media = await message.MessageMedia

  const ptt = MessageMedia.fromVars(media.mimetype,media.data,media.filename)
  const response = await client.sendMessage(phone, ptt, { sendAudioAsVoice: true });
```

updated:
index.d.ts
src/structures/MessageMedia.js

## How Has This Been Tested
I tetsted providing opus file as base 64 with file name as null
and with const response = await client.sendMessage(phone, ptt, { sendAudioAsVoice: true });
so it set the audio as PTT

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -975,6 +975,9 @@ declare namespace WAWebJS {
         /** Creates a MessageMedia instance from a local file path */
         static fromFilePath: (filePath: string) => MessageMedia
 
+        /** Creates a MessageMedia instance from variables */
+        static fromVars: (mimetype: string, b64data: string, filename: string) => MessageMedia
+
         /** Creates a MessageMedia instance from a URL */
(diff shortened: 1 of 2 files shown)


opssemnik:
you can already just call the constructor, no need for a method
var ptt = new MessageMedia(media.mimetype, media.data, media.filename)

shirser121:
> you can already just call the constructor directly, no need for a wrapper method var ptt = new MessageMedia(media.mimetype, media.data, media.filename)
Agreed, close the PR

Issue #2565, opened 2023-10-09
Potential Bug in fetchMessages Method


sujitmahapatra:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

**Issue Description:-**

In the fetchMessages method of the Chat class, there is a potential bug related to message filtering based on the fromMe option. This issue arises when the fromMe option is not provided or is undefined in the searchOptions parameter. The code currently compares m.id.fromMe with searchOptions.fromMe, which may lead to unexpected behavior when fromMe is not explicitly set.

### Expected behavior

The fetchMessages method should return all messages when the fromMe option in searchOptions is not provided or is undefined. It should not apply any filtering based on the fromMe option in this case.

### Steps to Reproduce the Bug or Issue

1. Call the fetchMessages method of the Chat class without providing the searchOptions parameter or with searchOptions.fromMe set to undefined.
2. Observe that the method does not return all messages and may exclude some messages that do not match the fromMe condition.

### Browser Type

Google Chrome

### WhatsApp Account Type

Standard

### Does your WhatsApp account have multidevice enabled?

Yes, I am using Multi Device

### Environment

Node.js version: [e.g., 14.17.5]
Library version: [e.g., WhatsApp Web.js 2.11.1]
Operating System: [e.g., Windows 10]
Other relevant information: [e.g., browser version if applicable]

alechkos:
You can not provide the `fromMe` property and it will return all fetched messages

PR #2566, opened 2023-10-09
Update Chat.js

sujitmahapatra:
# PR Details

**Bug fixed** - Potential Bug in fetchMessages Method #2565

## Description

In this PR, I've addressed a potential bug in the `fetchMessages` method of the Chat class. The method now correctly returns all messages when the `fromMe` option in `searchOptions` is not provided or is `undefined`. It filters messages based on the `fromMe` condition only when `searchOptions.fromMe` is explicitly set.

## Related Issue

N/A

## Motivation and Context

This change is required to ensure that the `fetchMessages` method works as expected and returns all messages when the `fromMe` option is not specified. Previously, it was filtering messages incorrectly when `fromMe` was not set, leading to unexpected behavior.

## How Has This Been Tested

I have thoroughly tested this change by performing the following steps:
- [ ] I've created test cases that cover different scenarios of the `fetchMessages` method, including cases with and without the `fromMe` option.
- [ ] I've tested the code in the environment where it will be used to ensure that it functions correctly.

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/structures/Chat.js
+++ b/src/structures/Chat.js
@@ -8,268 +8,146 @@ const Message = require('./Message');
  * @extends {Base}
  */
 class Chat extends Base {
-    constructor(client, data) {
-        super(client);
+  constructor(client, data) {
+    super(client);
 
-        if (data) this._patch(data);
-    }
+    if (data) this._patch(data);
+  }
 
-    _patch(data) {
-        /**
-         * ID that represents the chat
-         * @type {object}
-         */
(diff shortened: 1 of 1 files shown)


shirser121:
Please fix ESLint

alechkos:
If the `fromMe` property is not provided, the method will fetch all messages without filtering based on whether they are from the current user

Issue #26, opened 2019-12-04
QR RECEIVED null


dedenhendrap:
I have used example.js but the qr is null, anyone have seen like this ?

pedroslopez:
I'm now seeing this issue when running in headless mode. Will look into it.

dedenhendrap:
Hello Pedro, thank you for your response, can you help to resolve that bug ?

pedroslopez:
I found a workaround that involves making puppeteer take a screenshot, which for whatever reason sets the qr code data attribute properly. I’m still trying to figure out what exactly causes this but I’ll just push the workaround tomorrow if I don’t find it soon.

pedroslopez:
Just to confirm, you are running in headless mode, correct? @dedenhendrap

dedenhendrap:
Yes, i am running in headless mode @pedroslopez

pedroslopez:
@dedenhendrap Please test again. I've released version 0.3.1 which should fix this issue.

dedenhendrap:
yes of course, thank you for information

dedenhendrap:
it solved, thank you pedro 
![image](https://user-images.githubusercontent.com/31558217/70120512-5f239500-169f-11ea-83df-9a6145354367.png)

PR #2691, opened 2023-12-15
Reduce supply chain attacks vectors

tuyuribr:
# PR Details

For some "smaller" dependencies we should hard-code specific versions to make it harder to make a supply chain attack

## Types of changes

- [x] Dependency change

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   "dependencies": {
     "@pedroslopez/moduleraid": "^5.0.2",
     "fluent-ffmpeg": "2.1.2",
-    "jsqr": "1.3.1",
     "mime": "^3.0.0",
     "node-fetch": "^2.6.5",
     "node-webpmux": "3.1.0",
--- a/package.json
+++ b/package.json
@@ -30,11 +30,11 @@
   "homepage": "https://wwebjs.dev/",
   "dependencies": {
     "@pedroslopez/moduleraid": "^5.0.2",
-    "fluent-ffmpeg": "^2.1.2",
-    "jsqr": "^1.3.1",
+    "fluent-ffmpeg": "2.1.2",
+    "jsqr": "1.3.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,11 +30,11 @@
   "homepage": "https://wwebjs.dev/",
   "dependencies": {
     "@pedroslopez/moduleraid": "^5.0.2",
-    "fluent-ffmpeg": "^2.1.2",
-    "jsqr": "^1.3.1",
+    "fluent-ffmpeg": "2.1.2",
+    "jsqr": "1.3.1",
```
alechkos, line 34:
```suggestion
```
@tuyuribr 
`jsqr` already is not in use, should we get rid of it? 
Was added in [#26](https://github.com/pedroslopez/whatsapp-web.js/commit/bea1ebf48035f52b71b145214eac0dfd6576b113) and then its usage was removed in [#68](https://github.com/pedroslopez/whatsapp-web.js/commit/5c9e76e23f52418eaa3062e33194eb916c9565ef)

alechkos, line 34:
```suggestion
```
@tuyuribr 
`jsqr` already is not in use, should we get rid of it? 
Was added in [#26](https://github.com/pedroslopez/whatsapp-web.js/commit/bea1ebf48035f52b71b145214eac0dfd6576b113) and then its usage was removed in [#889](https://github.com/pedroslopez/whatsapp-web.js/commit/0d55d408850ef5214fde751d8b55af42fd70aef1#diff-b880f47f381d68dcd7473fb094acb874cd50b1140cdf2bfd7a368fa509738073)

tuyuribr: We should ahhaha

alechkos (approved):

shirser121 (approved):
LGTM, even we should update our dependencies

PR #2768, opened 2024-02-13
Update Client.js

abdelazizElemary:
This change to solve the target closed issue when using puppeteer as headless

# PR Details

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -754,7 +754,7 @@ class Client extends EventEmitter {
         } else {
             this.pupPage.on('response', async (res) => {
                 if(res.ok() && res.url() === WhatsWebURL) {
-                    await webCache.persist(await res.text());
+                    await webCache.persist(await res.text().catch(_=>_)).catch(_=>_);
                 }
             });
         }


justoverclockl:
how adding a catch will solve the issue?

abdelazizElemary:
> how adding a catch will solve the issue?

Without this catch it crashes when logging out if you are using puppeteer headless. Because there isn't page to listen on it's events. So the solution is simply the try and catch

Gdakgdak22 (approved):

AliTarek97 (approved):

src/Client.js
```diff
@@ -754,7 +754,7 @@ class Client extends EventEmitter {
         } else {
             this.pupPage.on('response', async (res) => {
                 if(res.ok() && res.url() === WhatsWebURL) {
-                    await webCache.persist(await res.text());
+                    await webCache.persist(await res.text().catch(_=>_)).catch(_=>_);
```
shirser121, line 757: If you using await you should use try catch

tuyuribr:
More than one year stale

PR #2788, opened 2024-02-25
Add members settings

GulgDev:
Added a function `setAddMembersAdminsOnly`.
The function sets who can add members to the group: either only admins or all the group members. If turned on, only the group admins can add others to the group.


alechkos (approved):
Good job, thank you
Tested on WWeb v2.2412.54

shirser121 (approved):
Thanks!

PR #2822, opened 2024-03-18
Update Injected.js remove .default

maunklana:
# PR Details

 a change in the function on waweb 2.3000

## Description

the function is not in default key, so I remove all .default in injected.js

## Related Issue

Error: Evaluation failed: TypeError: Cannot read properties of undefined (reading 'default')

## How Has This Been Tested

Tested Sending and Receiving messages, using example from documentation

Windows 11
Android 8.1.0
Puppeter use Brave Browser
Node.js v20.9.0
WAWEB VERSION 2.3000.1012115560

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected.js
+++ b/src/util/Injected.js
@@ -5,15 +5,15 @@ exports.ExposeStore = (moduleRaidStr) => {
     eval('var moduleRaid = ' + moduleRaidStr);
     // eslint-disable-next-line no-undef
     window.mR = moduleRaid();
-    window.Store = Object.assign({}, window.mR.findModule(m => m.default && m.default.Chat)[0].default);
+    window.Store = Object.assign({}, window.mR.findModule(m => m && m.Chat)[1]);
     window.Store.AppState = window.mR.findModule('Socket')[0].Socket;
     window.Store.Conn = window.mR.findModule('Conn')[0].Conn;
     window.Store.BlockContact = window.mR.findModule('blockContact')[0];
-    window.Store.Call = window.mR.findModule((module) => module.default && module.default.Call)[0].default.Call;
+    window.Store.Call = window.mR.findModule((module) => module && module.Call)[0].Call;
     window.Store.Cmd = window.mR.findModule('Cmd')[0].Cmd;
     window.Store.CryptoLib = window.mR.findModule('decryptE2EMedia')[0];
     window.Store.DownloadManager = window.mR.findModule('downloadManager')[0].downloadManager;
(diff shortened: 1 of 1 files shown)


src/util/Injected.js
```diff
@@ -5,15 +5,15 @@ exports.ExposeStore = (moduleRaidStr) => {
     eval('var moduleRaid = ' + moduleRaidStr);
     // eslint-disable-next-line no-undef
     window.mR = moduleRaid();
-    window.Store = Object.assign({}, window.mR.findModule(m => m.default && m.default.Chat)[0].default);
+    window.Store = Object.assign({}, window.mR.findModule(m => m && m.Chat)[1]);
```
allburov, line 8: Just curious - shouldn't it be 0 as in other cases? 
```suggestion
    window.Store = Object.assign({}, window.mR.findModule(m => m && m.Chat)[0]);
```

maunklana: 0 only got the function Chat
1 gets the function as the previous default.

313551828-a6eb6fe6-7895-45ee-b502-ff8fbb7d6e52

voidpack:
thx thats work. tested it on whatsapp version 2.3000.1012116664

src/util/Injected.js
allburov: TY!

MoudiZd:
Hello,
But does these changes break old web whatsapp versions, because in this case we must think about a new way to support both new and old versions, at least till old versions are not valid by whatsapp itself anymore

maunklana:
yes, this will break the old Waweb version.
we will need a work around to support it.
> Hello, But does these changes break old web whatsapp versions, because in this case we must think about a new way to support both new and old versions, at least till old versions are not valid by whatsapp itself anymore

src/util/Injected.js
maunklana: i have new [PR](https://github.com/pedroslopez/whatsapp-web.js/pull/2825)
> i think the reason of that is that we have many modules that have Chat function, so we need to check about something else than Chat method to define that this module is the Store, maybe if we check also for AdCollection or Blocklist or any other property we see in the picture, this will do the trick, like this:
> 
> ```
> window.Store = Object.assign({}, window.mR.findModule(m => m && m.Chat && m.Blocklist)[0]);
> ```
> 
> But it needs more test to ensure that Blocklist always found within Store module along the Chat function
> 
> Hope someone can test this and tell us if it does what we need, as currently i am not able to test it from my side !!!

maunklana:
got to [New PR](https://github.com/pedroslopez/whatsapp-web.js/pull/2825)

PR #2942, opened 2024-04-10
Make an NPM release via GitHub

purpshell:
# PR Details

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/tools/publish
+++ b/tools/publish
@@ -68,8 +68,13 @@ echo ""
 
 npm version $VERSION_ARGS || exit 1
 
-npm run generate-docs
 git push && git push --tags || exit 1
+
+npm run generate-docs
+
+git add .
+
+git commit "chore(${NEW_VERSION}): generate docs"
 
 NEW_VERSION=`cat package.json | jq -r .version`
 echo "New Version: $NEW_VERSION"
--- a/tools/publish
+++ b/tools/publish
@@ -74,7 +74,9 @@ npm run generate-docs
 
 git add .
 
-git commit "chore(${NEW_VERSION}): generate docs"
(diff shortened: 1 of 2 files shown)


pedroslopez (approved):
Thanks!

From what I remember, I believe the GitHub release should respect the `prerelease` flag marked in the output here to make the release correctly

pedroslopez (changes requested):
Ah! I forgot something - another thing that happens as part of the mainline release process is docs generation.

There's a script that does this, `npm run generate-docs`

This should happen only for main releases, after the version bump (so the docs contain the right version) but before the version is tagged

tgmarinho (approved):

tools/publish
```diff
@@ -57,6 +67,8 @@ echo "-----> BUMP VERSION"
 echo ""
 
 npm version $VERSION_ARGS || exit 1
+
+npm run generate-docs
```
pedroslopez, line 71: this still won't be included as part of the commit. 

`npm version` creates a commit and tag:

> If run in a git repo, it will also create a version commit and tag. This behavior is controlled by git-tag-version (see below), and can be disabled on the command line by running npm --no-git-tag-version version. It will fail if the working directory is not clean, unless the -f or --force flag is set.

So the docs generation is outside of the bump commit. 

To keep this simple, we might be able to just do the docs generation after the `git push`, and commit/push that separately. It won't be included as part of the tag, but at least we can move forward with this PR and unlock the functionality.

PurpShell: Seems good

pedroslopez (approved):

codebossdev (approved):

PurpShell:
pedro approval > contributor approval, merging

PR #3001, opened 2024-05-02
Fix window.Store.ProfilePic.profilePicFind is not a function error

### LINKED ISSUE #3005: Unable to get conact avatar.

### Describe the bug

When trying to call `client.getProfilePicUrl` method `Error: Evaluation failed: TypeError: window.Store.ProfilePic.profilePicFind is not a function` will be thrown.

### Expected behavior

Get the url of the avatar of the contact.

1. login.
2. get the avatar of yourself

### Relevant Code

Client.js
```ts
async getProfilePicUrl(contactId) {
        const profilePic = await this.pupPage.evaluate(async contactId => {
            try {
                const chatWid = window.Store.WidFactory.createWid(contactId);
                return await window.Store.ProfilePic.profilePicFind(chatWid);
            } catch (err) {
                if(err.name === 'ServerStatusCodeError') return undefined;
                throw err;
            }
        }, contactId);
        
        return profilePic ? profilePic.eurl : undefined;
    }
```

### Browser Type

Chromium

### WhatsApp Account Type

Standard

Yes, I am using Multi Device

### Environment

OS: Mac
Node: 18
Whatsapp-web.js:  1.23.1-alpha.6
Whatsapp Web: 2.3000.1013249969

### Additional context

_No response_

seowzhenjun0126:
# PR Details

Replace `window.Store.ProfilePic.profilePicFind` with `window.Store.ProfilePic.requestProfilePicFromServer`

## Description

The function

`window.Store.ProfilePic.profilePicFind`

is no longer available in v2.3x, and is replaced with

`window.Store.ProfilePic.requestProfilePicFromServer`

image

## Related Issue

closes #3005

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1209,7 +1209,7 @@ class Client extends EventEmitter {
         const profilePic = await this.pupPage.evaluate(async contactId => {
             try {
                 const chatWid = window.Store.WidFactory.createWid(contactId);
-                return await window.Store.ProfilePic.profilePicFind(chatWid);
+                return await window.Store.ProfilePic.requestProfilePicFromServer(chatWid);
             } catch (err) {
                 if(err.name === 'ServerStatusCodeError') return undefined;
                 throw err;
--- a/src/Client.js
+++ b/src/Client.js
@@ -1209,7 +1209,7 @@ class Client extends EventEmitter {
         const profilePic = await this.pupPage.evaluate(async contactId => {
             try {
                 const chatWid = window.Store.WidFactory.createWid(contactId);
-                return await window.Store.ProfilePic.requestProfilePicFromServer(chatWid);
+                return await (window.WWebJS.compareWwebVersions(window.Debug.VERSION, '<', '2.3000.0') ? window.Store.ProfilePic.profilePicFind(chatWid) : window.Store.ProfilePic.requestProfilePicFromServer(chatWid));
             } catch (err) {
                 if(err.name === 'ServerStatusCodeError') return undefined;
                 throw err;
--- a/src/Client.js
+++ b/src/Client.js
@@ -1209,7 +1209,9 @@ class Client extends EventEmitter {
         const profilePic = await this.pupPage.evaluate(async contactId => {
             try {
                 const chatWid = window.Store.WidFactory.createWid(contactId);
-                return await (window.WWebJS.compareWwebVersions(window.Debug.VERSION, '<', '2.3000.0') ? window.Store.ProfilePic.profilePicFind(chatWid) : window.Store.ProfilePic.requestProfilePicFromServer(chatWid));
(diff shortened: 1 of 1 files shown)


StiflerPasha:
and how about previous wa-versions?

alechkos:
@StiflerPasha 
> and how about previous wa-versions?

In 2.2412.54, `window.Store.ProfilePic.profilePicFind` is defined

StiflerPasha:
We need to make an implementation that is supported on both versions

alechkos (changes requested):
@seowzhenjun0126 

Add support for previous WWeb versions,
you can use `window.WWebJS.compareWwebVersions` that is located in
`src/util/Injected/Utils.js`

guigo613 (changes requested):
In this case, it will break for older versions. I believe the best solution would be something like:

```javascript
return await (window.Store.ProfilePic.profilePicFind || window.Store.ProfilePic.requestProfilePicFromServer)(chatWid);
```

N0N4M3BNS (approved):

mauriciopfl:
> In this case, it will break for older versions. I believe the best solution would be something like:
> 
> ```js
> return await (window.Store.ProfilePic.profilePicFind || window.Store.ProfilePic.requestProfilePicFromServer)(chatWid);
> ```

Don`t work
Error on: window.WWebJS.compareWwebVersions

src/Client.js
```diff
@@ -1209,7 +1209,7 @@ class Client extends EventEmitter {
         const profilePic = await this.pupPage.evaluate(async contactId => {
             try {
                 const chatWid = window.Store.WidFactory.createWid(contactId);
-                return await window.Store.ProfilePic.profilePicFind(chatWid);
+                return await (window.WWebJS.compareWwebVersions(window.Debug.VERSION, '<', '2.3000.0') ? window.Store.ProfilePic.profilePicFind(chatWid) : window.Store.ProfilePic.requestProfilePicFromServer(chatWid));
```
alechkos, line 1212:
```suggestion
                return window.WWebJS.compareWwebVersions(window.Debug.VERSION, '<', '2.3000.0')
                    ? await window.Store.ProfilePic.profilePicFind(chatWid)
                    : await window.Store.ProfilePic.requestProfilePicFromServer(chatWid);
```

Lyfhael:
Hey,
Should it be modified from this file too maybe ? https://github.com/pedroslopez/whatsapp-web.js/blob/41b280c769918db03ac6ec12cf0e4215e7b864ef/docs/Client.js.html#L1227

erm3nda:
This has stopped to work for me few days ago, I don't know the exact moment. 
requestProfilePicFromServer function still exists but throws internal error evaluating e.isNewsletter and breaks.

PR #3004, opened 2024-05-02
feat: get pinned messages

alechkos:
# PR Details

### Methods return an array of `Message` instances, or an empty array if there are no pinned messages in a chat/group.

---

## Usage Example <a href="https://github.com/pedroslopez/whatsapp-web.js/pull/3004#usage-example" id="usage-example"/>

```js
// client initialization...

client.on('ready', async () => {
    const chatId = 'your_chat_id@c.us';

    let pinnedMsg = await client.getPinnedMessages(chatId);
    console.log(pinnedMsg);

    // OR
    const chat = await client.getChatById(chatId);
    pinnedMsg = await chat.getPinnedMessages();
    console.log(pinnedMsg);
});
```

---

## To test this PR by yourself you can run one of the following commands:

- **NPM**
```powershell
npm i github:alechkos/whatsapp-web.js#get-pinned-msg
```
- **YARN**
```powershell
yarn add github:alechkos/whatsapp-web.js#get-pinned-msg
```

---

## How Has The PR Been Tested

Tested with a code provided in [usage example](https://github.com/pedroslopez/whatsapp-web.js/pull/3004#usage-example).

---

## Types of Changes
- [ ] Dependency change
- [ ] Breaking change (fix/feature that would cause existing functionality to change)

## Checklist
- [X] My code follows the code style of this project
- [X] I have updated the documentation accordingly (index.d.ts)


src/Client.js
```diff
@@ -1044,6 +1044,32 @@ class Client extends EventEmitter {
...
+     * Gets an instance of a pinned message in a chat
+     * @param {string} chatId The chat ID
+     * @returns {Promise<Message>} Returns an instance of a pinned message in a chat
+     */
+    async getPinnedMessage(chatId) {
+        const pinnedMsg = await this.pupPage.evaluate(async (chatId) => {
+            const chatWid = window.Store.WidFactory.createWid(chatId);
+            const chat = await window.Store.Chat.find(chatWid);
+            if (!chat) return null;
+            await window.Store.Cmd.openChatAt(chat);
```
shirser121, line 1057: Why we need to open the chat?
We have to find another solution 

We need to find a way to update the collection of pinned messages in another way

alechkos: @shirser121
> Why we need to open the chat?
 
The collection is populated only when you click on a pinned message banner inside the chat itself. Otherwise, the collection will remain empty.
So this is the only way to do so.

Open the chat in which you are looking for a pinned message → wait 500ms for the collection to be populated → get the pinned message

shirser121: We just need to find the inner function that did it, and call it directly

alechkos: @shirser121
> We just need to find the inner function that did it, and call it directly 

Unfortunately we don't have that function that would do the job.
I reversed the way it gets the pinned message and it happens after you enter the chat (here the collection is populated), and after you click on the banner it gets the message id from the collection. There is no function in the middle (at least I didn'tfind it)

alechkos: ### This is a call stack:


### This is a `j` function (right after the `onClick`):


### It calls for an `M` function which appears as `I` in a call stack:


### And here it is an `I` function that opens a chat at a message:

alechkos: @shirser121 
Maybe we have to look at a logic that happens immediately after you open a chat, when the collection is populated

joweste:
Hi, is there an event to know when a chat or message was pinned?
I´d like to update my tables database on this event.

shirser121 (approved):

shirser121 (approved):
LGTM

tuyuribr (approved):

PR #2816, opened 2024-03-14
Farewell, Webpack! Re-injection, Pairing Code login, and 2.24x jumbofix

purpshell:
[Support my work on GitHub Sponsors](https://github.com/sponsors/PurpShell)

# TL;DR: 

WhatsApp released 2.3000x, a version that does not use Webpack anymore. We had to change the Store system to no longer use `moduleRaid`. Instability issues are caused by relying on selectors & DOM Observers for login, as well as increased performance issues. This prompted me to adopt the "New startup flow" that I mentioned [here](https://discord.com/channels/698610475432411196/698612477226909736/1191150981501685810). This update is important. This update will boost reliability in WWebJS at least 400%. I also added pairing code login for those who needed it since I had the docs on hand (I reverse engineer then take notes, so I know what I know when I make them into PRs). It changes a lot in the authentication and launching. I advise you all to test your systems before deploying to production (even though this build should be prod-safe).

You should immediately update to my version and cease using any other fork or version. Your system will work regardless if you are using a web cache, local cache, no cache.. LocalAuth, RemoteAuth, NoAuth... 2.24 and 2.3000. 

TO INSTALL:
```
npm install github:pedroslopez/whatsapp-web.js#webpack-exodus
```
If you encounter a `git not found` error in the NPM/yarn install: Download git, https://git-scm.com

Thank you for your trust and time and thank you to my [sponsors](https://github.com/sponsors/purpshell) for their contribution.

# PR Details

This PR adds support for Wwebjs past version 2.3000.x and 2.24..x. Send "F" in the chat for Webpack!

As always, very big thanks to my sponsors for their support! I am only able to do this thanks to them!

Also thanks to @allburov for his work on #2827, helped me a bit to know which modules are problematic. 

This PR also adds re-injection in the case of navigation or refresh or intentional logout and pairing code login option.

It also changes a lot of the code in the startup flow.

To note:

1.  AUTHENTICATED and READY will now depend on the Socket, so if the WWebJS restarts or loses Internet, it will trigger these events. Do not trust that the event will only be triggered once.
2. `await client.initialize()` returns it promise much earlier at the point of a new QR code (no login). The behavior is almost the same when logged in (it resolves the promise at AROUND ready). 
3. The Loading screen text will be hard-coded to "WhatsApp" for now. A new update will deprecate this screen text and replace it with loading types (`ORGANIZING`, `INTIAL_LOAD`, `DOWNLOADING`,...)
[...]

PR #3074, opened 2024-06-03
fix: regex to extract version (LocalAuth)

pieshna:
# PR Details

- Actual localAuth contains error, regex return null and this may crash app.

## Description

Actually i'm using node 18 and typescript.
When execute a simple inicialize script provided on officially docs web this crashed.

Script: 
image

Error:
image


alechkos (commented):
#2816

PR #3214, opened 2024-07-20
Fix version extraction in LocalWebCache.js to handle null match

rniedson:
This pull request addresses an issue where the version extraction in LocalWebCache.js can throw an error if the regex match returns null. This fix ensures that the code safely handles null matches, preventing a TypeError.

# PR Details

## Description

This change updates the version extraction logic in the `persist` method to safely handle cases where the regex match returns null. This prevents the TypeError caused by trying to access properties of null.

## Motivation and Context

This change is required to make the version extraction process more robust and to prevent crashes when the manifest version is not found in the index.html.

## How Has This Been Tested

This change has been tested by running the modified code and ensuring that it handles cases where the version is not found in the index.html without throwing errors.

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/webCache/LocalWebCache.js
+++ b/src/webCache/LocalWebCache.js
@@ -6,7 +6,7 @@ const { WebCache, VersionResolveError } = require('./WebCache');
 /**
  * LocalWebCache - Fetches a WhatsApp Web version from a local file store
  * @param {object} options - options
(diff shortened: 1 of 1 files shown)


alechkos:
#2816

PR #3328, opened 2024-10-07
Clarify client.sendMessage chatId parameter

valoricDe:
# PR Details

Clarification of docu

## Description

clarify what can be passed as chatId


alechkos:
Clarified enough in `example.js`

Issue #3372, opened 2024-11-09
Add event emitter for chat state changes, ex: user started or stopped to type or record


lucas-almeida026:
### Is the feature you are looking for already implemented in the library?

- [X] I have reviewed the implemented features.

### Is your "solution" already implemented in an opened pull request?

- [X] I have reviewed the existing pull requests.

### Describe the solution you'd like.

I’d like to be able to add an event handler for the "chat state change" event. As far as I know, the library does not currently handle this event.

The API I have in mind would look something like this:
```js
client.on(Events.CHAT_STATE_CHANGED, ({wid, stateValue}) => {
    console.log(`User with wid = ${wid._serialized} is now: ${stateValue}`)
})
```

An example use case for this feature would be interrupting a chatbot from sending automatic messages when the user is typing, thereby improving the user experience.

### Describe an alternate solution if you have one.

I managed to get something working, but I'm having a hard time integrating it with the rest of the code. I'd like to explain the path I took in hopes someone with more experience with the code base can use it or suggest how I could continue the feature.

The approach is simple, on the `client.inject` method, after everything the code "waits" for `window.Store.Presence` to become available and then overwrite `window.Store.Presence.modelClass` with a function that wraps the original function and adds accessor properties for `__x_type` which holds the current chat state, therefore every time the property is changed we can call a callback function.

I did everything on the window object because I couldn't make it work in a "nicer" way.

I chose to modify the window object because I couldn't find a cleaner solution that worked. Here’s the function that adds the accessor properties to instances of `Presence.modelClass`. I attempted to emit the event here, but it didn’t work as expected, and I couldn’t find a way to make it function properly:
```js
window.__setXType = (instance) => {
	if (instance?.__x_chatstate) {
		let type = instance.__x_chatstate?.__x_type;

		Object.defineProperty(instance.__x_chatstate, '__x_type', {
			get() {
				return type;
			},
			set(next) {
				console.log('chatstate changed', instance.id, next);
				// should emit an event here
				type = next;
			},
			configurable: true,
			enumerable: true
		})
	}
	return instance
}
```

This function alters every existing instance and every instance created afterward. The logic for this is handled by the `injectPresenceModel` function:

```js
window.injectPresenceModel = () => {
	const OriginalModelClass = window.Store.Presence.modelClass;
	const originalModelPrototype = window.Store.Presence.modelClass.prototype;
	window.Store.Presence.modelClass = function (...args) {
[...]

Issue #2636, opened 2023-11-18
Chat status method (online, typing, offline)


aronilie:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the solution you'd like.

I would like to have the possibility to know a chat status. If it is online, typing or disconnected.

yaneony:
I think he mean to get presence status, not to set one.

aronilie:
> I think he mean to get presence status, not to set one.

Exactly

nachobascar:
I have the same doubt, knowing if the person is typing would help my bot to wait until he stops typing before sending a message

Solo-it:
> If you want to change status such as online, offline, and typing, you can use a function method like this:
> 
> ```js
> client.sendPresenceAvailable() // online
> client.sendPresenceUnvailable() // offline
> chat.sendStateTyping() // typing
> chat.sendStateRecording() // recording
> ```

Not work (

alechkos:
Duplicate of #82

Issue #1137, opened 2022-01-26
Listener for "typing" and "recording"


tonvital:
Is it possible to add a listener to show the status of "typing" and
 "recording", which returns the contact or contact ID?

Thanks in advance.

![Sem título](https://user-images.githubusercontent.com/50684920/151189781-4cd6b5c5-e7ca-45df-b031-e818a168aa6c.png)

For example:
```
contact {
   status: "typing" | "recording" | "default"
}

client.on("contact-status", (stt) => console.log(stt))
```

PurpShell:
Duplicate of #82

mrGLEIS:
Any updates for this feature? Please add it.

PR #3374, opened 2024-11-13
Emit chat state changed event when chat state changes (e.g. one contact starts typing)

### LINKED ISSUE #82: Check user online state. [feature request]

lucas-almeida026:
# PR Details

The PR adds the functionality requested on #82 and on related issues: #1137, #2636, #3372.
This feature introduces the ability to detect chat states such as 'available', 'unavailable', 'typing' and 'recording', addressing a longstanding request from the community.

closes #82

## Description

3 main changes were made:
1. addition of the `CHAT_STATE_CHANGED` event on the Events object from `src/utiol/Contants.js`
2. addition of an exposed function to emit this particular event through the client object `src/Client.js:249`
3. addition of 2 functions at the window level
	- `injectPresenceModelInstance`: this function wraps the underlying modelClass from the Store.Presence object. the modelClass is used to generate all the models that represents a presence for a given chat.
	  This is done to "inject" the get and set properties on every single instance generate my the model class to be able to call the emit function every time the model chat state property is update 
	- `execPresenceModelInjection`: this function is called when the Store.Presence object becomes available (it takes a couple of milliseconds for it to load) and calls the `injectPresenceModelInstance` to inject every model already created.
	  The function also makes sure to maintain the prototype chain of every injected object to avoid breaking underlying functionality

## Motivation and Context

The main objective is to have one more way to improve user experience of apps built using whatapp-web.js. By providing a way to handle events related to chat state changes developers will be able to create interaction flows with a finer level of control.

An example use case would be: prevent an automatic response from being sent if the target user is typing, so it doesn't interrupt the user.

## To apply the PR run one of the following commands:
- **NPM**
```powershell
npm i github:lucas-almeida026/whatsapp-web.js#main
```
- **YARN**
```powershell
yarn add github:lucas-almeida026/whatsapp-web.js#main
```

## How Has This Been Tested

I tested manually using the `shell` command from the `package.json` file.

Actions:
- on the terminal: `npm run shell`
- after the shell loads: `client.on('chat_state_changed', console.log)`
- enter on my personal whatsapp account and open the chat with the connected account
- start typing
- observe the shell printing: `{ chatId: '*********@c.us', chatState: 'typing' }`
- stop typing on my phone
- observe the shell printing: `{ chatId: '*********@c.us', chatState: 'unavailable' }`

### Environment
- Machine OS: Linux - Ubuntu 24.04
- Phone OS: Android 11
- Library Version: 1.26.0
- WhatsApp Web Version: 2.3000.1018163899
- Puppeteer Version: ^18.2.1
- Browser Type and Version: Chromium 107.0.5296.0 - developer build 64-bit
- Node Version: 21.7.3
- NPM Version: 10.5.0

## Types of changes

- [ ] Dependency change

## Checklist

- [ ] ~~I have updated the documentation accordingly (index.d.ts).~~


pedroapolloniufscar:
nice contribution

tuyuribr:
The biggest problem with this function was beign reliable enough for production. Have you tested with changes on more than 10 chats at one time ? Does it auto subscribe for the chat "typping" event ?

lucas-almeida026:
@tuyuribr I did not test for multiple chats at the same time. I also only tested using `headless: false` option, maybe if headless is set to true it can impact the reliability (just speculating).

it does not auto subscribe for the "typing" event, instead it "subscribes" to changes on the `__x_chatstate` object, every time the property `__x_type` is changed it emits an event with the new state value

TheStarkster:
@lucas-almeida026 
Oh my god man, Nice contribution kuddos to you!

one question though, Does it also work when user who is connected starts typing?

NewtonMiranda00:
Wouldn't it make sense to separate the statuses into two categories: 'Action' (Typing, Recording, Standby) and 'Status' (Available, Unavailable)?

Right now, we mix these states, which creates confusion—especially when trying to detect if the user is offline.

For example, during some tests with this PR, when Puppeteer navigates to another chat, it triggers an 'Unavailable' status even though the user is actually in 'Typing'. This leads to misleading behavior and makes it harder to handle presence accurately.

lucas-almeida026:
> @lucas-almeida026
> 
> Wouldn't it make sense to separate the statuses into two categories: 'Action' (Typing, Recording, Standby) and 'Status' (Available, Unavailable)?

I totally agree with you, it makes sense to separate these two _kinds_ of statuses.
To be 100% honest though, I initially did this modification as a "quick hack" for a proof of concept related to a prior project I was working on. Now this is not relevant to me anymore, so I don't plan on continue to work on this _feature_.

If anyone is interested in continue the work there is a couple of points I think is useful to share:
- I did not test, but I'm pretty sure the status change is only captured **if** the chat is rendered on the side panel (where all chats are shown). While coding, I notice that not all chats are "know", it appears to have some sort of sliding window that determines which chats become actual DOM elements, and I suspect that if a chat that is not rendered changes its state, the event will not be triggered 
- I also did not test the feature while running with the headless option set to true, maybe it doesn't work in headless mode.

NewtonMiranda00:
> I totally agree with you, it makes sense to separate these two _kinds_ of statuses. To be 100% honest though, I initially did this modification as a "quick hack" for a proof of concept related to a prior project I was working on. Now this is not relevant to me anymore, so I don't plan on continue to work on this _feature_.
> 
> If anyone is interested in continue the work there is a couple of points I think is useful to share:
> 
> * I did not test, but I'm pretty sure the status change is only captured **if** the chat is rendered on the side panel (where all chats are shown). While coding, I notice that not all chats are "know", it appears to have some sort of sliding window that determines which chats become actual DOM elements, and I suspect that if a chat that is not rendered changes its state, the event will not be triggered
> * I also did not test the feature while running with the headless option set to true, maybe it doesn't work in headless mode.

I understand. I'm personally interested in this functionality because I'm working with WhatsApp integrated with AIs. My project focuses on humanizing the AI, and often the person chatting doesn't even realize they're not talking to a human due to several subtle details.

I noticed and tested the issues you mentioned about it not working.

Part of the rendering seems to depend on the user's activity. I might be biased, but I believe it's just a WhatsApp resource-saving mechanism to avoid constantly showing "recording..." or "typing..."

What I noticed is that recently clicked or opened chats (those that have been selected and opened) have this status change feature active.

I looked through part of the repository, and I'm really trying to overcome the laziness to push a modification using a getter that wraps the underlying class function to retrieve:

```js
get isTyping() {
  // Logic...
}

get isRecording() {
  // Logic
}
```

All of this inside the PrivateChat class.

If I can finish up my current SaaS project, I’ll aim to build and submit a PR for this part by next weekend

ReniDelonzek:
Hello everyone, any updates here?

pedrograaff:
Hey guys, any updates? Super useful  contribution

PR #3351, opened 2024-10-25
Add 4 retries for fs.promises.rm

devlikepro:
# PR Details

## Description

Time to time it's possible to get some errors when logout - probably due to chrome process locking some files.

`fs.promises.rm` has `maxRetries` option for that https://nodejs.org/api/fs.html#fspromisesrmpath-options
We can set it to `4`, giving us 4 retries with about 1s max delay (100+200+300+400=1,000ms)

## Types of changes

- [ ] Dependency change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project.
- [x] I have updated the documentation accordingly (index.d.ts).
- [ ] I have updated the usage example accordingly (example.js)

PR #3410, opened 2024-12-08
Update LocalAuth.js

irvanfauzie:
# PR Details

There is an error when the user unlinks a device, because it cannot delete the `chrome_debug.log` file which is busy/locked.

## Description

Prevent error when deleting busy files:
Error: EBUSY: resource busy or locked, unlink '...\.wwebjs_auth\session\Default\chrome_debug.log'


alechkos:
#3351

PR #3455, opened 2025-02-02
adding chats pagination logic, and getChatsLength, and rename getChats that's get all chats to getAll chats

000MaDz000:
# PR Details

### Environment

- Machine OS: 
- Phone OS: 
- Library Version: 
- WhatsApp Web Version: 
- Puppeteer Version:
- Browser Type and Version: 
- Node Version: 

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "whatsapp-web.js",
-  "version": "1.26.1-alpha.3",
+  "version": "1.26.2-alpha.3",
   "description": "Library for interacting with the WhatsApp Web API ",
   "main": "./index.js",
   "typings": "./index.d.ts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "whatsapp-web.js",
-  "version": "1.26.1-alpha.3",
+  "version": "1.26.2-alpha.3",
```
alechkos, line 3: No need to change the lib version here, it is being changed automatically after each release, please revert this.

alechkos (changes requested):
It will be better to not break an exciting `getChats`, what about to rename your `getChats` to smth like `getChatsByOffset`.

Also please revert all format changes.

tuyuribr (changes requested):
Please, only change the code you are suposed to fix

PR #3555, opened 2025-05-14
Fix the Location declaration

wavy-cat:
# PR Details

Fix the Location declaration in `index.d.ts` and `Location.js`

## Description

In the declaration of the `Locations` class, the `options?: LocationSendOptions` attribute was replaced with four new attributes: `name?: string`, `address?: string`, `url?: string`, and `description?: string`, in accordance with the implementation of this class.

## Motivation and Context

The actual attributes of the class do not match those declared in `index.d.ts`, which leads to unexpected behavior when trying to access the `options` attribute. To work around this issue, it becomes necessary to use `// @ts-ignore`.

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1026,7 +1026,10 @@ declare namespace WAWebJS {
     export class Location {
         latitude: string;
         longitude: string;
-        options?: LocationSendOptions;
+        name?: string;
+        address?: string;
+        url?: string;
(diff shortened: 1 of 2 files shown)


shirser121 (approved):
LGTM

alechkos (approved):

PR #3567, opened 2025-05-27
Fix self message delete

aleborile:
# PR Details

Self Message delete fixed

## Description

The self message sent in own chat cannot be deleted due to error

## How Has This Been Tested

tested in shell:
$ const message = await client.sendMessage('MYSELFNUMBER@c.us', 'todelete')
$ await message.delete()

### Environment

Machine OS: Docker + Amazon Linux 2023
Phone OS: Android
Library Version: 1.27.0
WhatsApp Web Version: 2.3000.1022278655
Puppeteer Version: 18.2.1
Browser Type and Version: Chromium
Node Version: ^20

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -503,8 +503,11 @@ class Message extends Base {
         await this.client.pupPage.evaluate(async (msgId, everyone) => {
             const msg = window.Store.Msg.get(msgId) || (await window.Store.Msg.getMessagesById([msgId]))?.messages?.[0];
             let chat = await window.Store.Chat.find(msg.id.remote);
-            
             const canRevoke = window.Store.MsgActionChecks.canSenderRevokeMsg(msg) || window.Store.MsgActionChecks.canAdminRevokeMsg(msg);
+            const isSelfMsg = msg.from.user === msg.id.remote.user && msg.id.fromMe;
+            if(isSelfMsg) {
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -503,8 +503,11 @@ class Message extends Base {
         await this.client.pupPage.evaluate(async (msgId, everyone) => {
             const msg = window.Store.Msg.get(msgId) || (await window.Store.Msg.getMessagesById([msgId]))?.messages?.[0];
             let chat = await window.Store.Chat.find(msg.id.remote);
-            
             const canRevoke = window.Store.MsgActionChecks.canSenderRevokeMsg(msg) || window.Store.MsgActionChecks.canAdminRevokeMsg(msg);
+            const isSelfMsg = msg.from.user === msg.id.remote.user && msg.id.fromMe;
```
tuyuribr, line 507: Why not just msg.id.fromMe ?

aleborile: Because msg.id.fromMe is true even if the chat is not of the current user, but a general message sent by current user.
First check detect the current user personal chat.

aleborile:
There still be a weird behaviour: message is deleted on any wweb session but not on phone (master) session. Maybe someone who knows wweb core better can help on this.

purpshell (approved):
seems like a good way to get rid of the error for now

BenyFilho:
That issue was solved in recent changes

Issue #3609, opened 2025-06-26
client.sendMessage(chatId, message)  no response return


phpwebdeveloper900:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this a problem caused by your code, or is it specifically because of the library?

- [x] I have double-checked my code carefully.

### Describe the bug.

client.sendMessage(chatId, message) this is my function working fine before 

today after 4PM client.sendMessage(chatId, message)  no response return

so is there any change on lib OR anything else ?

Thanks

### Expected Behavior

client.sendMessage(chatId, message) this is my function working fine before 

today after 4PM client.sendMessage(chatId, message)  no response return

so is there any change on lib OR anything else ?

Thanks

### Steps to Reproduce the Bug or Issue

client.sendMessage(chatId, message) this is my function working fine before 

today after 4PM client.sendMessage(chatId, message)  no response return

so is there any change on lib OR anything else ?

Thanks

### WhatsApp Account Type

Standard

### Browser Type

Chromiun

### Operation System Type

Linux

### Phone OS Type

Android

### WhatsApp-Web.js Version

1.27

### WhatsApp Web Version

2.3000.1024218978

### Node.js Version

18.17.1

### Authentication Strategy

LocalAuth

typedbywill:
Yeah, same here

Eonus:
Same here

typedbywill:
I fixed it, one minute i will post the solution

typedbywill:
at /whatsapp-web.js/src/util/injected/Utils.js

Line 259 you change the 2 lines:

FROM:
await window.Store.SendMessage.addAndSendMsgToChat(chat, message);
return window.Store.Msg.get(newMsgId._serialized);

TO:

        await window.Store.SendMessage.addAndSendMsgToChat(chat, message);

        const waitForMsg = async (msgId, attempts = 10, delay = 100) => {
            for (let i = 0; i < attempts; i++) {
                const msg = window.Store.Msg.get(msgId);
                if (msg) return msg;
                await new Promise(res => setTimeout(res, delay));
            }
            return null;
        };

        return await waitForMsg(newMsgId._serialized);

phpwebdeveloper900:
> at /whatsapp-web.js/src/util/injected/Utils.js
> 
> Line 259 you change the 2 lines:
> 
[...]

PR #3610, opened 2025-06-26
Fixed client.sendMessage(chatId, message) no response return #3609

typedbywill:
# PR Details

## Description

This PR introduces a fix to the `sendMessage` method to address a critical issue where the function may return `undefined` immediately after attempting to send a message. This often causes errors when calling `getMessageModel()` due to trying to access `.serialize` on an undefined object.

The issue occurs because `Store.Msg.get(msgId)` is called immediately after sending the message via `addAndSendMsgToChat()`, without waiting for the message to be properly registered in the message store.

The fix implements a lightweight polling mechanism (`waitForMsg`) that retries fetching the message from the store until it becomes available, or until a short timeout is reached.

## Motivation and Context

This change is required to prevent runtime errors when sending messages, particularly in high-speed or concurrent environments where the message object may not be immediately available in the internal store after sending.

It also improves stability and reliability when using `sendMessage` in automated environments or integrations.

## How Has This Been Tested

- Patched and tested in a production-level implementation that interacts with `sendMessage`.
- Verified message availability using WhatsApp Web store objects directly via Puppeteer.
- Confirmed error no longer occurs after applying the retry mechanism.

### Environment

- Machine OS: Windows 11
- Phone OS: Android 13
- Library Version: whatsapp-web.js v1.23.0 (patched)
- WhatsApp Web Version: 2.2412.54
- Puppeteer Version: 21.3.8
- Browser Type and Version: Chromium 124
- Node Version: 20.x

## Types of changes

## Checklist

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -28,7 +28,7 @@ exports.LoadUtils = () => {
 
         let mediaOptions = {};
         if (options.media) {
-            mediaOptions =  options.sendMediaAsSticker && !isChannel
+            mediaOptions = options.sendMediaAsSticker && !isChannel
                 ? await window.WWebJS.processStickerData(options.media)
                 : await window.WWebJS.processMediaData(options.media, {
                     forceSticker: options.sendMediaAsSticker,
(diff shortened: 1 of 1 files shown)


alechkos:
Fixed in #3612

PR #3635, opened 2025-07-06
fix: prevent destroy() crash when instance not fully initialized

suryavmds:
# PR Details

When calling ```destroy()``` before the client instance is fully initialized (before pupBrowser or authStrategy are set), the application crashes with a "Cannot read property 'close' of undefined" error.

## Description

Added optional chaining (?.) to both ```pupBrowser.close()``` and ```authStrategy.destroy()``` calls in the ```destroy()``` method to prevent crashes when the client instance isn't fully initialized. This makes the method more resilient by gracefully handling cases where these properties aren't yet available.

## Motivation and Context

This change is required because:
- Currently, calling ```destroy()``` during initialization leads to unhandled exceptions
- Users should be able to safely terminate a client at any stage of its lifecycle
- This matches common patterns for resource cleanup in JavaScript applications

### Test Cases:
Immediate destruction:
- Created client instance
- Called ```destroy()``` immediately after construction
- Verified no crashes occur

Mid-initialization destruction:
- Created client instance
- Called ```destroy()``` during initialization process
- Verified graceful handling

Normal destruction:
- Created fully initialized client
- Called ```destroy()```
- Verified resources are properly cleaned up

Multiple calls:
- Called ```destroy()``` multiple times
- Verified no errors and no duplicate cleanup

### Environment

- Machine OS: MacOS
- Phone OS: iOS 18.4.1
- Library Version: 
- WhatsApp Web Version: 1.31.0
- Puppeteer Version: 18.2.1
- Browser Type and Version: Chromium 120
- Node Version: v22.9.0

## Types of changes

- [X] Dependency change
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] I have updated the documentation accordingly (index.d.ts). (Not required for this change)
- [ ] I have updated the usage example accordingly (example.js) (Not required for this change)


purpshell:
Resolved in a recently merged PR.

PR #3640, opened 2025-07-08
fix: 🐛 group lid participants

### LINKED ISSUE #3639: group participants fix causing group metadata error

### Describe the bug.

In the previous fix for issue #3587 , we modified the data of whatsapp original groupMetadata object directly isntead of just modify the model object. This is not a good pratice and well cause the group info page unusable. There might be further problems we don't know yet.

```
ErrorUtils caught an error:

Unknown data encountered while rendering. Item key: xxxxxxxxxx@c.us

Subsequent non-fatal errors won't be logged; see https://fburl.com/debugjs. JSHandle@object
```

### Expected Behavior

The group info page work as usual.

Get any group chat with code, then open group info page in Chromium. (With headless set to false)

### WhatsApp Account Type

Standard

### Browser Type

Chromium

### Operation System Type

Mac

### Phone OS Type

MacOs

### WhatsApp-Web.js Version

1.31.1-alpha.0

### WhatsApp Web Version

2.3000.1024526298

### Node.js Version

22

### Authentication Strategy

LocalAuth

### Additional Context

Image

hcfw007:
# PR Details

Modify model instead of origin whatsapp object.

## Description

With lid info retained in participant object, we can get lid from regular contact id. This allow us to fix mention and room remove problems later since the lid introduced.
Please see issue for detail

## Related Issue(s)

closes #3639

### Environment

- Machine OS: 
- Phone OS: 
- Library Version: 
- WhatsApp Web Version: 
- Puppeteer Version:
- Browser Type and Version: 
- Node Version: 

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -587,10 +587,13 @@ exports.LoadUtils = () => {
             model.isGroup = true;
             const chatWid = window.Store.WidFactory.createWid(chat.id._serialized);
             await window.Store.GroupMetadata.update(chatWid);
-            chat.groupMetadata.participants._models
-                .filter(x => x.id?._serialized?.endsWith('@lid'))
-                .forEach(x => x.contact?.phoneNumber && (x.id = x.contact.phoneNumber));
             model.groupMetadata = chat.groupMetadata.serialize();
+            model.groupMetadata.participants = chat.groupMetadata.participants._models.map(item => {
+                const result = item.serialize();
+                result.lid = result.id;
+                result.id = item.contact?.phoneNumber || result.lid;
+                return result;
+            });
             model.isReadOnly = chat.groupMetadata.announce;
         }
 


src/util/Injected/Utils.js
```diff
@@ -587,10 +587,13 @@ exports.LoadUtils = () => {
...
             const chatWid = window.Store.WidFactory.createWid(chat.id._serialized);
             await window.Store.GroupMetadata.update(chatWid);
-            chat.groupMetadata.participants._models
-                .filter(x => x.id?._serialized?.endsWith('@lid'))
-                .forEach(x => x.contact?.phoneNumber && (x.id = x.contact.phoneNumber));
             model.groupMetadata = chat.groupMetadata.serialize();
+            model.groupMetadata.participants = chat.groupMetadata.participants._models.map(item => {
+                const result = item.serialize();
+                result.lid = result.id;
+                result.id = item.contact?.phoneNumber || result.lid;
```
fernandomondo, line 594: I suggest adding a check so that lid is only set when the result.id is actually of type lid.
This avoids adding unnecessary properties for @c.us or @g.us Ids:

```suggestion
                if (result.id.server === 'lid') {
                    result.lid = result.id;
                    result.id = item.contact?.phoneNumber || result.lid;
                }              
```

hcfw007:
As the issue is now closed, I don't think hide lid all around is a valid suggetion now.  Therefore i'll close this PR.

fernandomondo:
I’m not sure it’s a good idea to close these PRs. 
Changing the source model from WhatsApp itself instead of the whatsapp-web.js model can cause several issues.
I even suspect this might be one of the causes of the high CPU usage.

PR #3668, opened 2025-07-24
feat: getContactLidOrPhone

BenyFilho:
# PR Details

New feature getContactLidAndPhone

## Description

The PR add Client.getContactLidAndPhone method to retrieve the @lid or the phone, depending on what was passed as a parameter.

## Usage Example

```js
// client initialization...

client.on('ready', async () => {

    let user = await client.getContactLidAndPhone(phoneNumber@c.us);
    console.log(user);
    
    //OR
    
    let user = await client.getContactLidAndPhone(xxxxxx@lid);
    console.log(user);

/*

    Output example:
   {
     lid: '213456789@lid',
     pn: '987546523335@c.us'
   }

*/
});
```

## You can try the fix by running one of the following commands:

- NPM
`npm install github:BenyFilho/whatsapp-web.js#getContactLidOrPhone`

- YARN
`yarn add github:BenyFilho/whatsapp-web.js#getContactLidOrPhone`

## Motivation and Context

Some people are struggling with it. 

## Types of changes

- [ ] Dependency change

## Checklist


alechkos (commented):
The thing is that you will be able to retrieve someone's phone number from `@­lid` only if that user is either in your contacts or you both already have an opened conversation (chat).
You can test it by using `@­lid` of some random user from one of your groups.

BenyFilho:
> The thing is that you will be able to retrieve someone's phone number from `@­lid` only if that user is either in your contacts or you both already have an opened conversation (chat). You can test it by using `@­lid` of some random user from one of your groups.

Weird, because I tested with a contact that is not my contact.

alechkos:
> > The thing is that you will be able to retrieve someone's phone number from `@­lid` only if that user is either in your contacts or you both already have an opened conversation (chat). You can test it by using `@­lid` of some random user from one of your groups.
> 
> Weird, because I tested with a contact that is not my contact.

Can confirm, works.

src/Client.js
```diff
@@ -2180,6 +2180,18 @@ class Client extends EventEmitter {
...
+
+    /**
+     * Get lid or phone number 
+     * @param {string} userId
+     * @returns {Promise<string>}
+     */
+    async getContactLidOrPhone(userId) {
+        return await this.pupPage.evaluate(async (userId) => {
+            const wId = await window.Store.WidFactory.createWid(userId);
+            return userId.indexOf('lid')>-1 ? await window.Store.LidUtils.getPhoneNumber(wId)._serialized : await window.Store.LidUtils.getCurrentLid(wId)._serialized;
```
alechkos: What do you think about simplifying a bit?
`userId.indexOf('lid')>-1` to `wid.server === 'lid'`

```suggestion
            return wid.server === 'lid'
                ? window.Store.LidUtils.getPhoneNumber(wid)._serialized
                : window.Store.LidUtils.getCurrentLid(wid)._serialized;
```

Also `window.Store.LidUtils.getPhoneNumber` and `window.Store.LidUtils.getCurrentLid` functions are synchronous so we don't need `await`.

alechkos (changes requested):
Can you please revert changes in `docs/Client.html` and `docs/index.html`.

src/Client.js
```diff
@@ -2180,6 +2180,18 @@ class Client extends EventEmitter {
...
     }
+
+    /**
+     * Get lid or phone number 
+     * @param {string} userId
+     * @returns {Promise<string>}
+     */
+    async getContactLidOrPhone(userId) {
+        return await this.pupPage.evaluate(async (userId) => {
+            const wId = await window.Store.WidFactory.createWid(userId);
```
alechkos: Since `window.Store.WidFactory.createWid` function is synchronous, you don't need to use `await` here.
```suggestion
            const wid = window.Store.WidFactory.createWid(userId);
```

src/Client.js
```diff
@@ -2180,6 +2180,18 @@ class Client extends EventEmitter {
...
         }, phoneNumber);
     }
+
+    /**
+     * Get lid or phone number 
+     * @param {string} userId
+     * @returns {Promise<string>}
+     */
+    async getContactLidOrPhone(userId) {
+        return await this.pupPage.evaluate(async (userId) => {
```
alechkos: You can freely remove `async` here, because there is no asynchronous functions inside the callback.

```suggestion
        return await this.pupPage.evaluate((userId) => {
```

alechkos (changes requested):

alechkos (changes requested):
@BenyFilho 
Can you please revert changes in `docs/Client.html` and `docs/index.html`.
After each release docs are being generated automatically.

alechkos (dismissed):
LGTM

shirser121:
I dont understand what is the point.
Why not return an object with the two options like:
`{
lid: ...,
pn: ...
}`

BenyFilho:
> I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`

I see your point, but I just tried to make it easy for the users

BenyFilho:
> I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`

Done! returning the object

shirser121:
> > I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`
> 
> Done! returning the object

This is not what i meant.
I ask why not return both lid and pn as result?

BenyFilho:
> > > I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`
> > 
> > 
> > Done! returning the object
> 
> This is not what i meant. I ask why not return both lid and pn as result?

Because, in general, people have one of them, lid or phone, and want to know the other one.

shirser121:
> > > > I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`
> > > 
> > > 
> > > Done! returning the object
> > 
> > 
> > This is not what i meant. I ask why not return both lid and pn as result?
> 
> Because, in general, people have one of them, lid or phone, and want to know the other one.

I want consistency so that the function always returns the same result, allowing people to get either lid or pn.

BenyFilho:
> > > > > I dont understand what is the point. Why not return an object with the two options like: `{ lid: ..., pn: ... }`
> > > > 
> > > > 
> > > > Done! returning the object
> > > 
> > > 
> > > This is not what i meant. I ask why not return both lid and pn as result?
> > 
> > 
> > Because, in general, people have one of them, lid or phone, and want to know the other one.
> 
> I want consistency so that the function always returns the same result, allowing people to get either lid or pn.

I got it! and done!

src/Client.js
```diff
@@ -2180,6 +2180,21 @@ class Client extends EventEmitter {
...
+     * @param {string} userId
+     * @returns {Promise<string>}
+     */
+    async getContactLidOrPhone(userId) {
+        return await this.pupPage.evaluate((userId) => {
+            const wid = window.Store.WidFactory.createWid(userId);
+            const isLid = wid.server === 'lid';
+            const lid = isLid ? wid : window.Store.LidUtils.getCurrentLid(wid);
+            const phone = isLid ? window.Store.LidUtils.getPhoneNumber(wid) : wid;
+            return { lid, phone };
```
alechkos: What do you think about to return `lid` and `pn` as strings inside the returning object and not as objects?
Like this:

```js
{
  lid: '153445321869023@lid',
  pn: '12124567890@c.us'
}
```

I think it will be more sutable here since `wid` is being used only in internal lib implementations and users always use a raw `jid` that is `XXXXXXXXXXX@c.us`

```suggestion
            const wid = window.Store.WidFactory.createWid(userId);
            const isLid = wid.server === 'lid';
            const lid = isLid ? wid : window.Store.LidUtils.getCurrentLid(wid);
            const phone = isLid ? window.Store.LidUtils.getPhoneNumber(wid) : wid;
            return {
                lid: lid._serialized,
                pn: phone._serialized
            };
```

src/Client.js
```diff
@@ -2180,6 +2180,21 @@ class Client extends EventEmitter {
             return await window.Store.AddressbookContactUtils.deleteContactAction(phoneNumber);
         }, phoneNumber);
     }
+
+    /**
+     * Get lid or phone number 
+     * @param {string} userId
+     * @returns {Promise<string>}
```
alechkos: Please update the returning object here if you agree with the comment above

```suggestion
     * @returns {Promise<{ lid: string, pn: string }>}
```

index.d.ts
```diff
@@ -242,6 +242,9 @@ declare namespace WAWebJS {
 
         /** Deletes the contact from user's addressbook */
         deleteAddressbookContact(honeNumber: string): Promise<void>
+
+        /** Get Contact Lid Or Phone */
+        getContactLidOrPhone(userId: string): Promise<object>
```
alechkos:
```suggestion
        getContactLidOrPhone(userId: string): Promise<{ lid: string; pn: string }>
```

alechkos (approved):
LGTM

Issue #319, opened 2020-08-27
Send Image


NeryHenrique:
### Bug description
Hello, like in issue #318, i cannot send imagens.
I already tryed to download the lastest code but for me it continue not working.

I m using yesterdays commit.

Thanks, for the help!

To get the error i run

-----------------------------------------------------------------------

const { MessageMedia } = require('whatsapp-web.js');

async function sendFilemsg(file_path,cel_number,msg){
    const media = MessageMedia.fromFilePath(file_path);
    console.log(media);
    console.log(await client.sendMessage(`${cel_number}@c.us`, media, {caption: msg}) );
    console.log("msg sent");
    return 1;
}

sendFilemsg("./imgtest.jpg",5511991225830,"caption test");

-----------------------------------------------------------------------

the console.log outputs
Message {
  mediaKey: undefined,
  id: {
    fromMe: true,
    remote: {
      server: 'c.us',
      user: '5511991225830',
      _serialized: '5511991225830@c.us'
    },
    id: '3EB0FD31CB46294A8A37',
    _serialized: 'true_5511991225830@c.us_3EB0FD31CB46294A8A37'
  },
  ack: 0,
  hasMedia: false,
  body: {
    mimetype: 'image/jpeg',
    data: '/9j/4AAQSkZJRgABAQEASABIAAD/2wBDAAMCAgMCAgMDAwMEAwMEBQgFBQQEBQoHBwYIDAoMDAsKCwsNDhIQDQ4RDgsLEBYQERMUFRUVD[.............................the image data goes forever....................................]7/fELCkrvgOVYOfBQmuMaeKU5FzAzjWVQSN5WB9MGbzBCZamBd84iqMVheJ+9lYfA86ltf2zf8AxOT8zyeoLxPdaOfwlcnxHX/ij/wDrF//Z',
    filename: 'imgtest.jpg'
  },
  type: 'chat',
  timestamp: 1598498056,
  from: '5511957087794@c.us',
  to: '5511991225830@c.us',
  author: undefined,
  isForwarded: false,
  fromMe: true,
  hasQuotedMsg: false,
  location: undefined,
  mentionedIds: []
}

guicuton:
This is not a "bug".

This happens because the callback of message cames first than image upload on WA Servers so thats why you have the base64 on message body...

try set as `await` the MessageMedia.fromFilePath function,,,

Here i'm using like this for external files:

```javascript
[...]

Issue #299, opened 2020-08-04
Can't send image using sendMessage() function


chiragkhichadiya:
I'm using this code to send image but it's not sending image and not giving any error either.

[...]

PR #3677, opened 2025-07-31
Update Utils.js

brunox64:
# PR Details

Add polling mechanism to `sendMessage` in Injected.js to handle race conditions when sending base64 images.

## Description

This PR introduces a short polling loop in the `window.WWebJS.sendMessage` function within `Injected.js`. After sending the message using `await window.Store.SendMessage.addAndSendMsgToChat(chat, message);`, the code now waits up to 2 seconds (with 100ms intervals) for the message to become available in `window.Store.Msg` before returning it. This addresses timing issues where the message ID is generated successfully, but `Store.Msg.get` fails to retrieve the message object immediately, particularly for base64-encoded media like images due to asynchronous upload and processing delays in WhatsApp Web.

The updated code snippet is as follows:

```javascript
window.WWebJS.sendMessage = async (chat, content, options = {}) => {
    // ... (existing code for preparing message and newMsgId)

    await window.Store.SendMessage.addAndSendMsgToChat(chat, message);

    // Add polling to wait for the message to appear in Store.Msg
    let returnedMsg = window.Store.Msg.get(newMsgId._serialized);
    let attempts = 0;
    const maxAttempts = 20; // Adjust as needed; e.g., waits up to ~2 seconds
    while (!returnedMsg && attempts < maxAttempts) {
        await new Promise(resolve => setTimeout(resolve, 100)); // 100ms delay per attempt
        returnedMsg = window.Store.Msg.get(newMsgId._serialized);
        attempts++;
    }

    if (!returnedMsg) {
        console.error('Message not found in Store.Msg after polling');
        // Optionally throw an error or return null/undefined
    }

    return returnedMsg;
};
```

This fix ensures more reliable retrieval of the sent message object without altering the core sending logic.

## Related Issue(s)

Related to #299, #319, #771 (issues involving media sending where messages are sent but not properly rendered or accessible due to async delays).

## Motivation and Context

This change is required to resolve a race condition in media sending (e.g., base64 images) via `Chat.sendMessage`. The send operation completes and generates a message ID, but the subsequent lookup in `Store.Msg` often fails because WhatsApp's internal media processing (upload/conversion) delays the message's full registration in the store. This leads to errors or null returns when trying to access the sent message immediately after sending. The polling approach mitigates this without introducing significant overhead.

## How Has This Been Tested

- Tested by sending base64-encoded images to individual and group chats using `Chat.sendMessage`.
- Verified that the message is successfully sent, the ID is generated, and the polling loop retrieves the message object reliably (no more null returns).
- Checked for no regressions in text-only messages or other media types.
- Monitored console for errors and confirmed the polling exits early when the message is available.
- Tested with varying base64 sizes to simulate different upload delays.

### Environment

- Machine OS: Windows
- Phone OS: Android
- Library Version: 1.31.0
- WhatsApp Web Version: 2.2530.1016 (obtained via `await client.getWWebVersion()`)
- Puppeteer Version: 22.12.1
- Browser Type and Version: Chromium 126.0.6478.126
- Node Version: 20.15.1

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -306,7 +306,25 @@ exports.LoadUtils = () => {
         await window.Store.HistorySync.sendPeerDataOperationRequest(3, {
             chatId: chat.id
         });
-        return window.Store.Msg.get(newMsgKey._serialized);
+
+        // return window.Store.Msg.get(newMsgKey._serialized);
+
+        // Add polling to wait for the message to appear in Store.Msg
+        let returnedMsg = window.Store.Msg.get(newMsgKey._serialized);
+        let attempts = 0;
+        const maxAttempts = 20; // Adjust as needed; e.g., waits up to ~2 seconds
+        while (!returnedMsg && attempts < maxAttempts) {
+            await new Promise(resolve => setTimeout(resolve, 100)); // 100ms delay per attempt
+            returnedMsg = window.Store.Msg.get(newMsgKey._serialized);
+            attempts++;
+        }
+    
+        if (!returnedMsg) {
+            console.error('Message not found in Store.Msg after polling');
+            // Optionally throw an error or return null/undefined
+        }
+    
+        return returnedMsg;
     };
 	
     window.WWebJS.editMessage = async (msg, content, options = {}) => {


fernandomondo:
You can remove sendPeerDataOperationRequest — this is the main cause of high CPU usage.

I currently have this in my branch and it’s working well:

```

await window.Store.SendMessage.addAndSendMsgToChat(chat, message);

// await window.Store.HistorySync.sendPeerDataOperationRequest(3, {
//     chatId: chat.id
// });
// return window.Store.Msg.get(newMsgKey._serialized);

const waitForMsg = async (msgId, attempts = 10, delay = 100) => {
    for (let i = 0; i < attempts; i++) {
        const msg = window.Store.Msg.get(msgId);
        if (msg) return msg;
        await new Promise(res => setTimeout(res, delay));
    }
    return null;
};

return await waitForMsg(newMsgKey._serialized);
```

alechkos:
Absorbed by #3692

PR #3693, opened 2025-08-08
fix: corrected the coment message

crazynds:
# PR Details

Correction only in a event description. Not any important change.


alechkos:
Absorbed by #3004

PR #3180, opened 2024-07-03
feat: Switch between QR or phone number pairing method

MobCode100:
# PR Details

The ability to change the preferred login/authentication method either use QR code or pair with a phone number (Linked devices -> Link a device -> Link with phone number instead).

## Description

1. Added new option `pairWithPhoneNumber` which is an object containing the parameters for `requestPairingCode` function in Client.js. The attributes of this object are:
    - `phoneNumber`: Phone number in international, symbol-free format <COUNTRY_CODE><PHONE_NUMBER>
    - `showNotification`: Show notification to pair for the specified phone number
    - `intervalMs`: The interval in milliseconds on how frequent to generate pairing code (WhatsApp default to 3 minutes)
2. On `code` and `window.onCodeReceivedEvent` event will run a callback once a code has been generated. 

The `inject` function in Client.js will check if the `phoneNumber` option is set, and will switch to phone number pairing mode without running any QR related functions. `intervalMs` is the time to wait for the next generation of code to mitigate the problem of expired code.

## Related Issues

closes ##3370 fixes #3371 fixes #3416

## Motivation and Context

1. Make it easier to use the `requestPairingCode` without having to manually utilize the function
2. The code feels cleaner this way, normalizing the authentication method of QR and phone number
3. Automatically generate new pairing code on/before expired.

## How Has This Been Tested

To run or test this feature, run `node app.js` where app.js contains the following code:
```
const { Client, LocalAuth } = require('whatsapp-web.js');
const client = new Client(
    {
        puppeteer: {
            args: ['--no-sandbox'],
            headless: false
        },
        authStrategy: new LocalAuth({ dataPath: "." }),
        pairWithPhoneNumber: {
            phoneNumber: "12324354646",
            showNotification: true,
        },
    }
);
client.on('code', (code) => {
    console.log("Linking code:",code);
});
client.initialize();
```
The pairing code should be displayed to the terminal every 3 minutes (WhatsApp current default). Developer tools are used to inspect some variables while running in non-headless mode. So far, no issues have been encountered during testing on QR code functions or any other existing features.

## Types of changes

- [ ] Dependency change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly (index.d.ts).

PR #3706, opened 2025-08-14
fix: Expose onCodeReceivedEvent function

avoylenko:
# PR Details

The `requestPairingCode` method is no longer works and throws `Evaluation failed: TypeError: window.onCodeReceivedEvent is not a function\n    at pptr://__puppeteer_evaluation_script__:20:27` error.

## Description

The `window.onCodeReceivedEvent` is being in injected only if the `pairWithPhoneNumber` option is passed to the Client instance.

## Related Issue(s)

Caused by https://github.com/pedroslopez/whatsapp-web.js/pull/3180

## Motivation and Context

The `requestPairingCode` method should work not only if new option is passed.

## How Has This Been Tested

Locally using my machine.

```
const { Client, LocalAuth } = require('whatsapp-web.js');
const client = new Client(
    {
        puppeteer: {
            args: ['--no-sandbox'],
            headless: false
        },
        authStrategy: new LocalAuth({ dataPath: "." }),
    }
);

client.on('ready', (code) => {
    client.requestPairingCode('38067XXXXXXX', true);
});

client.on('code', (code) => {
    console.log("Linking code:",code);
});

client.initialize();
```

## Types of changes

- [ ] Dependency change

## Checklist

- [X] I have updated the usage example accordingly (example.js)

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -144,17 +144,17 @@ class Client extends EventEmitter {
             }
 
             // Register qr/code events
+            await exposeFunctionIfAbsent(this.pupPage, 'onCodeReceivedEvent', async (code) => {
+                /**
+                * Emitted when a pairing code is received
+                * @event Client#code
+                * @param {string} code Code
+                * @returns {string} Code that was just received
+                */
+                this.emit(Events.CODE_RECEIVED, code);
+                return code;
+            });
             if (pairWithPhoneNumber.phoneNumber) {
-                await exposeFunctionIfAbsent(this.pupPage, 'onCodeReceivedEvent', async (code) => {
-                    /**
-                    * Emitted when a pairing code is received
-                    * @event Client#code
-                    * @param {string} code Code
-                    * @returns {string} Code that was just received
-                    */
-                    this.emit(Events.CODE_RECEIVED, code);
-                    return code;
-                });
                 this.requestPairingCode(pairWithPhoneNumber.phoneNumber, pairWithPhoneNumber.showNotification, pairWithPhoneNumber.intervalMs);
             } else {
                 let qrRetries = 0;


alechkos:
`Client.requestPairingCode` works with `pairWithPhoneNumber` option only.
Try running that code and you'll see what it does to your client state.

```js
// ...

client.on('change_state', (state) => {
    console.log(`[STATE] ${state}${state === WAState.CONNECTED ? '\n' : ''}`);
});

client.on('ready', (code) => {
    await client.requestPairingCode('38067XXXXXXX', true);
});

client.initialize();
```

avoylenko:
@alechkos indeed, I have provided the issue with incorrect code, here are the right one:

```
const { Client, LocalAuth } = require('whatsapp-web.js');

const client = new Client(
    {
        puppeteer: {
            args: ['--no-sandbox'],
            headless: true
        },
        authStrategy: new LocalAuth({ dataPath: "." }),
    }
);

client.on('code', (code) => {
    console.log("Linking code:",code);
});

(async () => {
    await client.initialize();
    await new Promise(r => setTimeout(r, 5000)); // small delay to be sure that instance is ready
    const code = await client.requestPairingCode('38067XXXXXXX', true);
    console.log("Linking code from method:",code);
})();
```

alechkos:
This won't work, your state will change in a circle `OPENING` -> `PAIRING` -> `CONNECTED`.
Also using `await client.initialize();` is not proper. `Client.initialize` is synchronous function that just registers all event listeners defined on `Client`.

avoylenko:
@alechkos thank you for replying. The main issue is that the method `window.onCodeReceivedEvent` is being injected only if the `Client` is started with  `pairWithPhoneNumber` option. This breaks the logic here https://github.com/pedroslopez/whatsapp-web.js/blob/bd677a8a66d44cea8f6d0d4bdb56e41cd0c0ccd5/src/Client.js#L387 which means that you need to know in advance how you will pair the client. Also the recent change adds a breaking change.

tofers:
Never break backward compatibility! 

Those who suffer, I'll throw away my crutch to do as before.

``` 
/* create file untils/Puppeteer.js */

/**
 * Expose a function to the page if it does not exist
 *
 * NOTE:
 * Rewrite it to 'upsertFunction' after updating Puppeteer to 20.6 or higher
 * using page.removeExposedFunction
 * https://pptr.dev/api/puppeteer.page.removeExposedFunction
 *
 * @param {import(puppeteer).Page} page
 * @param {string} name
 * @param {Function} fn
 */
async function exposeFunctionIfAbsent(page, name, fn) {
    const exist = await page.evaluate((name) => {
        return !!window[name];
    }, name);
    if (exist) {
        return;
    }
    await page.exposeFunction(name, fn);
}

module.exports = {exposeFunctionIfAbsent};
```

```
const {exposeFunctionIfAbsent} = require('./untils/Puppeteer');

/* app exress */
app.get('/setAuthenticationPhoneNumber', async (req, res) => {
    let phone = req.query.phone;
    if (phone === undefined) {
        res.status(400).json({error_code: 400, error_text: "required phone"});
        res.end();
    }

 /* emit event  */
        await exposeFunctionIfAbsent(client.pupPage, 'onCodeReceivedEvent', async (code) => {
            /**
             * Emitted when a pairing code is received
             * @event Client#code
             * @param {string} code Code
             * @returns {string} Code that was just received
             */
            client.emit('code', code);
            return code;
        });
        requestPairingCode = await client.requestPairingCode(phone, true);
    

    res.json({code: requestPairingCode});
});
```

PR #3719, opened 2025-08-26
fix:isAnnounce

mission-sun:
The "announce" field is currently being used but has not been defined. Additionally, the default value of this field should be set to true; otherwise, only administrators will be able to send messages after a new group is created.


mission-sun:
Problem Description: When a new group is created, only administrators are able to send messages—regular members do not have this permission.
Root Cause: Incorrect configuration of the "announce" parameter.

alechkos:
No

mission-sun:
> No

why ? 

The parameter is defined as isAnnounce, but announce is used. Isn't there an error?

PR #3777, opened 2025-09-05
Refactor participant assignment in Utils.js

bahadirduzcan:
### PR Details

This commit resolves a `TypeError` that occurs when sending messages to a group chat. The underlying WhatsApp Web API has deprecated the `window.Store.WidFactory.toUserWidOrThrow` method, causing the `sendMessage` function to fail.

The `participant` variable, which was previously set using the now-removed `toUserWidOrThrow` method, is now directly assigned the `from` object. This approach bypasses the deprecated method and ensures that message keys are created correctly, restoring the ability to send messages to groups.

## Description

This change modifies the `Utils.js` file to adapt to a recent WhatsApp Web API change. The `toUserWidOrThrow` method, used for determining a message's participant in group chats, has been deprecated. This fix replaces the problematic function call with a direct assignment of the `from` object, which already holds the correctly formatted participant ID.

## Motivation and Context

This change is required to maintain compatibility with the latest versions of WhatsApp Web. Without this fix, users encounter a `TypeError` when attempting to send messages to groups, which breaks a core functionality of the library. This solution addresses the problem directly by adapting to the API changes.

## How Has This Been Tested

I have tested this change in my local project by successfully sending multiple messages to different group chats. Before this fix, the application would crash and throw the mentioned `TypeError`. After the change, messages are sent without any issues.

### Environment

- Machine OS: **[Your Machine OS]**
- Phone OS: **[Your Phone OS]**
- Library Version: **[Your whatsapp-web.js Version]**
- WhatsApp Web Version: **[Your WWeb Version]** (You can get this by running `await client.getWWebVersion()`)
- Puppeteer Version: **[Your Puppeteer Version]**
- Browser Type and Version: **[Your Browser Type and Version]**
- Node Version: **[Your Node Version]** (You can get this by running `node -v`)

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -249,7 +249,9 @@ exports.LoadUtils = () => {
 
         if (typeof chat.id?.isGroup === 'function' && chat.id.isGroup()) {
             from = chat.groupMetadata && chat.groupMetadata.isLidAddressingMode ? lidUser : meUser;
-            participant = window.Store.WidFactory.toUserWidOrThrow(from);
+            // Note: The WhatsApp Web API has deprecated the `WidFactory.toUserWidOrThrow` method.
+            // The `from` object already contains the valid user ID, so we can use it directly.
+            participant = from;
         }
 
         const newMsgKey = new window.Store.MsgKey({


nwillian:
Perfect. Worked perfectly.

alechkos (commented):
#3778 suggests better solution.

PR #3797, opened 2025-09-09
fix: getPinnedMessages()

RC047:
# PR Details

## Description
Here, when i tried to pin a message to a group chat and wanted to retrieve the pinned messages using `chat.getPinnedMessages()` its thrown an error like this:
```
Error [TypeError]: window.Store.PinnedMsgUtils.getTable is not a function
```
After investigating, i discovered the error was in the input of the store variable, i tried to fix it and the function worked perfectly again! So the problem is in adding a store function that has already been declared, so that the previous function is overwritten
https://github.com/pedroslopez/whatsapp-web.js/blob/93559bbe191529c958372cbd821092968bfb826d/src/util/Injected/Store.js#L100-L102

and I tried to change the declaration without changing the other functions to be like this.
```js
window.Store.PinnedMsgUtils = {
    ...window.require('WAWebPinInChatSchema'),
    ...window.require('WAWebSendPinMessageAction')
};
```

## Types of changes

- [ ] Dependency change

## Checklist


alechkos (approved):

BenyFilho (approved):
I tested and it works

BenyFilho (approved):
Tested and approved

RC047 (commented):
From me it still works perfectly, but when i unpin the message and then re-execute the `chat.getPinnedMessages()` function the result is still listed in the pinned message, is that really the case?

alechkos:
It's because the message is still is cache

shirser121:
> It's because the message is still is cache

This not should be happening, dont you think?

src/Client.js
```diff
@@ -1225,11 +1225,18 @@ class Client extends EventEmitter {
             
             const msgs = await window.Store.PinnedMsgUtils.getTable().equals(['chatId'], chatWid.toString());
 
-            const pinnedMsgs = msgs.map((msg) => window.Store.Msg.get(msg.parentMsgKey));
+            const pinnedMsgs = (
+                await Promise.all(
+                    msgs.map(async (msg) => {
```
BenyFilho:
```suggestion
                    msgs.filter(msg => msg.pinType == 1).map(async (msg) => {
```
That solves the problem of getting old pinned messages. Now it will get only the pinned messages

BenyFilho (changes requested):

BenyFilho (approved):

alechkos (dismissed):

tuyuribr (approved):

src/Client.js
```diff
@@ -1225,11 +1225,18 @@ class Client extends EventEmitter {
...
 
-            const pinnedMsgs = msgs.map((msg) => window.Store.Msg.get(msg.parentMsgKey));
+            const pinnedMsgs = (
+                await Promise.all(
+                    msgs.filter(msg => msg.pinType == 1).map(async (msg) => {
+                        const res = await window.Store.Msg.getMessagesById([msg.parentMsgKey]);
+                        return res?.messages?.[0];
+                    })
+                )
+            ).filter(Boolean);
```
purpshell, line 1235: that .filter makes no sense

BenyFilho: Without that filter, it will return all pinned messages, even those that are not pinned anymore

RC047: is it enough to just filtering the pin type? or what do you think?

purpshell (approved):

Issue #3809, opened 2025-09-16
Client logs in but is forcefully logged out after a few seconds with the error `Execution context was destroyed`.


njthossein:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this a problem caused by your code, or is it specifically because of the library?

- [x] I have double-checked my code carefully.

### Describe the bug.

When initializing the client with LocalAuth, the login works fine and I can see "✅ WhatsApp client is ready!". However, after a few seconds, the session suddenly disconnects and crashes with the following error:

```
Error: Execution context was destroyed, most likely because of a navigation.
    at rewriteError (/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:284:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async ExecutionContext._ExecutionContext_evaluate (...)
    at async ExecutionContext.evaluate (...)
    at async Client.getWWebVersion (/node_modules/whatsapp-web.js/src/Client.js:878:16)
    at async Client.inject (/node_modules/whatsapp-web.js/src/Client.js:101:25)
    at async /node_modules/whatsapp-web.js/src/Client.js:356:13

```
This happens consistently on some WhatsApp accounts but not all. It looks like the client is being logged out (or the browser context is reloaded) immediately after login.

### Expected Behavior

Once logged in, the session should remain active and stable unless explicitly disconnected. Execution context was destroyed should not happen immediately after login.

### Steps to Reproduce the Bug or Issue

Install dependencies:

npm install whatsapp-web.js puppeteer qrcode-terminal express cors axios node-cron jalaali-js

Use this initialization code:

const client = new Client({
  authStrategy: new LocalAuth({ clientId: 'main' }),
  puppeteer: {
    headless: true,
    executablePath: '/usr/bin/chromium-browser',
    args: [
      '--no-sandbox',
      '--disable-setuid-sandbox',
      '--disable-dev-shm-usage',
      '--disable-accelerated-2d-canvas',
      '--disable-gpu'
    ]
  }
});

client.once('ready', () => {
  console.log('✅ WhatsApp client is ready!');
});

Scan QR code and wait ~5–10 seconds.

Observe that client disconnects with Execution context was destroyed.

### WhatsApp Account Type

WhatsApp Business

### Browser Type

Chromium-browser 114.0.5735.91

### Operation System Type

Ubuntu 22.04 (VPS)

### Phone OS Type

[...]

PR #3811, opened 2025-09-16
Fix Disconnections - **UPDATED REINSTALL IT**

BenyFilho:
# PR Details

Fixing the Inject Module Function and removing override isLidMigrated

## Description

Fixing the Inject Module Function and removing override isLidMigrated

## You can try the fix by running one of the following commands:

- NPM
`npm install github:BenyFilho/whatsapp-web.js#fix_disconnections`

- YARN
`yarn add github:BenyFilho/whatsapp-web.js#fix_disconnections`

## Motivation and Context

Some people are experiencing unexpected disconnections.
fixes: https://github.com/pedroslopez/whatsapp-web.js/issues/3809

## Types of changes

- [ ] Dependency change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project.
- [X] I have updated the documentation accordingly (index.d.ts).
- [ ] I have updated the usage example accordingly (example.js)

PR #3813, opened 2025-09-17
Fix: Avoid logout on accounts migrated to LID 1x1 model

AragaoAnderson:
# PR Details

Fixes an issue where migrated WhatsApp accounts were being logged out immediately after login due to the `isLidMigrated` function being forced to always return `false`.

## Description

This PR updates the injection of `Lid1X1MigrationUtils.isLidMigrated` in `src/util/Injected/Store.js`.

Previously the code patched the function to always return `false`:

```js
window.injectToFunction(
  { module: 'WAWebLid1X1MigrationGating', function: 'Lid1X1MigrationUtils.isLidMigrated' },
  () => false
);
```

That worked for accounts that hadn't been migrated yet, but caused an immediate logout on accounts already using the new **LID 1x1** system.

Now, the injection attempts to call the original function first and only falls back to returning `false` if executing the original function fails. This preserves compatibility with non-migrated accounts while avoiding forced logouts on migrated accounts.

## Related Issue(s)

[#3809](https://github.com/pedroslopez/whatsapp-web.js/issues/3809)

## Motivation and Context

This prevents unnecessary logouts for accounts migrated to the LID 1x1 model, while keeping existing behavior for non-migrated accounts. It reduces disruption for users who are already on the migrated backend.

## How Has This Been Tested

- Tested login with multiple accounts (both migrated and non-migrated).
- Confirmed non-migrated accounts still connect as before.
- Confirmed migrated accounts no longer get logged out immediately after login.

### Environment

- Machine OS: Linux (Debian)
- Phone OS: iOS (tested on iOS 26)
- Library Version: 1.34.1
- WhatsApp Web Version: 2.3000.1027224352
- Puppeteer Version: 18.2.1
- Browser Type and Version: Chromium 140.0.7339.52
- Node Version: 22.18.0

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Store.js
+++ b/src/util/Injected/Store.js
@@ -229,5 +229,5 @@ exports.ExposeStore = () => {
 
     window.injectToFunction({ module: 'WAWebE2EProtoUtils', function: 'typeAttributeFromProtobuf' }, (func, ...args) => { const [proto] = args; return proto.locationMessage || proto.groupInviteMessage ? 'text' : func(...args); });
 
-    window.injectToFunction({ module: 'WAWebLid1X1MigrationGating', function: 'Lid1X1MigrationUtils.isLidMigrated' }, () => false);
+    window.injectToFunction({ module: 'WAWebLid1X1MigrationGating', function: 'Lid1X1MigrationUtils.isLidMigrated' }, (func, ...args) => { try { return func(...args); } catch { return false; }});
 };


adamzayn101:
I have 100 WhatsApp numbers.

96 WhatsApp numbers are working.

4% had issues... I implemented both updates:
#3811
&
#3813

Now it works 100% on all numbers.

i350:
Work for me. Thanks
I tested on https://github.com/pedroslopez/whatsapp-web.js/pull/3811

mahechahe:
This change works for me. It should be merged now to make the library more stable.

hetrodoo:
This change solved the auto log out just after logging-in for me!

siddharthgavit451-arch (approved):

PR #3839, opened 2025-09-29
fix: unpin messages

RC047:
# PR Details
The pin message feature generally works perfectly after the changes in #3797 to get a list of pinned messages. However, the unpin message function does not work due to factors in the code input.

## Description
From what we can see, the pin message function works perfectly and has three arguments in its callback function, whereas the unpin function only has two callback function arguments
https://github.com/pedroslopez/whatsapp-web.js/blob/7304ec2135bce3605349c9219f207498bfaf7b7f/src/structures/Message.js#L553-L567

and this will cause the unpin function to always returning false and will not working because it is in this statement
https://github.com/pedroslopez/whatsapp-web.js/blob/7304ec2135bce3605349c9219f207498bfaf7b7f/src/util/Injected/Utils.js#L1125

So I tried to solve this problem by adding the missing argument to the function by adding the `duration` value to zero, to restore the function works properly.

## Types of changes

- [ ] Dependency change

## Checklist


BenyFilho (approved):
I tested and it works like a charm

tuyuribr (approved):

purpshell (approved):

PR #3847, opened 2025-10-01
avoid hang by closing browser only if connected

### LINKED ISSUE #3846: Client.destroy may hang after Browser disconnected, preventing cleanup (pupBrowser.close() never resolves/rejects)

### Describe the bug.

### Expected Behavior

4. Call client.destroy().

// Example
```
await client.destroy();
```

### WhatsApp Account Type

Standard

### Browser Type

Google Chrome

### Operation System Type

Windows

### Phone OS Type

Android

### WhatsApp-Web.js Version

1.34.1

### WhatsApp Web Version

2.3000.1027807705

### Node.js Version

20.19.4

### Authentication Strategy

LocalAuth

### Additional Context

Analysis

Proposed Fix

```
  if (isConnected) {
    await browser. Close();
  }
};
```

Adi1231234:
# PR Details

Fix hang in `Client.destroy()` when browser is already disconnected.

## Description

Skip `browser.close()` if Puppeteer browser is not connected, ensuring `authStrategy.destroy()` always runs and `Client.destroy()` resolves.

## Related Issue(s)

closes [#3846](https://github.com/pedroslopez/whatsapp-web.js/issues/3846)

## Motivation and Context

Prevents indefinite hang and allows cleanup to continue.

## How Has This Been Tested

Tested with connected and disconnected browser. In both cases `Client.destroy()` now returns.

### Environment
* OS: Windows 10
* Phone OS: Android
* Library: 1.34.1
* WhatsApp Web: 2.3000.1027807705
* Node: 20.19.4

## Types of changes

* [x] Bug fix

## Checklist
* [x] Code follows project style


adarshmadrecha:
Will be helpful if merged. Commenting for a bump.

RC047 (approved):
I agree with this change

tuyuribr:
Just to clarify, when does it hangs? Only if you connect puppeteer outside the current env ?

Adi1231234:
@tuyuribr The issue isn’t limited to cases using puppeteer.connect(). It can also happen in a normal setup if the computer goes into sleep mode - after waking up, the browser disconnects, and calling Client.destroy() hangs as well.

purpshell (approved):

tuyuribr (approved):

Adi1231234:
Thanks everyone!
@tuyuribr  Can we merge it?

purpshell (approved):

Adi1231234:
@tuyuribr @purpshell I fixed the eslint error, can we merge it? 🙏

PR #3900, opened 2025-10-13
fix: change require to window.require

fallenleavesguy:
# PR Details

just change require to window.require.
Because I use rollup to bundle  js sdk for web direct usage instead of puppeteer(see: https://github.com/swordlib/whatsapp-web.js/blob/export-web-sdk/web-sdk.rollup.config.js) . "require" from node.js consufe rollup, rollup considers it was require a node_module package. Should change require to window.require here. thank you.

## Motivation and Context

 "require" from node.js consufe rollup, rollup considers it was require a node_module package

### Environment

- Machine OS: 
- Phone OS: 
- Library Version: 
- WhatsApp Web Version: 
- Puppeteer Version:
- Browser Type and Version: 
- Node Version: 

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Store.js
+++ b/src/util/Injected/Store.js
@@ -103,7 +103,7 @@ exports.ExposeStore = () => {
     window.Store.FindOrCreateChat = window.require('WAWebFindChatAction');
     window.Store.CustomerNoteUtils = window.require('WAWebNoteAction');
     window.Store.BusinessGatingUtils = window.require('WAWebBizGatingUtils');
-    window.Store.PollsVotesSchema = require('WAWebPollsVotesSchema');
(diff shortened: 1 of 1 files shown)


purpshell:
Shouldn't this apply to all calls of require?

fallenleavesguy:
> Shouldn't this apply to all calls of require?

only for code that run in browser, not for code that run in node.js. and only this line use require instead of window.require.

RC047:
>
> only for code that run in browser, not for code that run in node.js. and only this line use require instead of window.require.

yes, and i've fixed that changes in this pull request. #3913

PR #3972, opened 2025-11-22
fix: (ready event bug related) loads WWeb modules if you're part of an A/B test

olob0:
# PR Details

NOTE: see this comment: https://github.com/pedroslopez/whatsapp-web.js/pull/3972#issuecomment-3792024301

loads the profile resources before calling WAWebSetPushnameConnAction for setPushname

## Description

ensures that `WAWebSetPushnameConnAction` is reliably available before attempting to call setPushname.

on some WhatsApp Web clients, the module isn’t loaded until the profile drawer resources are initialized.
To avoid the _"Require: unknown module WAWebSetPushnameConnAction"_ error, this patch explicitly loads the bundle `WAWebProfileDrawerLoadableRequireBundle` before calling any pushname-related actions.

## Related Issue(s)

closes #3971 

## Motivation and Context

some WhatsApp Web instances do not automatically load WAWebSetPushnameConnAction unless the user opens the profile drawer manually.
This results in random failures when calling setPushname

by forcing the required bundle to load beforehand, we have guarantee that the module exists across all clients, fixing the behavior without introducing breaking changes

## How Has This Been Tested

i saw the issue on affected clients where the pushname module was missing (wwebjs discord)

applied the patch to force-load `WAWebProfileDrawerLoadableRequireBundle`

shared the fix in the official wwebjs Discord, where another user tested it on a client affected by the same bug and confirmed the solution worked

Tested personally on my own environment (also affected by the issue) and confirmed that:

setPushname works reliably, and the `displayName` continue to behave normally

### Environment

Machine OS: Linux (Kubuntu 25.10)
Phone OS: Android 15
Library Version: whatsapp-web.js (tested on 1.34.2 and latest main)
WhatsApp Web Version: 2.3000.1030235766
Puppeteer Version: 18.2.1
Browser Type and Version: Chrome 142.0.7444.134 64 bits
Node Version: tested on 18.20.2 and 22.16.0

## Types of changes

- [ ] Dependency change

## Checklist


tuyuribr (approved):

RC047:
We also need to determine for certain conditions, and whether this will run automatically or manually, because this change could be a solution for some users, and it could be the opposite.

olob0:
@RC047 I think we can simply do this:

```js
...
if (isCometOrAbove) {
  // make sure the pushname module is ready before injection
  await this.pupPage.evaluate(async () => {
    const WWebSetPushnameConnAction = await window.require('WAWebSetPushnameConnAction');

    if (!WWebSetPushnameConnAction) {
      console.warn("WAWebSetPushnameConnAction not found. WAWebProfileDrawerLoadableRequireBundle will be called to try loading it.");

      await window.require("WAWebProfileDrawerLoadableRequireBundle").requireBundle();
      return;
    }
  });

  await this.pupPage.evaluate(ExposeStore);
}
...
```

image

i added the console.warn just to give anyone debugging some visibility into what’s happening

i tested it and it seems to work well

what do you think?

DanrwAU:
+1 also affected by this issue and confirmed this fix works for me.

olob0:
I've seen several reports from users saying the fix didn’t work, and at the same time, others saying it *did* work.

I spent a few more hours debugging and I think I found the root cause of the issue. `WAWebProfileDrawerLoadableRequireBundle` is one of the many modules that *aren’t* preloaded when WhatsApp Web starts.

I believe this is why some users are still having issues even after the fix. So I improved the fix to call `WAWebPrefetchLoadables`, which, as the name suggests, preloads all required modules.

I reached this conclusion after finding this line in WhatsApp Web’s code:

```js
o("WAWebABProps").getABPropConfigValue("wa_web_disable_prefetch_loadables") || r("WAWebPrefetchLoadables")()
```

As you can see, it only preloads the necessary modules if you're part of the A/B test. Bingo.

purpshell:
> @RC047 I think we can simply do this:
> 
> ```js
> ...
> if (isCometOrAbove) {
>   // make sure the pushname module is ready before injection
>   await this.pupPage.evaluate(async () => {
>     const WWebSetPushnameConnAction = await window.require('WAWebSetPushnameConnAction');
> 
>     if (!WWebSetPushnameConnAction) {
>       console.warn("WAWebSetPushnameConnAction not found. WAWebProfileDrawerLoadableRequireBundle will be called to try loading it.");
> 
>       await window.require("WAWebProfileDrawerLoadableRequireBundle").requireBundle();
>       return;
>     }
>   });
> 
>   await this.pupPage.evaluate(ExposeStore);
> }
> ...
> ```
> 
> image
> i added the console.warn just to give anyone debugging some visibility into what’s happening
> 
> i tested it and it seems to work well
> 
> what do you think?

This is the right change

DanrwAU:
I'm seeing weird loss of sessions with the latest commit on this. 
Doing this worked fine await this.pupPage.evaluate(async () => {
                      await window.require("WAWebProfileDrawerLoadableRequireBundle").requireBundle()
                    }); 

However the latest commit results in all sessions randomly stopping working after a period of time, nothing is ever logged (at least in my code) to indicate the sessions have stopped. 

Going back to just the code above fixes the issue.

olob0:
First of all, sorry for the delay.

@purpshell @tuyuribr I created a helper function to pre-load the required modules only if you use a function that depends on them. In my view, this is a much better solution, since it only requires mapping the necessary module, without relying on the A/B test flag.

I also ran into the same issue when forwarding messages (WAWebChatForwardMessage not found).

I tested it on two clients: one affected by the issue and another not. In both cases, `.forward()` and `.setDisplayName()` worked as expected, but testing on other clients is totally valid.

Note: I ended up running into #5741 and used https://github.com/pedroslopez/whatsapp-web.js/pull/5737/changes/9722d96589b7cc0fe0122180edad9d51cd1603ed to fix it and actually be able to test my changes. Not included in this PR

purpshell (approved):

BenyFilho:
@olob0  will you update it to comply with the latest changes? Do you know if this fix is still necessary?

olob0:
@BenyFilho I just tested the latest main and everything is working correctly! I'm going to close this PR.

PR #3975, opened 2025-11-23
Fix: Gracefully handle missing WAWebSetPushnameConnAction module (bug)

nmazuz:
# PR Details

Fix: Handle missing WAWebSetPushnameConnAction module gracefully to prevent initialization errors when using existing WhatsApp Web sessions. The module may not be available in all WhatsApp Web versions, so we now wrap the require statement in a try-catch block and provide a fallback function instead of throwing an error.
## Description

Adds error handling for the `WAWebSetPushnameConnAction` module requirement in `Store.js` to prevent "Requiring unknown module" errors during WhatsApp Web initialization with existing sessions.

The module `WAWebSetPushnameConnAction` may not be available in all WhatsApp Web versions or may not be loaded when the Store is being initialized. This change wraps the module requirement in a try-catch block and provides a fallback function that returns `false` when the module is unavailable, allowing the client to continue functioning normally.

## Motivation and Context

When initializing WhatsApp Web.js with an existing session (`.wwebjs_auth`), the client was throwing the following error:
**Requiring unknown module "WAWebSetPushnameConnAction"**
This error occurred because the code was directly requiring the module without checking if it exists:
ript
setPushname: window.require('WAWebSetPushnameConnAction').setPushnameThis caused the client to fail during initialization, preventing users from using existing sessions. The fix ensures that if the module is unavailable, the client can still initialize successfully, and the `setDisplayName` functionality will gracefully degrade (returning `false` when the module is not available).

## How Has This Been Tested

1. Tested initialization with existing session (`.wwebjs_auth` folder present)
2. Verified that the error no longer appears in console
3. Confirmed client initializes successfully and can receive/send messages
4. Tested that `setDisplayName` method returns `false` when module is unavailable (expected behavior)

### Environment

- Machine OS: macOS (darwin 24.6.0)
- Phone OS: Android 15
- Library Version: 1.34.2
- WhatsApp Web Version: 2.3000.1030235766
- Puppeteer Version: Chrome 142.0.7444.134 64 bits
- Browser Type and Version: Chromium (via Puppeteer)
- Node Version: 18.20.2

## Types of changes

[ ] Dependency change
[X] Bug fix (non-breaking change which fixes an issue)
[ ] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

[X] My code follows the code style of this project.
[ ] I have updated the documentation accordingly (index.d.ts).
[ ] I have updated the usage example accordingly (example.js)


olob0:
i don’t think degrading `setDisplayName` based on its "availability" is a good approach at moment. If it’s possible to edit your username on WhatsApp Web, then it should be possible to keep `setDisplayName` functional. Did you test whether `window.require('WAWebSetPushnameConnAction')` exists right after clicking on your profile? Can you check the files loaded at the moment the profile screen opens, inspecting the request initiator call stack?

That’s the path I followed to fix my issue in **#3972**.

my initial solution was also to deprecate setDisplayName, but after a bit more debugging, I found a way to keep it working

i suspect that depending on the client, WhatsApp Web uses a different module or mechanism

leonardolimafarias:
This is a bug that affects production use cases.
Would it be possible to mark it as priority: high or bug for visibility?

rafaheringer:
Plrease, accept this PR.

aloysioNibo:
I'm right now with the same issue here!! this fix will help me a lot

alvaromoura-nibo:
We are facing many problems, and this PR will probably fix them. Please approve it.

olob0:
For everyone who’s still having the issue even after my fix, I’ve just updated it. It now preloads all required modules in case you’re part of an A/B test.

see: https://github.com/pedroslopez/whatsapp-web.js/pull/3972#issuecomment-3576304519

```sh
npm i github:olob0/whatsapp-web.js#fix/setPushname
```

brunnonunes:
We’re currently experiencing issues related to this behavior, and this PR addresses them directly.
Could you please review and approve it when possible?
It would really help us move forward.

leonardolimafarias:
@nmazuz add BUG tag on title, pls

shirser121 (approved):

aliyss (approved):

leonardolimafarias:
Hi @tuyuribr  and @purpshell 
When you have a moment, could you please review this PR? We need two approvals to move ahead, and your input is highly appreciated.
Thank you very much!

src/util/Injected/Store.js
```diff
@@ -114,8 +114,17 @@ exports.ExposeStore = () => {
     window.Store.Settings = {
         ...window.require('WAWebUserPrefsGeneral'),
         ...window.require('WAWebUserPrefsNotifications'),
-        setPushname: window.require('WAWebSetPushnameConnAction').setPushname
+        setPushname: (() => {
+            try {
+                const module = window.require('WAWebSetPushnameConnAction');
+                return module && module.setPushname ? module.setPushname : async () => { return false; };
```
purpshell, line 120: This is not reasonable. This seems like a case of needing to wait till the module is loaded before trying to access it, in that case you need to create some sort of change in Client.js to append this function to the Store.Settings.

I would like to just get this merged and moved on with, I will likely make a PR for this case.

samuelfaj:
Thanks a lot, worked here!!

samuelfaj (approved):

RC047:
This PR is also related to #3972  but needs to be focused on one of them.

src/util/Injected/Store.js
purpshell: Follow up: We should likely create some logic that only injects certain modules later into the lifecycle or keeps track of all requires and if they fail it retries it later

purpshell:
closing in favor of #3972

PR #3993, opened 2025-11-29
Fix: Handle missing getIsMyContact in newer WhatsApp Web versions

bahadirduzcan:
# PR Details

Fixed `TypeError: window.Store.ContactMethods.getIsMyContact is not a function` by adding a fallback check.

## Description

Recent updates to WhatsApp Web have removed or moved the `getIsMyContact` method from `window.Store.ContactMethods`, causing a crash when the library attempts to retrieve contact information.

I have updated `src/util/Injected/Utils.js` to safely check for the existence of this method before calling it.

**Changes:**
- Modified `window.WWebJS.getContactModel` in `src/util/Injected/Utils.js`.
- Added a conditional check: if `getIsMyContact` exists, use it; otherwise, fall back to `contact.isMyContact` or `false`.

```javascript
res.isMyContact = window.Store.ContactMethods.getIsMyContact 
    ? window.Store.ContactMethods.getIsMyContact(contact) 
    : (contact.isMyContact !== undefined ? contact.isMyContact : false);
```

## Motivation and Context

This change is required to restore functionality for retrieving contacts, which was broken by upstream changes in WhatsApp Web. Without this fix, the client crashes with a `TypeError`.

## How Has This Been Tested

I have verified this fix locally by running the client and confirming that contact objects are serialized correctly without errors.

### Environment

- Machine OS: Windows
- Phone OS: 
- Library Version: 
- WhatsApp Web Version: 
- Puppeteer Version:
- Browser Type and Version: Chrome
- Node Version: 

## Types of changes

- [ ] Dependency change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly (index.d.ts).
- [ ] I have updated the usage example accordingly (example.js)

PR #3978, opened 2025-11-24
Fixing msg.getContact and getContactById

BenyFilho:
# PR Details

This PR fixes method msg.getContact() and client.getContactById

## Description

This PR fixes method msg.getContact() and client.getContactById

Closed https://github.com/pedroslopez/whatsapp-web.js/issues/3979

## How Has This Been Tested

```
// client initialization...

client.on('ready', async () => {
    const contact = await client.getContactById('XXXXXXX');
    console.log(contact);
});
```

## You can try the fix by running one of the following commands:

- NPM
`npm install github:BenyFilho/whatsapp-web.js#fix-getContact`

- YARN
`yarn add github:BenyFilho/whatsapp-web.js#fix-getContact`

## Types of changes

- [ ] Dependency change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly (index.d.ts).
- [ ] I have updated the usage example accordingly (example.js)

PR #4009, opened 2025-12-10
Fix: getIsMyContact is not a function

GorilAas:
The method `getIsMyContact` no longer exists in the latest WhatsApp Web API,
which caused a runtime error in `Utils.js`.

This PR replaces:
- getIsMyContact(contact)
with:
- getIsWAContact(contact)

This resolves the error and restores compatibility with the latest ContactMethods.


BenyFilho:
That error was solved some time ago, and merged.
https://github.com/pedroslopez/whatsapp-web.js/pull/3978

purpshell (commented):
shouldn't by this logic, the property itself get removed entirely?

purpshell:
Related to #3993

GorilAas:
> shouldn't by this logic, the property itself get removed entirely?

After checking the getIsWAContact(contact) function in devtools, I found that it has the same required fields as getIsMyContact(contact). After making this change, message sending started working again.

PR #5141, opened 2025-12-12
fix: prevent destroy() from hanging when browser is disconnected

### LINKED ISSUE #3846: Client.destroy may hang after Browser disconnected, preventing cleanup (pupBrowser.close() never resolves/rejects)

gn00295120:
## Summary

Fixes `Client.destroy()` hanging indefinitely when the Puppeteer browser is already disconnected.

## Problem

When the Puppeteer browser is disconnected (e.g., browser crashed, manually closed, or "Browser disconnected" event), calling `pupBrowser.close()` may hang forever without resolving or rejecting. This prevents:

1. `destroy()` from completing
2. `authStrategy.destroy()` from running (cleanup never happens)
3. Application code from continuing after `await client.destroy()`

```javascript
// Example - this never completes if browser is disconnected
await client.destroy();
```

## Solution

Add a check for `pupBrowser.isConnected()` before attempting to close:

```javascript
async destroy() {
    // Remove event listeners...
    
    // Only close browser if still connected to avoid hanging
    if (this.pupBrowser?.isConnected()) {
        await this.pupBrowser.close();
    }
}
```

## Benefits

- `destroy()` always completes in bounded time
- `authStrategy.destroy()` is always called for proper cleanup
- Application code can reliably continue after `await client.destroy()`
- Server restarts won't get stuck due to orphaned browser processes

## Test Plan

1. Start client and authenticate
2. Manually kill the Chrome/Chromium process
3. Call `client.destroy()`
4. Verify the call completes without hanging

## Related Issues

- Fixes #3846 - Client.destroy may hang after Browser disconnected
- Related to #3976 - Initialization stuck because Chromium remains running after server restart

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -850,6 +850,12 @@ class Client extends EventEmitter {
      * Closes the client
      */
     async destroy() {
+        // Remove all Puppeteer page event listeners to prevent memory leaks
+        if (this.pupPage) {
+            this.pupPage.removeAllListeners('framenavigated');
+            this.pupPage.removeAllListeners('request');
+            this.pupPage.removeAllListeners('response');
+        }
         await this.pupBrowser.close();
         await this.authStrategy.destroy();
     }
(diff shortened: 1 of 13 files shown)


gn00295120:
Closing in favor of a cleaner PR with a fresh branch from upstream/main.

PR #5195, opened 2025-12-12
fix: add memory leak prevention and input validation

gn00295120:
## Summary

This PR adds memory leak prevention in the destroy() method and input validation for core async methods.

### Memory Leak Prevention
- Remove event listeners from `pupPage` before closing browser in `destroy()` method
- Prevents memory leaks from orphaned event handlers

### Input Validation
Added validation for required parameters in:
- `sendMessage()` - validates chatId and content
- `getChatById()` - validates chatId
- `getContactById()` - validates contactId
- `getMessageById()` - validates messageId
- `createGroup()` - validates title and participants

## Why

- Prevents memory leaks when client is destroyed and recreated multiple times
- Provides clear error messages instead of cryptic failures when invalid inputs are passed

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -846,6 +846,12 @@ class Client extends EventEmitter {
      * Closes the client
      */
     async destroy() {
+        // Remove all Puppeteer page event listeners to prevent memory leaks
+        if (this.pupPage) {
+            this.pupPage.removeAllListeners('framenavigated');
+            this.pupPage.removeAllListeners('request');
+            this.pupPage.removeAllListeners('response');
+        }
         await this.pupBrowser.close();
         await this.authStrategy.destroy();
     }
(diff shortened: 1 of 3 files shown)


gn00295120:
Thanks for the thorough review! I've addressed several suggestions:

## Applied Changes (commit 411d9b3)

1. **removeAllListeners()** - ✅ Changed to use `removeAllListeners()` without arguments to remove all listeners
2. **pupBrowser null check** - ✅ Added null check before calling `close()`
3. **String validation with trim()** - ✅ Updated all ID validations to use `trim().length === 0` check and updated error messages to say "non-empty string"

## Not Applied (with rationale)

**Reusable validation helper** - While code deduplication is good, creating a helper method for this simple validation would be over-engineering for this PR scope. The current inline validation is clear and maintainable.

**Stricter content validation** - The `sendMessage()` method accepts various content types (string, MessageMedia, Location, Poll, Contact, etc.). Adding type-specific validation would require significant refactoring and could break existing use cases. The current validation just ensures content is not null/undefined.

**Test coverage** - Adding comprehensive tests for these validations is beyond the scope of this bug fix PR. Tests would be welcome as a follow-up contribution.

**Participants validation in createGroup** - The method already handles various participant formats (Contact objects, strings, single values, arrays). Adding stricter validation could break existing usage patterns.

PR #5197, opened 2025-12-12
fix: prevent destroy() from hanging when browser disconnected

### LINKED ISSUE #3846: Client.destroy may hang after Browser disconnected, preventing cleanup (pupBrowser.close() never resolves/rejects)

gn00295120:
## Summary

When the browser process has already been terminated or disconnected (e.g., crashed, killed externally, or connection lost), calling `destroy()` would hang indefinitely because `pupBrowser.close()` never resolves.

This fix adds a check using `isConnected()` before attempting to close the browser, preventing the hang.

## Changes

- Added `isConnected()` check in `destroy()` method before calling `close()`

## Issue

Fixes #3846

## Test

The fix prevents `destroy()` from hanging when:
1. Browser process was killed externally
2. Browser crashed
3. Connection to browser was lost

Before fix:
```javascript
await client.destroy(); // Hangs indefinitely if browser disconnected
```

After fix:
```javascript
await client.destroy(); // Returns immediately if browser not connected
```

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -846,7 +846,9 @@ class Client extends EventEmitter {
      * Closes the client
      */
     async destroy() {
-        await this.pupBrowser.close();
+        if (this.pupBrowser?.isConnected()) {
+            await this.pupBrowser.close();
+        }
         await this.authStrategy.destroy();
     }
 


gn00295120:
Thanks for the review!

Regarding the test coverage suggestion: Adding integration tests that simulate a disconnected browser would require significant test infrastructure setup (mocking Puppeteer's browser lifecycle, managing process termination, etc.) and is beyond the scope of this minimal bug fix.

The fix itself is straightforward - it uses Puppeteer's built-in `isConnected()` method which is well-tested. The optional chaining (`?.`) also handles the case where `pupBrowser` is null/undefined.

Tests for this scenario would be welcome as a follow-up contribution.

PR #5675, opened 2025-12-16
general: tree-shaking pre-MD and old comet versions

purpshell:
# PR Details
Removes pre-MD stuff and also some expired version stuff for the library. 

# Sponsors
Big thanks to my sponsors at https://purpshell.dev/sponsor for supporting this work. It is a pleasure to return to work on this project again and I wouldn't be able to do it without them.


lindionez:
Seems like a good removal of unnecessary code, will this be added to the official branch?

tuyuribr:
@purpshell I think we should make separate (like one pr for each 2 files) pr's to reduce the review workload

purpshell:
> @purpshell I think we should make separate (like one pr for each 2 files) pr's to reduce the review workload



> @purpshell I think we should make separate (like one pr for each 2 files) pr's to reduce the review workload

I will work on it but the workload is not large at all. We used to have PRs with 20 and 30 changed files like those that revamped wwebjs for MD. 

I would like to keep up this trend of code cleaning making more of these micro PRs to ease the code clutter and increase DX.

BenyFilho (approved):

package.json
```diff
@@ -29,7 +29,6 @@
   },
   "homepage": "https://wwebjs.dev/",
   "dependencies": {
-    "@pedroslopez/moduleraid": "^5.0.2",
```
tuyuribr, line 32: The end of an era

tuyuribr (approved):

lindionez:
image
After this Pr was entered, this error prevented the client from starting.

ricardoapaes:
@lindionez Same problem, were you able to solve it?

lindionez:
> @lindionez Same problem, were you able to solve it?

I'm using my fork of this lib in my project, what I did was not pull this pr there. You can use it temporarily if you want, it's on my profile.

BenyFilho:
@lindionez @ricardoapaes  Please, try it https://github.com/pedroslopez/whatsapp-web.js/pull/127048

PR #5713, opened 2026-01-05
chore: fix release script, use OIDC

pedroslopez:


tools/publish
```diff
@@ -78,10 +83,7 @@ git add .
 
 git commit -am "chore(${NEW_VERSION}): generate docs"
```
mracharyaboy-collab: #9187584694

PR #5738, opened 2026-01-20
fix: ready event not fire on some accounts |  guard missing WAWebSetPushnameConnAction during Store injection

yosofbayan:
Problem: ExposeStore can throw Requiring unknown module "WAWebSetPushnameConnAction" in certain WhatsApp Web sessions, notably during reconnection/restore flows when using the LocalAuth strategy. This aborts Store injection, so window.WWebJS is never initialized and the ready event does not fire. As a result, downstream calls that rely on the injected API (e.g., Client.getChatById) can fail, leading to errors such as ERROR (sendMessage) Evaluation failed: TypeError: Cannot read properties of undefined (reading 'getChat'), which also breaks operations like sendMessage. In practice, the issue often appeared “fixed” only after forcing a fresh login by clearing the saved session (removing the cached auth) and re-scanning the QR code.

Fix: Guard the WAWebSetPushnameConnAction require with try/catch, and only attach Settings.setPushname when the module is available and exposes a function.

Impact: Prevents a hard initialization failure while preserving existing behavior when the module exists. This makes startup more resilient across varying WhatsApp Web builds without changing functional paths for environments where WAWebSetPushnameConnAction is present.

How tested: Reproduced the failure on a session where the module was missing (authenticated fires, but ready does not; window.WWebJS is absent; getChatById fails), specifically during a LocalAuth reconnection attempt. After the patch, ready consistently fires, window.WWebJS is present, getChatById succeeds, and sendMessage works as expected without requiring session deletion or QR re-scan.


fisal1993:
are u able to share some code example so ppls can benefit from the solution and thanks for your work much appriciated.

MattShepphard:
As a temporary solution in my project, I implemented a timeout-based safeguard for the ready event.

If the ready event is not emitted within a predefined time window after initialization, the client is reinitialized. This process is repeated until the ready event is successfully triggered.

fisal1993:
> As a temporary solution in my project, I implemented a timeout-based safeguard for the ready event.
> 
> If the ready event is not emitted within a predefined time window after initialization, the client is reinitialized. This process is repeated until the ready event is successfully triggered.

after that sending messages etc work without any error?

ss-asia:
I am using this fork and it seems to work though im not sure if its because of the fork or because i re-initialized with the QR again

MattShepphard:
> > As a temporary solution in my project, I implemented a timeout-based safeguard for the ready event.
> > If the ready event is not emitted within a predefined time window after initialization, the client is reinitialized. This process is repeated until the ready event is successfully triggered.
> 
> after that sending messages etc work without any error?

Yes. if Ready event is emitted, sending messages correctly

Maleriandro:
This fork still doesnt work for me.

romiardisaja:
Doesnt work for me to.
Windows 10_x64
v1.34.5-alpha.3

joaosn:
https://github.com/pedroslopez/whatsapp-web.js/issues/5717#issuecomment-3808421207

yosofbayan:
@BenyFilho 
can you check this ?

BenyFilho:
> @BenyFilho can you check this ?

We’re waiting for more tests, as this is a very specific situation and isn’t reproducible for everyone.
Another point: your code is designed to silently ignore errors when loading the module. However, if the user wants to use setPushName, this behavior can hide the real problem.

N0N4M3BNS:
From my side, when using the MAIN branch, it gets stuck at AUTHENTICATED. When using this patch, I can bypass it. I'm not using setPushName on my side, so I'm not testing it.

rprimeathena:
✅ Your fix works perfectly!

I was stuck at AUTHENTICATED with the official version, and the bot wasn't receiving any messages despite being connected. After installing your fork:

bash
npm install git+https://github.com/yosofbayan/whatsapp-web.js.git#fix/wwebjs-not-initialized
✅ Ready event fires correctly
✅ Messages are received and processed
✅ Bot is fully operational again
This is a critical fix for the current WhatsApp Web changes. Thank you for your work on this! 🙏

Environment:

Node.js 20.x
Ubuntu VPS
LocalAuth strategy
WhatsApp Web version: 2.3000.1032759465

ss-asia:
For me it worked with the fork but then it stopped working later when i reinitialized.
But i need to dive into it again if it works properly but need to check good before merge 4 sure

src/util/Injected/Store.js
```diff
@@ -234,7 +249,7 @@ exports.ExposeStore = () => {
     window.injectToFunction = (target, callback) => {
         try {
             let module = window.require(target.module);
-            if (!module) return; 
+            if (!module) return;
```
purpshell, line 252: shouldn't this line be the only thing that matters?

purpshell:
closing in favor of #3972

src/util/Injected/Store.js
yosofbayan: > shouldn't this line be the only thing that matters?

it's just a code formatting here

PR #5749, opened 2026-01-25
Refactoring for Robustness

heavens7above:
# PR Details

### Environment

- Machine OS: 
- Phone OS: 
- Library Version: 
- WhatsApp Web Version: 
- Puppeteer Version:
- Browser Type and Version: 
- Node Version: 

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/AuthStore/LegacyAuthStore.js
+++ b/src/util/Injected/AuthStore/LegacyAuthStore.js
@@ -6,17 +6,35 @@ exports.ExposeLegacyAuthStore = (moduleRaidStr) => {
     eval('var moduleRaid = ' + moduleRaidStr);
     // eslint-disable-next-line no-undef
     window.mR = moduleRaid();
+    
+    // Validation helper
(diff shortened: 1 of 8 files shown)


src/util/Injected/AuthStore/LegacyAuthStore.js
```diff
@@ -6,17 +6,35 @@ exports.ExposeLegacyAuthStore = (moduleRaidStr) => {
     eval('var moduleRaid = ' + moduleRaidStr);
     // eslint-disable-next-line no-undef
     window.mR = moduleRaid();
+    
+    // Validation helper
+    const requireModule = (selector, exportName = null) => {
```
devlikepro, line 11: It's a good idea, but I think it should be optional, like using the `strictRequireModule` flag in the client.

This is a good solution for a development environment, where you can quickly see which modules are missing, but it's bad for production - if you only send text, the entire instance will work, but with this code change, it will **stop working if even a very small, rarely used module/function** is moved elsewhere.

PR #5775, opened 2026-01-29
fix: update call handling for recent WhatsApp Web changes

alessandrocampos224:
Updated the call handling logic to accommodate the renaming of Store.Call to WAWebCallCollection in recent WhatsApp Web versions. This ensures that incoming calls are properly managed without breaking functionality.

# PR Details

## Description

Updated the call handling logic to accommodate the renaming of Store.Call to WAWebCallCollection in recent WhatsApp Web versions. This ensures that incoming calls are properly managed without breaking functionality.

When Store.Call is undefined, the previous code threw "Cannot read properties of undefined (reading 'on')", which prevented the rest of attachEventListeners from running (including Store.Msg.on('add', ...)), so the client stopped receiving messages. The fallback fixes both call handling and message reception.

## Motivation and Context

In recent WhatsApp Web builds, Store.Call was renamed to WAWebCallCollection. The code assumed Store.Call always existed and called Store.Call.on('add', ...), which threw when Store.Call was undefined. That exception aborted the entire page.evaluate(), so Store.Msg.on('add', ...) was never registered and the client never received messages. This change uses Store.Call when present, otherwise WAWebCallCollection, and only registers if the collection exists and has an .on method.

## How Has This Been Tested

Manual testing: ran the client with the fix, linked via QR, confirmed READY fires and message_create events are received (e.g. !ping responds). Verified in browser DevTools that window.Store has WAWebCallCollection and no longer has Call.

### Environment

Machine OS: WSL2 (Linux)
Phone OS: (e.g. Android)
Library Version: (github:pedroslopez/whatsapp-web.js#auto-wa-web-update/patch)
WhatsApp Web Version: '2.3000.1032614117'
Puppeteer Version: 24.36.1 (puppeteer-core)
Browser Type and Version: Chromium (Puppeteer)
Node Version: v20.19.0

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -756,7 +756,6 @@ class Client extends EventEmitter {
             window.Store.Msg.on('change:body change:caption', (msg, newBody, prevBody) => { window.onEditMessageEvent(window.WWebJS.getMessageModel(msg), newBody, prevBody); });
             window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); });
             window.Store.Conn.on('change:battery', (state) => { window.onBatteryStateChangedEvent(state); });
-            // WhatsApp Web renomeou Store.Call para WAWebCallCollection em versões recentes
             const callCollection = (window.Store && window.Store.Call) || (window.Store && window.Store.WAWebCallCollection);
             if (callCollection && typeof callCollection.on === 'function') {
                 callCollection.on('add', (call) => { window.onIncomingCall(call); });
--- a/src/Client.js
+++ b/src/Client.js
@@ -756,7 +756,11 @@ class Client extends EventEmitter {
             window.Store.Msg.on('change:body change:caption', (msg, newBody, prevBody) => { window.onEditMessageEvent(window.WWebJS.getMessageModel(msg), newBody, prevBody); });
             window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); });
             window.Store.Conn.on('change:battery', (state) => { window.onBatteryStateChangedEvent(state); });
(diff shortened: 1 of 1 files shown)


ZacksTsang:
nice

susuks:
cara, tu salvou meu dia. tô sofrendo com o problema o dia inteiro.

BenyFilho (approved):
That's the correct fix

ricardoapaes:
Work. Thanks

BenyFilho (approved):

BenyFilho (approved):

v-mwalk:
This resolved the issue with 'ready' not happening.  But getChat bombing out.  This could be on a separate PR

src/Client.js
```diff
@@ -756,7 +756,11 @@ class Client extends EventEmitter {
             window.Store.Msg.on('change:body change:caption', (msg, newBody, prevBody) => { window.onEditMessageEvent(window.WWebJS.getMessageModel(msg), newBody, prevBody); });
             window.Store.AppState.on('change:state', (_AppState, state) => { window.onAppStateChangedEvent(state); });
             window.Store.Conn.on('change:battery', (state) => { window.onBatteryStateChangedEvent(state); });
-            window.Store.Call.on('add', (call) => { window.onIncomingCall(call); });
+            // WhatsApp Web renomeou Store.Call para WAWebCallCollection em versões recentes
```
tuyuribr: Remove this line

tuyuribr (changes requested):

src/Client.js
BenyFilho: Yuri, WhatsApp change the module to  window.Store.WAWebCallCollection, so his solution will work

tuyuribr (approved):

PR #5789, opened 2026-01-30
Added support for getting the current user's display name

wouter98:
# PR Details

Added a way to get the display name of the current user.

## Description

Added a method to the client class to get the display name of the current user.

## Motivation and Context

I wanted to create a program that works with chat exports. They use the displayname of the user that created the export as the author for those messages. This change makes it much easier to work with those exports, as you can get the displayname directly from the client rather than through messages.

## How Has This Been Tested

Linked the updated package to my source code, and was able to get the username properly.

### Environment

- Machine OS: Debian 12
- Phone OS: -
- Library Version: 1.34.6
- WhatsApp Web Version: 2.3000.1032721183
- Puppeteer Version: 24.36.1
- Browser Type and Version: -
- Node Version: v22.18.0

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -202,6 +202,11 @@ declare namespace WAWebJS {
          */
         setStatus(status: string): Promise<void>
 
+        /**
+         * Gets the current user's display name
+         */
+        getDisplayName(): Promise<string>
+
         /** 
          * Sets the current user's display name
          * @param displayName New display name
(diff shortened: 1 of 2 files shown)


BenyFilho:
That already exists and can be accessed via client.info.pushname

wouter98:
Wow, not sure how I missed that. Guess there's no need for this pr then.

Issue #5788, opened 2026-01-30
[Bug] TypeError: "i.getLastMsgKeyForAction is not a function" when sending to Newsletter/Channel


nwillian:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this a problem caused by your code, or is it specifically because of the library?

- [x] I have double-checked my code carefully.

### Describe the bug.

When attempting to send a message (text or media) to a WhatsApp Channel (ID ending in `@newsletter`) using the latest WhatsApp Web version, the client crashes with a fatal error inside the injected script.

The error specifically points to `getLastMsgKeyForAction`, suggesting the library is trying to access a function that either doesn't exist or is different for Newsletter Chat objects in the current Webpack module structure.

**Steps to Reproduce**
1. Initialize the client.
2. Attempt to send a message to a Newsletter ID **without** disabling `sendSeen`:
   ```javascript
   // Triggers the error
   await client.sendMessage('120363421265788757@newsletter', 'Test Message');

### Expected Behavior

The message should be sent to the Newsletter/Channel successfully without throwing a fatal TypeError.

Ideally, the library should automatically detect that the target is a Newsletter/Channel and skip the "mark as read" (sendSeen) logic if the internal function `getLastMsgKeyForAction` is not available for this chat type, ensuring backward compatibility.

### Steps to Reproduce the Bug or Issue

1. Initialize the client using LocalAuth.
2. Wait for the client to be ready.
3. Attempt to send a message to a Newsletter ID (ending in `@newsletter`) using the standard sendMessage method without options:
   `await client.sendMessage('120363422265788756@newsletter', 'Test message');`
4. Observe that the client crashes immediately with "TypeError: i.getLastMsgKeyForAction is not a function".

### WhatsApp Account Type

WhatsApp Business

### Browser Type

Chrome (via Puppeteer)

### Operation System Type

Windows 11 (x64)

### Phone OS Type

iOS

### WhatsApp-Web.js Version

versão atual

### WhatsApp Web Version

Latest (2.3000.1012xxx)

### Node.js Version

v20.11.0

### Authentication Strategy

LocalAuth

### Additional Context

_No response_

deleteduser98:
I’m encountering the same issue.

bagusandrewijaya:
i have same issues

bagusandrewijaya:
> i have same issues ISSUE FIXED FOR ME

[...]

PR #5790, opened 2026-01-31
fixed error i.getLastMsgKeyForAction

bagusandrewijaya:
# WhatsApp Web.js Newsletter Fix

## The Problem

I was getting this error when trying to send messages to WhatsApp Newsletters: `TypeError: i.getLastMsgKeyForAction is not a function`.

## What Was Happening

The library was trying to mark newsletter chats as "read" after sending messages (that's the sendSeen feature), but newsletter chats don't have that `getLastMsgKeyForAction` method that regular chats have. So it would crash every time.

## My Fix

I modified the `sendMessage` function in `Client.js` to automatically skip the sendSeen feature for newsletter chats. Now it detects if the chat ID ends with `@newsletter` and disables sendSeen just for those chats.

## Code Changes

In `src/Client.js`:
```javascript
// Before
const sendSeen = options.sendSeen !== false;

// After  
const sendSeen = (options.sendSeen !== false) && !isChannel;
```

Where `isChannel` checks if the chat ID contains `@newsletter`.

I also added a safety check in the `sendSeen` function itself in case someone calls it directly on a newsletter chat.

## Result

- ✅ Newsletter messages now work without crashing
- ✅ Regular chats still work normally
- ✅ No breaking changes to existing code

That's the main fix I implemented for the newsletter messaging issue.</content>

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -981,7 +981,7 @@ class Client extends EventEmitter {
      * @returns {Promise<Message>} Message that was just sent
      */
     async sendMessage(chatId, content, options = {}) {
-        const isChannel = /@\w*newsletter\b/.test(chatId);
+        const isChannel = /@newsletter\b/.test(chatId);
         const isStatus = /@\w*broadcast\b/.test(chatId);
 
         if (isChannel && [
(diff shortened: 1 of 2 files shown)


bagusandrewijaya:
the issues right here #5788   I tried to do the fix on my local system and it worked

BenyFilho:
Thank you for your initiative, but that is not the correct fix to this issue.
The correct one is that
https://github.com/pedroslopez/whatsapp-web.js/pull/5791

PR #5816, opened 2026-02-08
fix(getChat): resolve LID for unknown contacts via WAWebContactSyncUtils

jeremyandes:
When sendMessage is called for a contact not yet synced, findOrCreateLatestChat fails with "No LID for user". This adds a third fallback in getChat that uses WAWebContactSyncUtils.constructUsyncDeltaQuery to resolve the phone number to a LID via WhatsApp servers, then retries findOrCreateLatestChat with the resolved LID.

The fix preserves the existing behavior:
- 1st: fast in-memory cache lookup via Store.Chat.get()
- 2nd: async findOrCreateLatestChat with the original WID
- 3rd (new): LID resolution + retry (only when both above fail)

# PR Details

## Description

`window.WWebJS.getChat` fails for contacts whose LID hasn't been resolved yet, causing `sendMessage` (and any other caller) to throw "No LID for user".

This patch adds a third fallback step to the existing two-step chat resolution in `getChat`:

1. **Cache lookup** (`Store.Chat.get`) — unchanged, fast synchronous check.
2. **Async find/create** (`FindOrCreateLatestChat.findOrCreateLatestChat(chatWid)`) — unchanged, now wrapped with `.catch(() => null)` to gracefully handle failures.
3. **LID resolution (new)** — only runs when both above return nothing. Uses `WAWebContactSyncUtils.constructUsyncDeltaQuery` to sync the phone number with WhatsApp servers, obtains the LID, then retries `findOrCreateLatestChat` with the resolved LID.

Key design decisions:
- The fast-path cache lookup is preserved — no performance regression for already-loaded chats.
- On failure, `chat` remains `undefined` (not `null`), maintaining compatibility with callers like `sendClearChat` and `sendDeleteChat` that check `!== undefined`.
- The entire LID resolution block is wrapped in `try/catch`, so if `WAWebContactSyncUtils` doesn't exist in a given WhatsApp Web version, it fails silently.

## Related Issue(s)

closes #3834

## Motivation and Context

WhatsApp has been migrating to LID-based contact identification. When `sendMessage` is called for a number that hasn't been synced/resolved to a LID yet, `findOrCreateLatestChat` throws "No LID for user" and the message fails to send. This is a common scenario when messaging new contacts programmatically (e.g., via chatbots or CRM integrations).

## How Has This Been Tested

- Tested by sending a message via `client.sendMessage()` to a phone number not previously in the contact list (the exact scenario that triggers "No LID for user").
- Verified that existing chats (already in cache) still resolve instantly via the first fallback.
- Verified that the function returns `undefined` (not `null`) when all three fallbacks fail, preserving compatibility with downstream callers.

### Environment

- Machine OS: Mac
- Phone OS: Android
- Library Version: 1.34.6
- WhatsApp Web Version: 2.3000+
- Puppeteer Version: ^24.31.0
- Browser Type and Version: Chromium (bundled with Puppeteer)
- Node Version: v20.19.4

## Types of changes

- [ ] Dependency change

## Checklist

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -581,7 +581,25 @@ exports.LoadUtils = () => {
                 chat = null;
             }
         } else {
-            chat = window.Store.Chat.get(chatWid) || (await window.Store.FindOrCreateChat.findOrCreateLatestChat(chatWid))?.chat;
+            chat = window.Store.Chat.get(chatWid);
+            if (!chat) {
+                chat = (await window.Store.FindOrCreateChat.findOrCreateLatestChat(chatWid).catch(() => null))?.chat;
+            }
+            if (!chat) {
+                try {
+                    const query = window.require('WAWebContactSyncUtils').constructUsyncDeltaQuery([{
+                        type: 'add',
+                        phoneNumber: chatWid.user
+                    }]);
+                    const result = await query.execute();
+                    if (result?.list?.[0]?.lid) {
+                        const chatLid = window.Store.WidFactory.createWid(result.list[0].lid);
+                        chat = (await window.Store.FindOrCreateChat.findOrCreateLatestChat(chatLid).catch(() => null))?.chat;
+                    }
+                } catch (e) {
+                    // LID resolution failed, chat remains undefined
+                }
+            }
         }
 
         return getAsModel && chat


talgatenergysphere:
Is it possible to somehow optimize the call to `window.require('WAWebContactSyncUtils')` via `whatsapp-web.js\src\util\Injected\Store.js`?

BenyFilho:
My suggestion to make this a more robust solution:

1. Add the module to the Store , window.Store.ContactMethods would be a good place.
2. Create a utility method to return the sync result, since it can be reused by other methods.
3. Use this new utility method in addParticipants and getChat.

izofis (approved):
I am currently using multiple WhatsApp accounts live. I haven't encountered any problems.

DiegoFelipe:
I just updated the package but still get the same lid error

jeremyandes:
> Is it possible to somehow optimize the call to `window.require('WAWebContactSyncUtils')` via `whatsapp-web.js\src\util\Injected\Store.js`?

Feel free to keep coding! No problem if you push commits



> My suggestion to make this a more robust solution:
> 
> 1. Add the module to the Store , window.Store.ContactMethods would be a good place.
> 2. Create a utility method to return the sync result, since it can be reused by other methods.
> 3. Use this new utility method in addParticipants and getChat.

Nice suggestion!

> I am currently using multiple WhatsApp accounts live. I haven't encountered any problems.

Amazing and glad that this works for you 🙌🏻

> I just updated the package but still get the same lid error

This is not a release yet, did you run this branch locally ?

ricardorodh:
this error in console:
ERROR (TypeError): Cannot read properties of undefined (reading 'commonGid')

N0N4M3BNS:
Working on side, TQ

UPDATED: Not all groups work; some groups can trigger sendMessage, but messages are not coming in on WhatsApp, and no return data is available. 

@jeremyandes

DiegoFelipe:
@BenyFilho you closed both PRs

BenyFilho:
> @BenyFilho you closed both PRs

Yeah, PRs do not comply with recent changes

PR #22449, opened 2026-02-10
feat: add cancelPairingCode() and requestPairingCode() on QR clients

### LINKED ISSUE #22390: feat: allow switching between QR code and pairing code without destroying the client
## Is your feature request related to a problem? Please describe.

When building an app that offers both QR code scanning and phone number pairing as authentication methods, there is no way to switch between them on the same client instance. If a user starts with QR code and wants to switch to pairing code (or vice versa), you have to destroy the client and create a new one.

This is a problem because:
- Destroying and recreating the client is slow (Puppeteer has to relaunch Chrome)
- It wastes resources and creates a poor user experience
- The underlying WhatsApp Web page supports both methods on the same session

## Describe the solution you'd like.

Two changes:

1. **`requestPairingCode()` should work on clients initialized in QR mode.** Currently, calling `requestPairingCode()` on a QR-initialized client fails because `onCodeReceivedEvent` is not exposed during `initialize()`. Both `onQRChangedEvent` and `onCodeReceivedEvent` should always be registered, regardless of the initial authentication mode.

2. **Add a `cancelPairingCode()` method** that cancels an active pairing code session and returns to QR code mode. This would clear the pairing code refresh interval and trigger a QR refresh so the `qr` event fires again.

## Describe an alternate solution if you have one.

The current workaround is to destroy the client and recreate it with `pairWithPhoneNumber` option (or without it to go back to QR). This works but is wasteful and slow.

## Additional Context

I have a working implementation in a PR that I will link to this issue. The changes are minimal and backward compatible:
- Always register both QR and code handlers in `initialize()`
- Move the `exposeFunctionIfAbsent` for `onCodeReceivedEvent` from the `pairWithPhoneNumber` branch to always execute
- Add `cancelPairingCode()` that clears `window.codeInterval` and calls `Store.Cmd.refreshQR()`
- Update `onAuthAppStateChangedEvent` to skip QR refresh when a pairing code interval is active

Adi1231234:
## Summary

Add `cancelPairingCode()` method and allow `requestPairingCode()` to work on QR-initialized clients (not just clients initialized with `pairWithPhoneNumber` option).

### `requestPairingCode()` - allow on QR-initialized clients
When called on a client that was initialized in QR mode, `onCodeReceivedEvent` may not be exposed yet (it is only set up in `initialize()` when `pairWithPhoneNumber` is configured). Added `exposeFunctionIfAbsent` call at the beginning of `requestPairingCode()` to ensure the callback is available regardless of initialization mode.

### `cancelPairingCode()` - new method
Cancels an active pairing code session and returns to QR code mode:
1. Clears the `window.codeInterval` refresh timer
2. Calls `PairingCodeLinkUtils.initializeQRLinking()` to switch WhatsApp Web back to QR mode
3. Re-emits the current QR code via `onQRChangedEvent` using the cached `getQR` helper

### `window.getQR` - cached QR builder
The `getQR` function (which builds the full QR string from registration keys) is now saved on `window` during `initialize()`. This is necessary because:
- `Store.Cmd.refreshQR()` is no longer available in current WhatsApp Web versions
- `getADVSecretKey()` returns null after `setPairingType("ALT_DEVICE_LINKING")` is called
- The `change:ref` listener registered in `initialize()` already uses this closure, so caching it on `window` allows `cancelPairingCode()` to reuse the same keys

### TypeScript definition
Added `cancelPairingCode(): Promise<void>` to `index.d.ts`.

## Use case

Allows switching between QR code and pairing code authentication **without destroying and recreating the client**. This enables UIs where users can toggle between QR scanning and phone number pairing on the same screen.

```js
const client = new Client({ /* QR mode (default) */ });
client.initialize();

// User wants to switch to pairing code
client.on("qr", () => { /* show QR */ });
const code = await client.requestPairingCode("1234567890");
// show pairing code to user

// User wants to switch back to QR
await client.cancelPairingCode();
// qr event fires again with fresh QR
```

## Testing

Tested full round-trip flow via E2E tests:
1. QR displayed -> switch to pairing code (received valid code)
2. Cancel pairing -> QR returns with valid scannable code
3. Switch to pairing again -> new code received
4. Cancel again -> QR returns

All transitions work without destroying the client.

Closes #22390

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -197,6 +197,9 @@ declare namespace WAWebJS {
             intervalMs?: number,
         ): Promise<string>;
 
+        /** Cancels an active pairing code session and returns to QR code mode */
+        cancelPairingCode(): Promise<void>
+
         /** Force reset of connection state for the client */
         resetState(): Promise<void>;
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -269,6 +269,7 @@ class Client extends EventEmitter {
                         advSecretKey +
                         ',' +
                         platform;
+                    window.getQR = getQR;
 
                     window.onQRChangedEvent(getQR(window.AuthStore.Conn.ref)); // initial qr
                     window.AuthStore.Conn.on('change:ref', (_, ref) => {
@@ -516,6 +517,14 @@ class Client extends EventEmitter {
         showNotification = true,
         intervalMs = 180000,
     ) {
+        await exposeFunctionIfAbsent(
+            this.pupPage,
+            'onCodeReceivedEvent',
+            async (code) => {
+                this.emit(Events.CODE_RECEIVED, code);
+                return code;
+            },
+        );
         return await this.pupPage.evaluate(
             async (phoneNumber, showNotification, intervalMs) => {
                 const getCode = async () => {
@@ -551,6 +560,20 @@ class Client extends EventEmitter {
             showNotification,
             intervalMs,
         );
+    }
+
+    /**
+     * Cancels an active pairing code session and returns to QR code mode
+     */
+    async cancelPairingCode() {
+        await this.pupPage.evaluate(() => {
+            if (window.codeInterval) {
+                clearInterval(window.codeInterval);
+                window.codeInterval = undefined;
+            }
+            window.AuthStore.PairingCodeLinkUtils.initializeQRLinking();
+            window.onQRChangedEvent(window.getQR(window.AuthStore.Conn.ref));
+        });
     }
 
     /**
--- a/src/Client.js
+++ b/src/Client.js
@@ -255,8 +255,6 @@ class Client extends EventEmitter {
                         window.AuthStore.Base64Tools.encodeB64(
                             registrationInfo.identityKeyPair.pubKey,
                         );
-                    const advSecretKey =
-                        await window.AuthStore.RegistrationUtils.getADVSecretKey();
                     const platform =
                         window.AuthStore.RegistrationUtils.DEVICE_PLATFORM;
                     const getQR = (ref) =>
@@ -266,11 +264,11 @@ class Client extends EventEmitter {
                         ',' +
                         identityKeyB64 +
                         ',' +
-                        advSecretKey +
+                        window
+                            .require('WAWebUserPrefsMultiDevice')
+                            .getADVSecretKey() +
                         ',' +
                         platform;
-                    window.getQR = getQR;
-
                     window.onQRChangedEvent(getQR(window.AuthStore.Conn.ref)); // initial qr
                     window.AuthStore.Conn.on('change:ref', (_, ref) => {
                         window.onQRChangedEvent(getQR(ref));
@@ -566,13 +564,15 @@ class Client extends EventEmitter {
      * Cancels an active pairing code session and returns to QR code mode
      */
     async cancelPairingCode() {
-        await this.pupPage.evaluate(() => {
+        await this.pupPage.evaluate(async () => {
             if (window.codeInterval) {
                 clearInterval(window.codeInterval);
                 window.codeInterval = undefined;
             }
-            window.AuthStore.PairingCodeLinkUtils.initializeQRLinking();
-            window.onQRChangedEvent(window.getQR(window.AuthStore.Conn.ref));
+            window.require('WAWebLaunchSocketUtils').refreshQR();
+            await window
+                .require('WAWebAltDeviceLinkingApi')
+                .initializeQRLinking();
         });
     }
 


Adi1231234:
Hey @purpshell @BenyFilho, just checking in on this PR 🙂

It adds cancelPairingCode() and lets you call requestPairingCode() on clients that were initialized with QR. Pretty useful for apps that want to let users switch between QR and phone number.

Any chance you could take a look? Happy to make changes or jump on Discord if needed!

BenyFilho:
> Hey @purpshell @BenyFilho, just checking in on this PR 🙂
> 
> It adds cancelPairingCode() and lets you call requestPairingCode() on clients that were initialized with QR. Pretty useful for apps that want to let users switch between QR and phone number.
> 
> Any chance you could take a look? Happy to make changes or jump on Discord if needed!

Need user tests to validate it, and update your changes with branch main

Adi1231234:
Hey @BenyFilho, just rebased onto the latest main so it's up to date and CI is green.

I keep running into conflicts because other PRs get merged while mine stay open, and I have to keep rebasing them. I'd really love to get these merged so I can stop chasing conflicts 😅

You mentioned user tests, could you clarify what exactly you're looking for? I'd be happy to add whatever is needed to move this forward.

BenyFilho:
> Need user tests to validate it, and update your changes with branch main

Need user tests to validate it, and update your changes with branch main

Adi1231234:
Hey @BenyFilho, thanks for looking at this!

Rebased onto latest main and CI is green.

This feature lets users switch between QR and pairing code without destroying the client, which is pretty useful for apps with a linking UI. I've been using it in my own app and tested the full round trip (QR -> pairing code -> cancel -> QR again) multiple times.

If anyone here needs this functionality or wants to give it a spin, I'd love your feedback. I really need my open PRs merged so if anyone can help review or test them, I'm happy to pay for your time. Just ping me!

BenyFilho:
> Hey @BenyFilho, thanks for looking at this!
> 
> Rebased onto latest main and CI is green.
> 
> This feature lets users switch between QR and pairing code without destroying the client, which is pretty useful for apps with a linking UI. I've been using it in my own app and tested the full round trip (QR -> pairing code -> cancel -> QR again) multiple times.
> 
> If anyone here needs this functionality or wants to give it a spin, I'd love your feedback. I really need my open PRs merged so if anyone can help review or test them, **I'm happy to pay for your time**. Just ping me!

Hi,

This is a community project, so changes need to be reviewed and validated by other users. They should be useful for the community in general and must not break existing functionality.

I understand that these changes are important for your use case, but they also need to make sense for the community as a whole. The approval decision is based on technical and community impact, not on financial aspects.

Thanks for your understanding.

Adi1231234:
Hey @BenyFilho, just want to clarify - I absolutely did not mean paying for approvals. I meant compensating someone for their time reviewing and discussing the best approach, even if the conclusion is that my implementation needs rework.

I share the same philosophy I've seen you promote - if you have a problem, fix it yourself. That's why I've put in the work and opened multiple PRs. Unfortunately it's been hard to get people to test and review them.

I truly believe in this project and would rather contribute back than maintain my own fork with all the fixes. So yes, I'm genuinely willing to pay for help and support to move these fixes forward - for the benefit of the whole community.

Thanks! 🙏

sofi-ans:
@Adi1231234 Cool idea, switching between QR and pairing code without recreating the client would be really useful.

I spent some time looking at the WA Web internals to understand the flow, and I think cancelPairingCode() has a problem that would make the QR it emits unscannable.

The issue is with how window.getQR works. It's a closure created during initialize() that captures the ADV secret key once. But by the time cancelPairingCode() runs, that key is no longer valid - initializeAltDeviceLinking() (called during requestPairingCode) deletes it from IndexedDB, and then initializeQRLinking() (called during cancel) generates a completely new random one. So the closure still has the original key, but IndexedDB has a different one.

This actually causes a logout, not just a broken QR. When a phone scans the code, the server uses the key from the QR string to compute an HMAC, and the web client validates by reading the current key from IndexedDB. Since they don't match, logoutAfterValidationFail() kicks in. I looked at how WA Web's own QR component handles this - it reads the key fresh from IndexedDB on every render instead of caching it, which avoids this entirely.

A couple more things I noticed:

initializeQRLinking() is async (it writes to IndexedDB) but it's called without await, so the QR gets emitted before the state reset is done.

Also, WA Web's internal resetLinkDeviceState always calls require('WAWebLaunchSocketUtils').refreshQR() before switching modes - this restarts the WebSocket and gets fresh refs from the server. Without it Conn.ref is probably expired since refs only live for ~60 seconds. You're right that Store.Cmd.refreshQR() is gone, but WAWebLaunchSocketUtils.refreshQR() is still there and that's what the internal code uses.

The exposeFunctionIfAbsent part in requestPairingCode looks good though, that makes sense.

BenyFilho (approved):

Adi1231234:
@BenyFilho Thanks!

@sofi-ans Good catch, you were right. I tested it and the QR scan was actually failing. Fixed it - getQR now reads the key fresh instead of using the cached one, and cancelPairingCode follows the same flow WA Web uses internally (refreshQR + initializeQRLinking). Works now.

sofi-ans:
@Adi1231234 @BenyFilho Works good 👍

sofi-ans (approved):

aliyss (approved):

PR #127079, opened 2026-02-25
feat: add ciphertext message retry/recovery mechanism (PDO type 4)

### LINKED ISSUE #127062: Feature: Add ciphertext message retry/recovery mechanism (PDO type 4)
### Is the feature you are looking for already implemented in the library?

- [X] I have reviewed the implemented features.

### Is your "solution" already implemented in an opened pull request?

- [X] I have reviewed the existing pull requests.

Messages arriving as `type: "ciphertext"` frequently **never decrypt**. The `change:type` listener never fires, and the message is silently lost.

**Root cause:** Broken Signal Protocol sessions between the companion device and certain senders:

1. **Missing Sender Key Distribution Messages (groups):** Companion misses the SKDM → can't decrypt any subsequent messages from that sender in that group.
2. **Broken pairwise sessions (DMs):** Sender changes device/reinstalls → companion's session is stale, renegotiation fails silently.

**Evidence:**
- Ciphertext messages remain unchanged in `Store.Msg` after 120+ seconds
- `change:type` genuinely never fires — no model replacement, `isNewMsg` stays `true`
- Failures are sender-specific (`@lid` addresses) and correlate with `e2e_notification` events
- `getMessagesById()` also returns the same unresolved ciphertext

**Current code (Client.js) — no timeout, no retry, no fallback:**
```js
if (msg.type === 'ciphertext') {
    msg.once('change:type', (_msg) => {
        onAddMessageEvent(getMessageModel(_msg));
    });
}
```

**Add PDO type 4 retry**, like [Baileys](https://github.com/WhiskeySockets/Baileys) does. This asks the sender's phone to re-send decrypted content, bypassing the broken session entirely.

The function already exists in WhatsApp Web's Store:
```js
Store.HistorySync.sendPeerDataOperationRequest(4, {
    historySyncOnDemandRequest: {
        chatJid: msg.id.remote,
        oldestMsgId: msg.id.id,
        oldestMsgFromMe: false,
        onDemandMsgCount: 1,
        oldestMsgTimestampMs: msg.t * 1000
    }
});
```
Same function as type 3 (history sync) — type 4 requests on-demand re-delivery.

**Baileys has three mechanisms wwjs lacks:**
1. Signal retry receipt
2. PDO type 4
3. Session recreation

### Describe an alternate solution if you have one.

- **Minimal:** Emit `message_ciphertext_timeout` after N seconds so apps can detect lost messages
- **Medium:** Add Signal retry receipts (works for DMs, not group SKDMs)

### Additional Context

**How I found this:** Messages visible on the phone never triggered `message` in the library. Added logging at `Store.Msg.on('add')` with 30s/60s/120s rechecks — confirmed messages stay as ciphertext permanently.

**Baileys reference:**
- [Message retry handling](https://github.com/WhiskeySockets/Baileys/blob/master/src/Utils/messages.ts)
- [PDO response handling](https://github.com/WhiskeySockets/Baileys/blob/master/src/Socket/messages-recv.ts)

**Related:** #3920, #127053

igoreira0:
# PR Details

Add ciphertext message retry/recovery mechanism using PDO type 4.

## Description

Implements a ciphertext message retry/recovery mechanism using WhatsApp's Peer Data Operation (PDO) type 4. When a message arrives with `type: "ciphertext"` and fails to decrypt naturally within a configurable timeout, the library sends a PDO type 4 request via `Store.HistorySync.sendPeerDataOperationRequest()` asking the sender's primary device to re-send the decrypted message content, bypassing broken Signal Protocol sessions.

**Changes:**
- New `ciphertextRetry` client option (`enabled`, `initialTimeoutMs`, `retryTimeoutMs`) with sensible defaults (enabled, 30s + 30s)
- Retry-aware ciphertext handler that tracks pending messages, attempts PDO type 4 recovery after initial timeout, and cleans up timers on natural decryption
- New `message_ciphertext_failed` event emitted when decryption permanently fails after retry
- Timer cleanup on `client.destroy()` to prevent dangling references
- Updated TypeScript definitions (`index.d.ts`) and usage example (`example.js`)

**Backward compatible:** The existing `message_ciphertext` event fires exactly as before. The `change:type` listener remains the primary decryption mechanism. The retry is additive and can be disabled with `ciphertextRetry: { enabled: false }`.

## Related Issue(s)

closes #127062

## Motivation and Context

Messages arriving as `ciphertext` sometimes never decrypt due to broken Signal Protocol sessions — caused by missed Sender Key Distribution Messages (groups) or stale pairwise sessions (DMs) when contacts change devices or reinstall WhatsApp. The current implementation waits indefinitely via `msg.once('change:type', ...)` with no timeout, no retry, and no fallback, causing messages to be permanently lost from the application's perspective. This PR addresses that gap by leveraging WhatsApp's existing PDO type 4 mechanism.

## How Has This Been Tested

- Verified ESLint passes with zero code style errors
- Verified client initializes correctly with default `ciphertextRetry` options
- Verified `ciphertextRetry: { enabled: false }` preserves original behavior (indefinite wait)
- Verified timer cleanup on `client.destroy()` produces no errors

### Environment

- Machine OS: Windows 11 Pro
- Phone OS: Android
- Library Version: 1.34.6
- WhatsApp Web Version: 2.3000.x
- Puppeteer Version: 24.37.5
- Browser Type and Version: Chromium (bundled with Puppeteer)
- Node Version: 18.18.0

## Types of changes

- [ ] Dependency change

## Checklist

- [X] I have updated the usage example accordingly (example.js)

### DIFF
--- a/example.js
+++ b/example.js
@@ -554,8 +554,13 @@ client.on('message_ciphertext', (msg) => {
     // Receiving new incoming messages that have been encrypted
     // msg.type === 'ciphertext'
     msg.body = 'Waiting for this message. Check your phone.';
-    
+
     // do stuff here
+});
+
+client.on('message_ciphertext_failed', (msg) => {
+    // A message could not be decrypted even after retry
+    console.log('Message decryption failed permanently:', msg.id._serialized);
 });
 
 client.on('message_revoke_everyone', async (after, before) => {
--- a/index.d.ts
+++ b/index.d.ts
@@ -611,14 +611,14 @@ declare namespace WAWebJS {
         */
         pairWithPhoneNumber?: {phoneNumber: string, showNotification?: boolean, intervalMs?: number}
         /** Configuration for ciphertext message retry/recovery mechanism
-         * @default { enabled: true, initialTimeoutMs: 30000, retryTimeoutMs: 30000 }
+         * @default { enabled: true, initialTimeoutMs: 10000, retryTimeoutMs: 15000 }
          */
         ciphertextRetry?: {
             /** Whether the retry mechanism is enabled @default true */
             enabled?: boolean,
-            /** Time in ms to wait for natural decryption before attempting PDO retry @default 30000 */
+            /** Time in ms to wait for natural decryption before attempting retry @default 10000 */
             initialTimeoutMs?: number,
-            /** Time in ms to wait after PDO retry before declaring failure @default 30000 */
+            /** Time in ms to wait after each retry step before proceeding @default 15000 */
             retryTimeoutMs?: number,
         }
     }
--- a/index.d.ts
+++ b/index.d.ts
@@ -435,6 +435,12 @@ declare namespace WAWebJS {
             message: Message
         ) => void): this
 
+        /** Emitted when a ciphertext message failed to decrypt after retry */
+        on(event: 'message_ciphertext_failed', listener: (
+            /** The message that failed to decrypt */
+            message: Message
+        ) => void): this
+
         /** Emitted when a message is deleted for everyone in the chat */
         on(event: 'message_revoke_everyone', listener: (
             /** The message that was revoked, in its current state. It will not contain the original message's data */
@@ -604,6 +610,17 @@ declare namespace WAWebJS {
          * }
         */
         pairWithPhoneNumber?: {phoneNumber: string, showNotification?: boolean, intervalMs?: number}
+        /** Configuration for ciphertext message retry/recovery mechanism
+         * @default { enabled: true, initialTimeoutMs: 30000, retryTimeoutMs: 30000 }
+         */
+        ciphertextRetry?: {
+            /** Whether the retry mechanism is enabled @default true */
+            enabled?: boolean,
+            /** Time in ms to wait for natural decryption before attempting PDO retry @default 30000 */
+            initialTimeoutMs?: number,
+            /** Time in ms to wait after PDO retry before declaring failure @default 30000 */
+            retryTimeoutMs?: number,
+        }
     }
 
     export interface LocalWebCacheOptions {
@@ -846,6 +863,7 @@ declare namespace WAWebJS {
         CHAT_ARCHIVED = 'chat_archived',
         MESSAGE_RECEIVED = 'message',
         MESSAGE_CIPHERTEXT = 'message_ciphertext',
+        MESSAGE_CIPHERTEXT_FAILED = 'message_ciphertext_failed',
         MESSAGE_CREATE = 'message_create',
         MESSAGE_REVOKED_EVERYONE = 'message_revoke_everyone',
         MESSAGE_REVOKED_ME = 'message_revoke_me',
--- a/src/Client.js
+++ b/src/Client.js
@@ -34,7 +34,11 @@ const {exposeFunctionIfAbsent} = require('./util/Puppeteer');
  * @param {string} options.deviceName - Sets the device name of a current linked device., i.e.: 'TEST'.
  * @param {string} options.browserName - Sets the browser name of a current linked device, i.e.: 'Firefox'.
  * @param {object} options.proxyAuthentication - Proxy Authentication object.
- * 
+ * @param {object} options.ciphertextRetry - Configuration for ciphertext message retry/recovery mechanism
+ * @param {boolean} options.ciphertextRetry.enabled - Whether the retry mechanism is enabled (default: true)
+ * @param {number} options.ciphertextRetry.initialTimeoutMs - Time in ms to wait for natural decryption before attempting PDO retry (default: 30000)
+ * @param {number} options.ciphertextRetry.retryTimeoutMs - Time in ms to wait after PDO retry before declaring failure (default: 30000)
+ *
  * @fires Client#qr
  * @fires Client#authenticated
  * @fires Client#auth_failure
@@ -45,6 +49,7 @@ const {exposeFunctionIfAbsent} = require('./util/Puppeteer');
  * @fires Client#message_revoke_me
  * @fires Client#message_revoke_everyone
  * @fires Client#message_ciphertext
+ * @fires Client#message_ciphertext_failed
  * @fires Client#message_edit
  * @fires Client#media_uploaded
  * @fires Client#group_join
@@ -708,6 +713,16 @@ class Client extends EventEmitter {
             this.emit(Events.MESSAGE_CIPHERTEXT, new Message(this, msg));
         });
 
+        await exposeFunctionIfAbsent(this.pupPage, 'onCiphertextRetryFailedEvent', msg => {
+
+            /**
+                 * Emitted when a ciphertext message failed to decrypt after retry
+                 * @event Client#message_ciphertext_failed
+                 * @param {Message} message
+                 */
+            this.emit(Events.MESSAGE_CIPHERTEXT_FAILED, new Message(this, msg));
+        });
+
         await exposeFunctionIfAbsent(this.pupPage, 'onPollVoteEvent', (votes) => {
             for (const vote of votes) {
                 /**
@@ -719,7 +734,7 @@ class Client extends EventEmitter {
             }
         });
 
-        await this.pupPage.evaluate(() => {
+        await this.pupPage.evaluate((ciphertextRetryOptions) => {
             const { Msg, Chat, WAWebCallCollection } = window.require('WAWebCollections');
             const AppState = window.require('WAWebSocketModel').Socket;
             Msg.on('change', (msg) => { window.onChangeMessageEvent(window.WWebJS.getMessageModel(msg)); });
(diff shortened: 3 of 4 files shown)


Adi1231234:
Hey @igoreira0, thanks for picking this up! This is exactly the mechanism I described in the issue, and I appreciate the clean implementation with the configurable timeouts and the new `message_ciphertext_failed` event.

That said, I want to share some thoughts from my experience running this in production.

While PDO type 4 is a solid recovery mechanism for individual messages, it feels more like a workaround than a root cause fix. In my setup, I periodically start receiving a large volume of ciphertext messages (not just one or two), which means the client would now be sending many PDO type 4 requests in a short time window. That concerns me for two reasons:

1. It adds latency to every affected message (waiting 30s before even attempting recovery)
2. Sending a burst of PDO requests could potentially raise flags on WhatsApp's side

I did some digging into the WhatsApp Web internals and found that the Signal Protocol session management API is actually accessible through `WAWebSignal.Session`. Specifically:

- `deleteRemoteSession(wid)` removes the broken Signal session for a specific sender
- `deleteRemoteInfo(wid)` removes the identity info, forcing a fresh key exchange
- `createSignalSession(wid)` rebuilds the session from scratch
- `hasSignalSessions([wid])` checks whether a session exists

The session data is stored in IndexedDB via `WAWebSignalStorage.getSessionTable()`, and each session is keyed by `{lid}.{deviceId}` (e.g. `157552386044150@lid.0`).

So in theory, a more robust fix would be: when we detect a ciphertext message that isn't decrypting, instead of (or before) sending a PDO type 4, we could delete the broken session and rebuild it. That would fix the root cause so that all future messages from that sender decrypt normally, rather than needing a PDO request for each one.

The flow would look something like:

```js
const signal = require('WAWebSignal');
const wid = Store.WidFactory.createWid(senderLid);

// Delete the broken session
await signal.Session.deleteRemoteSession(wid);

// Optionally force a fresh key exchange
await signal.Session.deleteRemoteInfo(wid);

// Rebuild the session
await signal.Session.createSignalSession(wid);

// Then use PDO type 4 only for messages that were already lost
```

But honestly, I'd rather not need either of these approaches. What I really want to understand is why the sessions break in the first place. In my case it's not just the occasional message from someone who changed devices. Every once in a while I start getting a flood of ciphertext messages, and it feels like something systemic is going wrong with the session state.

Do you have any ideas about what could be causing this? Have you noticed any patterns in your testing? I'd love to figure out the root cause together rather than only treating the symptoms.

Either way, great work on this PR. Would be happy to help test or dig deeper into this!

igoreira0:
@Adi1231234 Thanks for the detailed feedback — really appreciate you digging into the Signal internals!

I've pushed a second commit that incorporates your session recreation approach. I've also lowered the default timeouts since 30s was too conservative — WhatsApp Web normally decrypts within a few seconds, so waiting 30s before acting was unnecessary.

The recovery flow is now a 3-step pipeline:

1. **Wait `initialTimeoutMs` (10s)** for natural decryption
2. **Recreate Signal session** — `deleteRemoteSession(wid)` + `createSignalSession(wid)` via `Store.Signal` (newly exposed from `WAWebSignal`)
3. **Wait `retryTimeoutMs` (15s)**, then fall back to **PDO type 4** for the specific message that was already lost
4. **Wait another `retryTimeoutMs` (15s)**, then emit `message_ciphertext_failed`

Total worst-case: **40s** instead of the original 90s.

This way we fix the root cause first (broken session → future messages from that sender decrypt normally), and only use PDO as a last resort for the message that was already in-flight.

Both steps are wrapped in try/catch with guard checks (`window.Store.Signal && window.Store.Signal.Session`) so it degrades gracefully if the internal API changes.

Regarding the burst scenario you mentioned — since the session recreation targets the sender, multiple ciphertext messages from the same sender would share the benefit of the rebuilt session. Only the first one triggers the recreation; subsequent messages from the same sender should decrypt normally after the session is restored.

As for why sessions break systemically — I haven't been able to pinpoint a definitive root cause yet. Would love to dig into that together. If you have any diagnostic logs from those flood scenarios (e2e_notification patterns, timing, sender distribution), that could help narrow it down.

seowzhenjun0126:
Hi @igoreira0 @Adi1231234 , thank you both for your work. I am also facing the same issue and hope that this PR can fix it.

May I know after applying the PR, did you find the amount of `ciphertext` text reduce significantly? I am wondering how can I measure the changes, maybe by comparing the number of `ciphertext` vs `message_ciphertext_failed`?

igoreira0:
@nicholaschum Thanks for trying it out! To measure the impact, you can track three metrics:

1. `message_ciphertext` — total encrypted messages received (fires immediately on arrival)
2. `message` — messages that eventually decrypted successfully (either naturally or after session recreation / PDO retry)
3. `message_ciphertext_failed` — messages that permanently failed to decrypt after all retry steps

The success rate would be: `(message_ciphertext - message_ciphertext_failed) / message_ciphertext`

A simple way to track this:

```js
let stats = { ciphertext: 0, recovered: 0, failed: 0 };

client.on('message_ciphertext', () => stats.ciphertext++);
client.on('message_ciphertext_failed', () => stats.failed++);
    if (stats.ciphertext > stats.recovered + stats.failed) stats.recovered++;
});

setInterval(() => console.log(stats), 60000);
```

Ideally `message_ciphertext_failed` should be close to zero — the session recreation step (step 1) fixes the root cause for that sender, so subsequent messages should decrypt normally. The PDO type 4 fallback (step 2) handles the specific message that was already in-flight when the session broke.

If you're still seeing a high number of `message_ciphertext_failed`, that would help us understand whether the issue is with the Signal session recreation not working for certain cases, or with the PDO request parameters. Would appreciate any data you can share!

izofis:
## 🔍 E2E Notification Issue - Looking for Input

While testing your ciphertext retry mechanism, I encountered a peculiar issue with `e2e_notification` type messages and wanted to share my findings.

### 📋 What I Observed

**Scenario:**
1. User sends a message in WhatsApp Web ✅
2. Message appears in WhatsApp Web interface ✅
3. Using `fetchMessages()`, I can retrieve the message ✅
4. **BUT**: The message `body` is empty ⚠️
5. **AND**: The `message_create` event never fires ❌

**Example Message Object:**

```
{
  type: 'e2e_notification',
  body: '',  // Empty!
  from: '105XXXXXXXXXX@c.us',
  isNewMsg: false,  // This might be why
  notificationParameters: []
}
```


🤔 My Understanding (Could Be Wrong!)
It seems like e2e_notification messages don't trigger the event listener because:

They exist in WhatsApp Web's local storage
fetchMessages() can retrieve them
But the event system doesn't fire for them
I'm not certain if this is because of [isNewMsg] being false or some other filtering mechanism.

🧪 My Experimental Solution
I tried checking the message type before other filters:

```
window.Store.Msg.on('add', (msg) => {
    const isE2ENotification = msg.type === 'e2e_notification';
    
    if (isE2ENotification) {
        // Try to distinguish real messages from protocol notifications
        const hasBody = msg.body && msg.body.trim() !== '';
        const hasNotificationParams = msg.notificationParameters && 
                                       msg.notificationParameters.length > 0;
        
        if (hasBody || hasNotificationParams) {
            // Process as real message
            window.onAddMessageEvent(window.WWebJS.getMessageModel(msg));
            return;
        } else {
            // Skip empty notification (might be protocol message)
            return;
        }
    }
    
    // ... rest of existing logic
    if (msg.isNewMsg) {
        if(msg.type === 'ciphertext') {
            // Your ciphertext retry logic
        }
    }
});

```
❓ Questions for You
Have you encountered this behavior with e2e_notification messages?
Is my understanding correct? Am I missing something about how WhatsApp Web handles these messages?
Is this approach reasonable, or is there a better way to handle e2e_notification types?
Should this be a separate issue, or is it related to the encryption handling you're addressing?
🎯 Why I'm Asking
I'm trying to understand if:

This is a known limitation
Others are experiencing this
There's a pattern between ciphertext issues and e2e_notification issues
My solution is on the right track or completely misguided
📊 Context
In my testing environment:

Real user messages arriving as e2e_notification 
WhatsApp Web shows them fine
But my library's event system misses them
This is distinct from your ciphertext decryption issue
Your ciphertext retry PR looks excellent, by the way! I'm just trying to understand if there are related encryption issues that might need attention.

Any insights would be greatly appreciated! 🙏

igoreira0:
@nicholaschum Thanks for the detailed write-up! This is a separate issue from the ciphertext retry mechanism in this PR, but I can help clarify what's happening.

`e2e_notification` messages are **not user messages** — they're internal WhatsApp protocol notifications about encryption state changes (identity key rotations, device changes, etc.). That's why:

- `body` is empty — they carry no user content
- `isNewMsg` is `false` — WhatsApp Web intentionally doesn't treat them as new messages
- `message_create` never fires — the `Msg.on('add')` handler checks `isNewMsg` first, so they're filtered out by design

The fact that you can retrieve them via `fetchMessages()` is because they exist in the message store as metadata entries, but they're not meant to be consumed as regular messages.

Your experimental solution of bypassing `isNewMsg` for `e2e_notification` would cause protocol-level notifications (with empty bodies) to be emitted as regular messages to consumers, which isn't ideal.

If you're seeing **actual user messages** that should have content but are arriving as `e2e_notification` with empty bodies, that would be a different bug entirely. But in most cases, these are just WhatsApp signaling that an encryption key changed for a contact — the same notification you see in the chat as "Messages to this chat are end-to-end encrypted" or "Security code changed".

I'd suggest opening a separate issue if you believe user messages are being incorrectly classified as `e2e_notification`. That way we can investigate it independently from the ciphertext retry work here.

izofis:
## 🧪 Test Results - Production Environment

### 📊 Statistics
- **Total encrypted messages detected:** 339
- **Successfully decrypted:** 163 (48.1%)
- **Failed to decrypt:** 176 (51.9%)

### ✅ Successful Decryption Flow
The `message_ciphertext` event is firing correctly and messages are being processed:
- Log: `🔒 Şifreli mesaj algılandı, çözümleniyor... ID: false_XXX@lid_XXX`
- Log: `✅ Şifreli mesaj başarıyla çözüldü! ID: false_XXX@lid_XXX`

Event handler implementation:
```javascript
client.on('message_ciphertext', (msg) => {
    ciphertextStats.ciphertext++;
    logger.info(`🔒 Encrypted message detected: ${msg.id._serialized}`);
});

⚠️ Unexpected Behavior: Protocol Messages
Some messages that trigger the ciphertext event are resolving to [type: "protocol"](vscode-file://vscode-app/c:/Users/izofisNET/AppData/Local/Programs/Microsoft%20VS%20Code/0870c2a0c7/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with empty body:

{
  "type": "protocol",
  "subtype": "ephemeral_setting",
  "body": "",
  "from": "XXX@lid",
  "to": "XXX@c.us",
  "ephemeralDuration": 7776000,
  "disappearingModeTrigger": "account_settings",
}


❌ Failed Decryption: Fanout Subtype
Messages that remain undecrypted have the following structure:

{
  "type": "ciphertext",
  "subtype": "fanout"
}

Adi1231234:
Hey @igoreira0, really appreciate all the work you put into this PR! It pointed me in the right direction and saved me a lot of time when I started investigating this issue more deeply.

I ended up going down a rabbit hole into the WhatsApp Web minified source to understand exactly how the built-in recovery mechanism works, and I wanted to share what I found since it might be interesting.

The main discovery was that WhatsApp Web already has a complete recovery pipeline for ciphertext messages. There's a function called `handlePlaceholderMsgsSeen` in `WAWebNonMessageDataRequestPlaceholderMessageResendUtils` that gets called from the conversation renderer when the user opens a chat containing unresolved placeholder messages. It collects their message keys and calls `sendPeerDataOperationRequest(PLACEHOLDER_MESSAGE_RESEND, { msgKeys: [...] })`.

There are two things preventing this from working in whatsapp-web.js:

1. We never "open" a chat in the UI sense, so `handlePlaceholderMsgsSeen` never fires
2. There's an AB test gate (`isPlaceholderMessageResendEnabled` in `WAWebSyncGatingUtils`) that currently returns `false`. Even if we send the PDO request, the response handler checks this gate and silently drops the response. This gate needs to be overridden for the whole flow to work.

I also noticed that the PDO parameter format is slightly different from what your PR sends. WhatsApp's own `handlePlaceholderMsgsSeen` uses `{ msgKeys: [msg.id] }` (an array of MsgKey objects), and Baileys's `requestPlaceholderResend` confirms this same format. Your PR sends `{ chatId, msgId, senderJid }` which might be why @izofis saw the 51.9% failure rate on fanout messages.

Speaking of @izofis's test data, it was super interesting. The fact that all 176 failures were `subtype: "fanout"` tells us something important. Fanout messages fail because the server didn't have the content available (it's a distribution/BACKFILL issue, not a broken Signal session). So the session recreation step can't help those cases, and there's a small risk it could break working sessions for other messages from the same sender. WhatsApp Web already handles actual Signal decryption errors automatically through `WAWebSendRetryReceiptJob`, so we probably don't need to touch sessions ourselves.

I've opened a new PR (#182677) that takes a more minimal approach based on all of this:

- Overrides the `isPlaceholderMessageResendEnabled` gate (the key missing piece)
- Sends the PDO request with the correct `{ msgKeys }` format after a 5 second grace period
- Adds `message_ciphertext_failed` (loved this idea from your PR!)
- Skips the Signal session manipulation and configurable timeouts to keep things simple

Thanks again for the groundwork here. The discussion in this PR, especially @izofis's production data, was really valuable in understanding the full picture. Would love to hear your thoughts on the findings!

BenyFilho:
@igoreira0  will you update it to comply with the latest changes? Do you know if this fix is still necessary?

BenyFilho:
Close because the new PR about it
https://github.com/wwebjs/whatsapp-web.js/pull/182677

PR #127098, opened 2026-03-05
fix: prevent store model mutation in getContact and getChatModel LID handling

### LINKED ISSUE #201656: perf: getContactById calls fetchBizProfile on every call instead of using cached collection
## Issue Priority

Medium

## Issue Description

`getContactById` (and `msg.getContact()`) calls `Store.BusinessProfile.fetchBizProfile(wid)` on every invocation. This function always makes a network round-trip to WhatsApp servers (~85-100ms per call), even when the result is already available in the local `BusinessProfileCollection`.

The `BusinessProfileCollection` already has `find()` and `get()` methods that check the local cache first and only fetch from the server when the entry is missing, exactly like `Store.Contact.find()` works for contacts.

Since `getContact` is often called multiple times per message (from event handlers, middleware, etc.), this adds hundreds of milliseconds of unnecessary latency per message. Under load (e.g. receiving many messages at once), the concurrent network calls can stack up significantly.

## Steps To Reproduce

1. Connect a client and receive a message
2. Call `msg.getContact()` - takes ~100ms due to `fetchBizProfile`
3. Call `msg.getContact()` again for the same contact - still ~100ms
4. Meanwhile, `Store.BusinessProfile.get(wid)` returns the same result in 0ms

## Evidence

Tested directly in WhatsApp Web DevTools console:

```
// fetchBizProfile - always makes a network call
Store.BusinessProfile.fetchBizProfile(wid)  ->  ~95ms (every time)

// get() - reads from local collection cache
Store.BusinessProfile.get(wid)              ->  0.005ms

// find() - cache first, fetches only if missing
Store.BusinessProfile.find(wid)             ->  0.065ms (cached) / ~85ms (uncached)
```

All three methods return the exact same object reference (`===` is true) when the entry exists in the cache.

For LID contacts, the first `fetchBizProfile` call can take 500-750ms, making the impact even larger.

## Proposed Fix

In `src/util/Injected/Utils.js`, replace:

```js
const bizProfile = await window.Store.BusinessProfile.fetchBizProfile(wid);
```

with:

```js
const bizProfile = await window.Store.BusinessProfile.find(wid);
```

`find()` checks the local `BusinessProfileCollection` cache first and only fetches from the server when the entry is not present. This matches the pattern already used for `Store.Contact.find(wid)` on the line above.

## WhatsApp Account Type

Standard

## Authentication Strategy

LocalAuth

## whatsapp-web.js Version

1.26.1-alpha.2

## WhatsApp Web Version

2.3000+

## Browser Type

Chromium

## Browser Version

Chrome 145

## Computer OS Version

Windows 11

## Phone OS Version

Android 14

## Node.js Version

20.x

## Checklist

- [x] I'm using the latest version of whatsapp-web.js
- [x] I've searched existing issues for similar reports
- [x] I've verified this is a library issue, not a WhatsApp limitation

Adi1231234:
## Summary

`getContact()` and `getChatModel()` currently overwrite `.id` directly on live Backbone model references from WhatsApp Web's Contact/Participant store. Since these are shared references, the mutation permanently corrupts the store for the lifetime of the session.

After the first call to `getContact()` with a LID contact, all subsequent lookups for that contact crash with:
- `TypeError: Cannot read properties of undefined (reading '_serialized')`
- `Error: Data passed to getter must include an id property (it's how we memoize) but got undefined`

### Root cause

In `getContact()` (Utils.js):
```js
let contact = await Contact.find(wid);
if (contact.id._serialized.endsWith('@lid')) {
    contact.id = contact.phoneNumber; // mutates the live store model
}
```

`Contact.find()` returns a reference to the model in the store, not a copy. Overwriting `contact.id` replaces the WID object (which has `_serialized`, `user`, `server`) with `contact.phoneNumber` (another WID object) or `undefined` if the contact has no phone number mapped yet. Once corrupted, the store entry stays broken until the WhatsApp Web page is fully reloaded.

The same pattern exists in `getChatModel()` for group participants:
```js
chat.groupMetadata.participants._models
    .filter(x => x.id?._serialized?.endsWith('@lid'))
    .forEach(x => x.contact?.phoneNumber && (x.id = x.contact.phoneNumber));
```

### Fix

**`getContact()`**: serialize the contact first via `getContactModel()`, then apply the LID to phone resolution on the returned (serialized) model only. Also adds a null guard for `contact`/`contact.id` and wraps `fetchBizProfile` in try/catch since it can fail for non-business contacts.

**`getChatModel()`**: serialize group metadata first, then patch LID to phone on the serialized participants array instead of the live Backbone models.

In both cases the caller still receives the phone number instead of the LID, but the store stays clean.

### Performance: use `BusinessProfile.find()` instead of `fetchBizProfile()`

This PR also replaces `BusinessProfile.fetchBizProfile(wid)` with `BusinessProfile.find(wid)`.

`fetchBizProfile()` always makes a WebSocket network call to WhatsApp servers, taking ~85-100ms per contact regardless of whether the profile is already cached locally. `find()` checks the local `BusinessProfileCollection` cache first (returning in ~0.065ms) and only fetches from the server when the profile is missing. Both return the same data structure.

This matches the established pattern used for contacts (`Contact.find()`) and avoids unnecessary network round-trips when processing multiple contacts in quick succession.

Fixes #127054
Fixes #201656

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -1055,14 +1055,10 @@ exports.LoadUtils = () => {
         const contact = await window
             .require('WAWebCollections')
             .Contact.find(wid);
-        try {
-            const bizProfile = await window
-                .require('WAWebCollections')
-                .BusinessProfile.find(wid);
-            bizProfile.profileOptions && (contact.businessProfile = bizProfile);
-        } catch (_) {
-            /* find() can fail for non-business contacts */
-        }
+        const bizProfile = await window
+            .require('WAWebCollections')
+            .BusinessProfile.find(wid);
+        bizProfile.profileOptions && (contact.businessProfile = bizProfile);
         return window.WWebJS.getContactModel(contact);
     };
 
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -286,9 +286,7 @@ exports.LoadUtils = () => {
 
         let vcardOptions = {};
         if (options.contactCard) {
-            let contact = window
-                .require('WAWebCollections')
-                .Contact.get(options.contactCard);
+            let contact = await window.WWebJS.findContact(options.contactCard);
             vcardOptions = {
                 body: window
                     .require('WAWebFrontendVcardUtils')
@@ -298,8 +296,10 @@ exports.LoadUtils = () => {
             };
             delete options.contactCard;
         } else if (options.contactCardList) {
-            let contacts = options.contactCardList.map((c) =>
-                window.require('WAWebCollections').Contact.get(c),
+            let contacts = await Promise.all(
+                options.contactCardList.map((c) =>
+                    window.WWebJS.findContact(c),
+                ),
             );
             let vcards = contacts.map((c) =>
                 window
@@ -956,14 +956,16 @@ exports.LoadUtils = () => {
                 window.require('WAWebCollections').GroupMetadata ||
                 window.require('WAWebCollections').WAWebGroupMetadataCollection;
             await groupMetadata.update(chatWid);
-            chat.groupMetadata.participants._models
-                .filter((x) => x.id?._serialized?.endsWith('@lid'))
-                .forEach(
-                    (x) =>
-                        x.contact?.phoneNumber &&
-                        (x.id = x.contact.phoneNumber),
-                );
-            model.groupMetadata = chat.groupMetadata.serialize();
+            const serializedMetadata = chat.groupMetadata.serialize();
+            const participantModels = chat.groupMetadata.participants._models;
+            for (const p of serializedMetadata.participants || []) {
+                if (!p.id?._serialized?.endsWith('@lid')) continue;
+                const phone = participantModels.find(
+                    (m) => m.id?._serialized === p.id._serialized,
+                )?.contact?.phoneNumber;
+                if (phone) p.id = phone;
+            }
+            model.groupMetadata = serializedMetadata;
             model.isReadOnly = chat.groupMetadata.announce;
         }
 
@@ -1005,8 +1007,18 @@ exports.LoadUtils = () => {
         return model;
     };
 
+    window.WWebJS.findContact = async (id) => {
+        const wid = window.require('WAWebWidFactory').createWid(id);
+        return window.require('WAWebCollections').Contact.find(wid);
+    };
+
     window.WWebJS.getContactModel = (contact) => {
         let res = contact.serialize();
+
+        if (res.id?._serialized?.endsWith('@lid') && contact.phoneNumber) {
+            res.id = contact.phoneNumber;
+        }
+
         res.isBusiness =
             contact.isBusiness === undefined ? false : contact.isBusiness;
 
@@ -1040,16 +1052,17 @@ exports.LoadUtils = () => {
 
     window.WWebJS.getContact = async (contactId) => {
         const wid = window.require('WAWebWidFactory').createWid(contactId);
-        let contact = await window
+        const contact = await window
             .require('WAWebCollections')
             .Contact.find(wid);
-        if (contact.id._serialized.endsWith('@lid')) {
-            contact.id = contact.phoneNumber;
+        try {
+            const bizProfile = await window
+                .require('WAWebCollections')
+                .BusinessProfile.fetchBizProfile(wid);
+            bizProfile.profileOptions && (contact.businessProfile = bizProfile);
+        } catch (_) {
+            /* fetchBizProfile can fail for non-business contacts */
         }
-        const bizProfile = await window
-            .require('WAWebCollections')
-            .BusinessProfile.fetchBizProfile(wid);
-        bizProfile.profileOptions && (contact.businessProfile = bizProfile);
         return window.WWebJS.getContactModel(contact);
     };
 
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -286,7 +286,9 @@ exports.LoadUtils = () => {
 
         let vcardOptions = {};
         if (options.contactCard) {
-            let contact = await window.WWebJS.findContact(options.contactCard);
+            let contact = await window
+                .require('WAWebCollections')
+                .Contact.find(options.contactCard);
             vcardOptions = {
                 body: window
                     .require('WAWebFrontendVcardUtils')
@@ -298,7 +300,7 @@ exports.LoadUtils = () => {
         } else if (options.contactCardList) {
             let contacts = await Promise.all(
                 options.contactCardList.map((c) =>
-                    window.WWebJS.findContact(c),
+                    window.require('WAWebCollections').Contact.find(c),
                 ),
             );
             let vcards = contacts.map((c) =>
@@ -956,14 +958,10 @@ exports.LoadUtils = () => {
                 window.require('WAWebCollections').GroupMetadata ||
                 window.require('WAWebCollections').WAWebGroupMetadataCollection;
             await groupMetadata.update(chatWid);
+            const { toPn } = window.require('WAWebLidMigrationUtils');
             const serializedMetadata = chat.groupMetadata.serialize();
-            const participantModels = chat.groupMetadata.participants._models;
             for (const p of serializedMetadata.participants || []) {
-                if (!p.id?._serialized?.endsWith('@lid')) continue;
-                const phone = participantModels.find(
-                    (m) => m.id?._serialized === p.id._serialized,
-                )?.contact?.phoneNumber;
-                if (phone) p.id = phone;
+                p.id = toPn(p.id) ?? p.id;
             }
             model.groupMetadata = serializedMetadata;
             model.isReadOnly = chat.groupMetadata.announce;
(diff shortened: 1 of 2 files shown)


DiegoFelipe (approved):

Adi1231234:
Hey @purpshell @BenyFilho 👋

This one fixes a sneaky bug where getContact/getChatModel permanently corrupt the WhatsApp Web store by mutating live Backbone model references. After the first call, the store stays broken for the entire session.

Would love a review when you get a chance! I'm around on Discord if you want to chat about it.

BenyFilho:
> Hey @purpshell @BenyFilho 👋
> 
> This one fixes a sneaky bug where getContact/getChatModel permanently corrupt the WhatsApp Web store by mutating live Backbone model references. After the first call, the store stays broken for the entire session.
> 
> Would love a review when you get a chance! I'm around on Discord if you want to chat about it.

Need user tests to validate it, and update your changes with branch main

Adi1231234:
Hey @BenyFilho, thanks for looking at this! 🙏

I've rebased the branch onto the latest main, so the conflicts are resolved now. I also fixed the ESLint issue that was causing the CI to fail, so it should pass now.

About user tests, could you clarify what you mean exactly? Are you looking for automated test cases (like unit/integration tests), or manual testing steps that someone can follow to reproduce and verify the fix? Happy to add whatever is needed to help move this forward.

BenyFilho:
> Hey @BenyFilho, thanks for looking at this! 🙏
> 
> I've rebased the branch onto the latest main, so the conflicts are resolved now. I also fixed the ESLint issue that was causing the CI to fail, so it should pass now.
> 
> About user tests, could you clarify what you mean exactly? Are you looking for automated test cases (like unit/integration tests), or manual testing steps that someone can follow to reproduce and verify the fix? Happy to add whatever is needed to help move this forward.

Other users' validation, make sure it solves real problems.

Adi1231234:
Hey @BenyFilho, thanks again for your time!
I see that @DiegoFelipe  approve that.
I fixed the conflicts and the CI

BenyFilho:
> Hey @BenyFilho, thanks again for your time! I see that @DiegoFelipe approve that. I fixed the conflicts and the CI

Hi,

This approach is not a good pattern for the project, as it may fail silently, and we want to avoid that. In most situations, it is better to throw an error than to return nothing, since silent failures make debugging harder and can hide real problems.

```
try {
      const bizProfile = await window
            .require('WAWebCollections')
            .BusinessProfile.find(wid);
       bizProfile.profileOptions && (contact.businessProfile = bizProfile);
} catch (_) {
      /* find() can fail for non-business contacts */
}
```

I’m asking other contributors to review your PRs, and we are also waiting for feedback from users who are affected by the issues you are trying to solve. This process can be slow, but we are working to make it better over time.

Please keep in mind that everyone here is a volunteer and has many other responsibilities. There are also other factors involved in reviewing and approving changes. I truly understand that these fixes are important for your use case, but as mentioned before, they also need to be relevant for other users.

Constantly pinging me or other contributors will not make the process faster, unfortunately.

Adi1231234:
@BenyFilho fair point. I actually went and tested this directly in WhatsApp Web DevTools. Turns out the comment was wrong: `BusinessProfile.find()` doesn't throw for non-business contacts. The server returns a valid BusinessProfile model with `profileOptions: null` for them, so the existing `bizProfile.profileOptions &&` guard already handles it. No try/catch needed.

Fixed and pushed.

On the pings, understood, sorry about that.

src/util/Injected/Utils.js
```diff
@@ -1005,8 +1007,18 @@ exports.LoadUtils = () => {
         return model;
     };
 
+    window.WWebJS.findContact = async (id) => {
```
sofi-ans: The createWid call here is not needed. Contact.find() accepts a plain string directly, so findContact ends up being a trivial wrapper over Contact.find(id) with no added logic. All the call sites (contactCard, contactCardList, block, unblock) can just call Contact.find(id) inline and the helper can be dropped

src/util/Injected/Utils.js
```diff
@@ -956,14 +956,16 @@ exports.LoadUtils = () => {
...
-                    (x) =>
-                        x.contact?.phoneNumber &&
-                        (x.id = x.contact.phoneNumber),
-                );
-            model.groupMetadata = chat.groupMetadata.serialize();
+            const serializedMetadata = chat.groupMetadata.serialize();
+            const participantModels = chat.groupMetadata.participants._models;
+            for (const p of serializedMetadata.participants || []) {
+                if (!p.id?._serialized?.endsWith('@lid')) continue;
+                const phone = participantModels.find(
```
sofi-ans: This makes the resolution O(n²) since participantModels.find(m => m.id._serialized === p.id._serialized) does a linear scan for every LID participant. WA already has WAWebLidMigrationUtils.toPn(wid) for this exact case, it resolves a LID WID to its PN using an internal cache so each lookup is O(1). The loop becomes:

const { toPn } = window.require('WAWebLidMigrationUtils');
for (const p of serializedMetadata.participants || []) {
    p.id = toPn(p.id) ?? p.id;
}

src/util/Injected/Utils.js
```diff
@@ -1005,8 +1007,18 @@ exports.LoadUtils = () => {
...
 
+    window.WWebJS.findContact = async (id) => {
+        const wid = window.require('WAWebWidFactory').createWid(id);
+        return window.require('WAWebCollections').Contact.find(wid);
+    };
+
     window.WWebJS.getContactModel = (contact) => {
         let res = contact.serialize();
+
+        if (res.id?._serialized?.endsWith('@lid') && contact.phoneNumber) {
```
sofi-ans: WAWebWidFactory.createWidFromWidLike(contact.id).isLid() would be more correct here. endsWith works most of the time but createWidFromWidLike handles the case where contact.id is a plain object { _serialized: "...@lid" } without WID prototype methods on it, which at least one contact in production actually has. Calling .isLid() directly on that would throw.

sofi-ans (changes requested):

sofi-ans:
@Adi1231234 Hey! Just a quick ping about the comments I left on the CR yesterday.
This issue affects me as well, so fixing it would be really helpful.

Adi1231234:
@sofi-ans  Great catches, honestly impressive you spotted all of these. Took me a while to verify everything against the live WA source before pushing the fix  🙏

sofi-ans:
@Adi1231234  Thank you!
But I saw one more issue - why are we calling BusinessProfile.find() for every contact? For non-business contacts this is a wasted that always comes back empty.

I dug into WA's own source and they guard this behind isBusiness || isEnterprise (WAWebUseBusinessProfile.react:11193). Makes sense - no point querying business profiles for regular contacts.

Also worth noting: if this network call fails for any reason, the whole getContact() blows up even though the contact itself was already loaded fine.

BenyFilho:
Other question, why this PR is changing the contact block and unblock?

Adi1231234:
@BenyFilho The block/unblock changes are two bugs I noticed while I was going through the contact resolution flow:

block() was calling getChat() and passing the result to blockContact() as if it were a contact. blockContact actually expects a Contact model though - it runs unproxy(contact) on it and resolves the chat on its own. It happens to work because both have .id, but it's wrong.

unblock() was using Contact.get() which is synchronous and just returns undefined if the contact isn't in the cache yet. Switched it to await Contact.find() so it actually waits for the contact to load.

They're small fixes so I threw them in here, but I can split them out if that's cleaner. Would you prefer that?

Adi1231234:
@sofi-ans Nice one, pushed a fix for this. Skipping the BusinessProfile.find() now for non-business contacts.

BenyFilho:
> @BenyFilho The block/unblock changes are two bugs I noticed while I was going through the contact resolution flow:
> 
> block() was calling getChat() and passing the result to blockContact() as if it were a contact. blockContact actually expects a Contact model though - it runs unproxy(contact) on it and resolves the chat on its own. It happens to work because both have .id, but it's wrong.
> 
> unblock() was using Contact.get() which is synchronous and just returns undefined if the contact isn't in the cache yet. Switched it to await Contact.find() so it actually waits for the contact to load.
> 
> They're small fixes so I threw them in here, but I can split them out if that's cleaner. Would you prefer that?

First, a PR should be focused and should not include multiple unrelated changes in different parts of the code.
There was also a reason why getChat was used instead of getContact. Please test your changes by trying to block a contact you have never interacted with before, as this scenario may behave differently.

Adi1231234:
@BenyFilho You're right, appreciate you catching that. I tested it and Contact.find() does fail for contacts without a chat because blockContact needs a LID or a chat with accountLid.
Reverted the block/unblock changes from this PR and opened #201671  with a proper fix.
Honestly this is exactly the kind of review that makes the project better. Really appreciate you taking the time to dig into this, learned a lot from it.

sofi-ans:
@Adi1231234 Looks great, nice work. This fixes an issue I’ve been dealing with for a while. Approved.
@BenyFilho Can we go ahead and merge this? It’s been stable for me, and I’m not seeing the issue anymore.

sofi-ans (approved):

BenyFilho (approved):

BenyFilho (approved):

sofi-ans:
@BenyFilho @Adi1231234 Thank you 👍

aliyss (approved):

PR #182677, opened 2026-03-11
fix: recover ciphertext messages via PLACEHOLDER_MESSAGE_RESEND (PDO type 4)

### LINKED ISSUE #127062: Feature: Add ciphertext message retry/recovery mechanism (PDO type 4)

### LINKED ISSUE #201668: message event fires for revoked messages when original arrived as ciphertext
**Issue Priority**

Medium

**Issue Description**

The `message` event fires for messages of type `revoked`, which should never happen. Revoked messages are supposed to be handled by `message_revoke_everyone` only.

This is a race condition triggered when:

1. A message arrives as `ciphertext` (Signal session decryption is still pending)
2. Before the ciphertext resolves, the sender deletes the message for everyone

The ciphertext deferral logic in `Client.js` registers a `once('change:type')` listener on the model and emits `message` when the type changes. The problem is that WhatsApp's internal revoke handler also triggers `change:type` on the same model (it calls `model.set({ type: 'revoked', ... })` in place, on the existing ciphertext model in the collection). The listener fires, doesn't check the resolved type, and emits `message` with a revoked message.

**Steps To Reproduce**

1. Listen on the `message` event
2. Have another account send a message to your client's number
3. Quickly (within ~5 seconds, before ciphertext resolves) have the sender delete the message for everyone
4. The `message` event fires with `msg.type === 'revoked'`

This is easier to reproduce on the first message between two accounts, or right after a Signal session reset, since decryption takes longer in those cases.

**Code Sample**

```typescript
    // unexpectedly fires with msg.type === 'revoked'
    console.log(msg.type);
});
```

**Root Cause**

In `Client.js` (around line 1052), the ciphertext deferral callback passes the resolved message directly to `onAddMessageEvent` without checking whether the type became `revoked`:

```typescript
msg.once('change:type', (_msg) =>
    window.onAddMessageEvent(
        window.WWebJS.getMessageModel(_msg),
    ),
);
```

**Proposed Fix**

Add a check in the callback:

```typescript
msg.once('change:type', (_msg) => {
    if (_msg.type === 'revoked') return;
    window.onAddMessageEvent(
        window.WWebJS.getMessageModel(_msg),
    );
});
```

**Checklist**

- [x] **I use the latest released version of whatsapp-web.js.**
- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have verified this is a library issue, not a problem with my code.

**WhatsApp Account Type:** Standard
**Authentication Strategy:** LocalAuth
**whatsapp-web.js Version:** 1.34.6
**WhatsApp Web Version:** 2.3000.1021202220
**Browser Type:** Chromium (Puppeteer)
**Browser Version:** 132.0.6834.83
**Computer OS Version:** Ubuntu 22.04
**Phone OS Version:** Android 14
**Node.js Version:** 22.14.0

Adi1231234:
## Summary

Adds automatic recovery for ciphertext messages using WhatsApp's built-in PLACEHOLDER_MESSAGE_RESEND mechanism (Peer Data Operation type 4).

When a message arrives as `ciphertext` and doesn't decrypt naturally within 5 seconds, the library now requests the message content from the sender's primary device. This is the exact same mechanism WhatsApp Web uses internally through `handlePlaceholderMsgsSeen`.

Also adds a new `message_ciphertext_failed` event that fires at 15 seconds if recovery also fails, so consumers can track and handle permanently lost messages.

Also fixes a related bug where the `message` event fires for revoked messages when the original arrived as `ciphertext` (#201668).

Closes #127062
Fixes #201668
Related: #127079

## The Problem

The current `Msg.on('add')` handler listens for ciphertext messages and sets up a `msg.once('change:type')` listener to wait for natural decryption. But if the type never changes, the message is silently lost. No timeout, no recovery attempt, nothing.

In WhatsApp Web's UI this is rarely noticeable because there's a built-in recovery mechanism that kicks in when the user opens a chat containing placeholder messages. But since whatsapp-web.js is headless and never "opens" a chat in the UI sense, this recovery path is never triggered.

There's also a second issue with the same `once('change:type')` listener: if the sender revokes the message before it decrypts, WhatsApp's internal revoke handler calls `model.set({ type: 'revoked', ... })` on the existing model in place, which triggers `change:type`. The listener was firing and emitting `message` with `type === 'revoked'`, which should never happen.

## Root Cause Analysis

I spent some time digging through WhatsApp Web's minified source to understand exactly how placeholder recovery works. Here's what I found:

**The recovery function**: `WAWebNonMessageDataRequestPlaceholderMessageResendUtils.handlePlaceholderMsgsSeen` is called from WhatsApp Web's conversation renderer when placeholder (ciphertext) messages become visible in the chat. It filters ciphertext messages, collects their message keys, and calls `sendPeerDataOperationRequest(PLACEHOLDER_MESSAGE_RESEND, { msgKeys: [...] })`.

**The AB test gate**: There's a function called `isPlaceholderMessageResendEnabled()` in `WAWebSyncGatingUtils` that checks an AB prop called `placeholder_message_resend`. This currently returns `false`, which means even if we send the PDO request correctly, the response handler (`handlePlaceholderResendOperationRequestResponse`) will silently drop the response. Both the sender and the response handler reference this gate via lazy `require()` calls (not destructured imports), so overriding the property on the module object propagates correctly to all consumers. I verified that `require()` returns the same cached object.

**The response flow**: When the phone responds with the message content, `handlePlaceholderResendOperationRequestResponse` decodes the protobuf via `WebMessageInfoSpec`, parses it with `parseWebMessageInfo`, and feeds it into `processMsgs`. This updates the message model in the Backbone collection, which triggers the `change:type` event that our existing listener already handles. So we don't need to add any new response handling logic.

**Baileys reference**: The Baileys library implements the same mechanism in `requestPlaceholderResend` (`src/Socket/messages-recv.ts`), confirming both the approach and the correct parameter format: `{ placeholderMessageResendRequest: [{ messageKey }], peerDataOperationRequestType: PLACEHOLDER_MESSAGE_RESEND }`.

**The revoked race condition**: When a sender revokes a message, `WAWebRevokeMsgAction.revoke` is called on the existing model in the MsgCollection. It calls `model.set({ type: 'revoked', id: newKey, ... })` which triggers `change:type` on the ciphertext model. The deferred listener had no guard for this case.

## Changes

Three small changes in `Client.js` and one line each in `Constants.js` and `index.d.ts`:

1. **Gate override**: At the top of the `pupPage.evaluate()` block, overrides `isPlaceholderMessageResendEnabled` to return `true` so the response handler actually processes recovered messages. Wrapped in try/catch in case the module doesn't exist in future WhatsApp Web versions.

2. **Recovery timer**: When a ciphertext message arrives, a 5 second timer starts. If the message hasn't resolved naturally by then, sends `sendPeerDataOperationRequest(4, { msgKeys: [msg.id] })`. The 5 second grace period lets natural server-side resolution happen first (which typically takes 1-3 seconds). If `change:type` fires before the timer, it's cleared and no PDO request is sent.

3. **Failure event**: A second timer at 15 seconds emits `message_ciphertext_failed` if recovery also didn't work. Also cleared on `change:type`.

4. **Revoked guard**: Added `if (_msg.type === 'revoked') return;` in the `once('change:type')` callback so the `message` event is never emitted for revoked messages, regardless of how the type change was triggered.

## Relationship to #127079

@igoreira0's PR (#127079) tackles the same problem and was really helpful during my research. I built on his work but made some different choices after studying the WhatsApp Web source directly:

**PDO parameters**: Changed from `{ chatId, msgId, senderJid }` to `{ msgKeys: [msg.id] }`. This matches what `handlePlaceholderMsgsSeen` actually uses internally. The `msgKeys` format sends the full MsgKey object array, which is what the response handler expects when reconstructing the message.

**Gate override**: Added the `isPlaceholderMessageResendEnabled` override. Without it, the response handler silently discards recovered messages because the AB test returns `false`. This was the key missing piece.

**Signal session deletion**: Removed. @izofis's production data from #127079 was very telling: 51.9% failure rate, and every single failure was `subtype: "fanout"`. Fanout messages fail because the server didn't have the content available (it's a BACKFILL/distribution issue), not because the Signal session is broken. Session recreation can't help those cases and risks breaking working sessions for other messages from the same sender. WhatsApp Web already handles actual Signal decryption errors automatically through `WAWebSendRetryReceiptJob`.

**Configurable timeouts**: Removed. The 5 second grace period is a technical choice based on WhatsApp's own behavior, not a user preference. Keeping the API surface minimal.

## Verified Against Live WhatsApp Web

All module references were tested on a live WhatsApp Web instance:

- `WAWebSyncGatingUtils.isPlaceholderMessageResendEnabled` exists and returns `false`
- `WAWebSendNonMessageDataRequest.sendPeerDataOperationRequest` exists and is async
- `handlePlaceholderMsgsSeen` exists and uses `{ msgKeys }` format
- `require()` returns cached module objects, confirming gate override propagates
- Response handler checks gate via lazy `require()`, not a destructured import

## Test Plan

- [ ] Ciphertext messages that resolve naturally (< 5s) still work as before
- [ ] Ciphertext messages that don't resolve naturally get recovered after ~5s via PDO
- [ ] `message_ciphertext_failed` fires at ~15s when both natural and PDO recovery fail
- [ ] Gate override doesn't cause errors on WhatsApp Web versions without `WAWebSyncGatingUtils`
- [ ] `message` event does not fire when a ciphertext message is revoked before it decrypts

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1070,43 +1070,60 @@ class Client extends EventEmitter {
                     prevState,
                 );
             });
-            Msg.on('add', (msg) => {
-                if (msg.isNewMsg) {
-                    if (msg.type === 'ciphertext') {
-                        // defer message event until ciphertext is resolved (type changed)
-                        const resendTimer = setTimeout(() => {
-                            try {
-                                window
-                                    .require(
-                                        'WAWebNonMessageDataRequestPlaceholderMessageResendUtils',
-                                    )
-                                    .handlePlaceholderMsgsSeen([msg], true);
-                            } catch (_) {
-                                // module may not be available
-                            }
-                        }, 5000);
-                        const failTimer = setTimeout(() => {
-                            window.onCiphertextFailedEvent(
-                                window.WWebJS.getMessageModel(msg),
-                            );
-                        }, 15000);
-                        msg.once('change:type', (_msg) => {
-                            clearTimeout(resendTimer);
-                            clearTimeout(failTimer);
-                            if (_msg.type === 'revoked') return;
-                            window.onAddMessageEvent(
-                                window.WWebJS.getMessageModel(_msg),
-                            );
-                        });
-                        window.onAddMessageCiphertextEvent(
-                            window.WWebJS.getMessageModel(msg),
-                        );
-                    } else {
-                        window.onAddMessageEvent(
-                            window.WWebJS.getMessageModel(msg),
-                        );
+            const pendingResend = new Set();
+            let resendFlush = null;
+
+            function requestResend(msg) {
+                pendingResend.add(msg);
+                if (resendFlush) return;
+                resendFlush = setTimeout(() => {
+                    resendFlush = null;
+                    const msgs = [...pendingResend];
+                    pendingResend.clear();
+                    if (msgs.length === 0) return;
+                    try {
+                        window
+                            .require(
+                                'WAWebNonMessageDataRequestPlaceholderMessageResendUtils',
+                            )
+                            .handlePlaceholderMsgsSeen(msgs, true);
+                    } catch (_) {
+                        // module may not be available
                     }
+                }, 5000);
+            }
+
+            Msg.on('add', (msg) => {
+                if (!msg.isNewMsg) return;
+
+                if (msg.type !== 'ciphertext') {
+                    window.onAddMessageEvent(
+                        window.WWebJS.getMessageModel(msg),
+                    );
+                    return;
                 }
+
+                requestResend(msg);
+
+                const failTimer = setTimeout(() => {
+                    if (msg.type !== 'ciphertext') return;
+                    window.onCiphertextFailedEvent(
+                        window.WWebJS.getMessageModel(msg),
+                    );
+                }, 15000);
+
+                msg.once('change:type', (_msg) => {
+                    clearTimeout(failTimer);
+                    pendingResend.delete(_msg);
+                    if (_msg.type === 'revoked') return;
+                    window.onAddMessageEvent(
+                        window.WWebJS.getMessageModel(_msg),
+                    );
+                });
+
+                window.onAddMessageCiphertextEvent(
+                    window.WWebJS.getMessageModel(msg),
+                );
             });
             Chat.on('change:unreadCount', (chat) => {
                 window.onChatUnreadCountEvent(chat);
--- a/src/Client.js
+++ b/src/Client.js
@@ -1002,12 +1002,8 @@ class Client extends EventEmitter {
             const AppState = window.require('WAWebSocketModel').Socket;
 
             // Enable placeholder message resend (recovery for ciphertext messages)
-            try {
-                const gatingUtils = window.require('WAWebSyncGatingUtils');
-                gatingUtils.isPlaceholderMessageResendEnabled = () => true;
-            } catch (_) {
-                // Module may not be available in all versions
-            }
+            const gatingUtils = window.require('WAWebSyncGatingUtils');
+            gatingUtils.isPlaceholderMessageResendEnabled = () => true;
 
             Msg.on('change', (msg) => {
                 window.onChangeMessageEvent(window.WWebJS.getMessageModel(msg));
@@ -1081,15 +1077,11 @@ class Client extends EventEmitter {
                     const msgs = [...pendingResend];
                     pendingResend.clear();
                     if (msgs.length === 0) return;
-                    try {
-                        window
-                            .require(
-                                'WAWebNonMessageDataRequestPlaceholderMessageResendUtils',
-                            )
-                            .handlePlaceholderMsgsSeen(msgs, true);
-                    } catch (_) {
-                        // module may not be available
-                    }
+                    window
+                        .require(
+                            'WAWebNonMessageDataRequestPlaceholderMessageResendUtils',
+                        )
+                        .handlePlaceholderMsgsSeen(msgs, true);
                 }, 5000);
             }
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -1118,6 +1118,13 @@ class Client extends EventEmitter {
                     return;
                 }
 
+                window.onAddMessageCiphertextEvent(
+                    window.WWebJS.getMessageModel(msg),
+                );
+
+                if (msg.subtype && msg.subtype.endsWith('_unavailable_fanout'))
+                    return;
+
                 requestResend(msg);
 
                 const failTimer = setTimeout(() => {
@@ -1135,10 +1142,6 @@ class Client extends EventEmitter {
                         window.WWebJS.getMessageModel(_msg),
                     );
                 });
-
-                window.onAddMessageCiphertextEvent(
-                    window.WWebJS.getMessageModel(msg),
-                );
             });
             Chat.on('change:unreadCount', (chat) => {
                 window.onChatUnreadCountEvent(chat);
--- a/src/Client.js
+++ b/src/Client.js
@@ -1076,13 +1076,11 @@ class Client extends EventEmitter {
                         // defer message event until ciphertext is resolved (type changed)
                         const resendTimer = setTimeout(() => {
                             try {
-                                const { sendPeerDataOperationRequest } =
-                                    window.require(
-                                        'WAWebSendNonMessageDataRequest',
-                                    );
-                                sendPeerDataOperationRequest(4, {
-                                    msgKeys: [msg.id],
-                                }).catch(() => {});
+                                window
+                                    .require(
+                                        'WAWebNonMessageDataRequestPlaceholderMessageResendUtils',
+                                    )
+                                    .handlePlaceholderMsgsSeen([msg], true);
                             } catch (_) {
                                 // module may not be available
                             }
--- a/src/Client.js
+++ b/src/Client.js
@@ -1095,6 +1095,7 @@ class Client extends EventEmitter {
                         msg.once('change:type', (_msg) => {
                             clearTimeout(resendTimer);
                             clearTimeout(failTimer);
+                            if (_msg.type === 'revoked') return;
                             window.onAddMessageEvent(
                                 window.WWebJS.getMessageModel(_msg),
                             );
--- a/src/Client.js
+++ b/src/Client.js
@@ -957,11 +957,27 @@ class Client extends EventEmitter {
             'onAddMessageCiphertextEvent',
             (msg) => {
                 /**
-                 * Emitted when messages are edited
+                 * Emitted when a message is received as ciphertext (not yet decrypted)
                  * @event Client#message_ciphertext
                  * @param {Message} message
                  */
                 this.emit(Events.MESSAGE_CIPHERTEXT, new Message(this, msg));
+            },
+        );
+
+        await exposeFunctionIfAbsent(
+            this.pupPage,
+            'onCiphertextFailedEvent',
+            (msg) => {
+                /**
+                 * Emitted when a ciphertext message failed to decrypt after recovery attempt
+                 * @event Client#message_ciphertext_failed
+                 * @param {Message} message
+                 */
+                this.emit(
+                    Events.MESSAGE_CIPHERTEXT_FAILED,
+                    new Message(this, msg),
+                );
             },
         );
 
@@ -984,6 +1000,15 @@ class Client extends EventEmitter {
             const { Msg, Chat, WAWebCallCollection } =
                 window.require('WAWebCollections');
             const AppState = window.require('WAWebSocketModel').Socket;
+
+            // Enable placeholder message resend (recovery for ciphertext messages)
+            try {
+                const gatingUtils = window.require('WAWebSyncGatingUtils');
+                gatingUtils.isPlaceholderMessageResendEnabled = () => true;
+            } catch (_) {
+                // Module may not be available in all versions
+            }
+
             Msg.on('change', (msg) => {
                 window.onChangeMessageEvent(window.WWebJS.getMessageModel(msg));
             });
@@ -1049,11 +1074,31 @@ class Client extends EventEmitter {
                 if (msg.isNewMsg) {
                     if (msg.type === 'ciphertext') {
                         // defer message event until ciphertext is resolved (type changed)
-                        msg.once('change:type', (_msg) =>
+                        const resendTimer = setTimeout(() => {
+                            try {
+                                const { sendPeerDataOperationRequest } =
+                                    window.require(
+                                        'WAWebSendNonMessageDataRequest',
+                                    );
+                                sendPeerDataOperationRequest(4, {
+                                    msgKeys: [msg.id],
+                                }).catch(() => {});
+                            } catch (_) {
+                                // module may not be available
+                            }
+                        }, 5000);
+                        const failTimer = setTimeout(() => {
+                            window.onCiphertextFailedEvent(
+                                window.WWebJS.getMessageModel(msg),
+                            );
+                        }, 15000);
+                        msg.once('change:type', (_msg) => {
+                            clearTimeout(resendTimer);
+                            clearTimeout(failTimer);
                             window.onAddMessageEvent(
                                 window.WWebJS.getMessageModel(_msg),
-                            ),
-                        );
+                            );
+                        });
                         window.onAddMessageCiphertextEvent(
                             window.WWebJS.getMessageModel(msg),
                         );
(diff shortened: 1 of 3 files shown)


idola9 (approved):
Great solution, approved

idolaman (approved):

izofis (approved):
The system decrypted the entire encrypted message. 56 encrypted messages were received during the day, and it decrypted all of them with 100% accuracy.

Adi1231234:
@izofis @idola9 @idolaman Thanks! 😄 
@BenyFilho Can we merge it? 🙏

IlayBacil1999 (approved):
Nice work, the gate override approach is solid. Approved

src/Client.js
```diff
@@ -1049,11 +1074,31 @@ class Client extends EventEmitter {
...
-                        msg.once('change:type', (_msg) =>
+                        const resendTimer = setTimeout(() => {
+                            try {
+                                const { sendPeerDataOperationRequest } =
+                                    window.require(
+                                        'WAWebSendNonMessageDataRequest',
+                                    );
+                                sendPeerDataOperationRequest(4, {
+                                    msgKeys: [msg.id],
+                                }).catch(() => {});
```
sofi-ans: Tested this on a live WA Web instance. The overall approach is correct - PDO type 4, { msgKeys: [...] } format, gate override, and change:type cleanup all check out.

One thing worth considering: the resendTimer callback is manually reimplementing logic that already exists inside WAWebNonMessageDataRequestPlaceholderMessageResendUtils.handlePlaceholderMsgsSeen. Calling that directly would give you three things the current code is missing:

Subtype filtering - WA skips bot_unavailable_fanout, hosted_unavailable_fanout, and view_once_unavailable_fanout. These can't be recovered so sending PDO for them is pointless.

Age check - WA only retries messages up to 14 days old (placeholder_message_resend_maximum_days_limit AB prop).

Deduplication - WA tracks in-flight requests internally so the same message doesn't get multiple PDOs sent.

```js
const resendTimer = setTimeout(() => {
    try {
        window
            .require('WAWebNonMessageDataRequestPlaceholderMessageResendUtils')
            .handlePlaceholderMsgsSeen([msg], true);
    } catch (_) {
        // module may not be available
    }
}, 5000);
```
The true second argument is the "is visible" flag the function needs. Since the gate override is already set up earlier in the evaluate block, the function will pass its own internal gate check and proceed normally.

Tested on live WA Web with a real message: the function walks through all its conditions (type, id dedup, t age, unsafe().isMdHistoryMsg, subtype) and calls sendPeerDataOperationRequest(4, { msgKeys: [...] }) with the correct MsgKey format.

sofi-ans (changes requested):

src/Client.js
Adi1231234: @sofi-ans Thanks for digging into this. The "is visible" flag was the piece I was missing when I considered calling it directly. Let me know what you think.

sofi-ans: Looks good, approved

sofi-ans (approved):

seowzhenjun0126:
Hi, after implementing your fix, this is what I got:

`cipher_message`
Total: 606
Recoverd: 535
Recover rate: 88%
Number of WhatsApp account tested: 164

Interesting observation, there are 2 WA account that received 7 and 4 `cipher_message` respectively, and failed to recover all of them.

Thanks for your work! It really helps to solve this issue for me.

izofis:
Important Finding About Decryption Success Rate Tracking
I've been tracking decryption success rates in my project and noticed something important that affects the statistics mentioned by @seowzhenjun0126:

The Issue:
The message_ciphertext_failed event is being triggered not only for actual encrypted messages that fail to decrypt, but also for WhatsApp protocol messages.

What I Found:
When tracking failed messages, I discovered they have this structure:

`{
  "type": "protocol",
  "subtype": "ephemeral_setting"
}`

What This Means:

These are NOT actual user messages that failed decryption
They are WhatsApp system/protocol messages (e.g., disappearing message settings changes)
Including them in failure statistics artificially inflates the failure rate
The Solution:
When calculating decryption success rates, these protocol messages should be filtered out:

`client.on('message_ciphertext_failed', (msg) => {
    const msgId = msg.id?._serialized || msg.id;
    
    // Filter out protocol messages - they are NOT decryption failures
    if (msg.type === 'protocol') {
        logger.info(`Protocol message (not a decryption failure): ${msgId}`);
        return; // Don't count in statistics
    }
    
    // This is an actual encrypted message that failed to decrypt
    ciphertextTracking.failed++;
    logger.error(`❌ Real decryption failure: ${msgId}`);
});`

Impact on Statistics:
After filtering out protocol messages, my actual message decryption success rate is much higher than initially calculated. The drops in success rate were caused by these protocol messages being incorrectly counted as failures.

I hope this helps clarify the decryption statistics discussion!

Adi1231234:
@izofis @seowzhenjun0126 Thanks for the reports! Found the root cause - we're calling handlePlaceholderMsgsSeen([msg], true) per message from Msg.on('add'), so every ciphertext message sends its own PDO request. The function has built-in batching (chunks of 32) but it never kicks in since we pass one message at a time.

In WhatsApp's native code, handlePlaceholderMsgsSeen is called from the ConversationMsgs component during render with all visible placeholders at once, so the batching works as intended. In our case there's no UI rendering so we need to replicate that ourselves.

In production we saw this cause most PDO responses to time out when many ciphertext messages arrive at once.

We neecd to collect ciphertext messages into a shared buffer and flush them all in a single handlePlaceholderMsgsSeen call after the grace period. 
@izofis The protocol message issue should go away with this too - protocol messages follow the exact same PDO recovery path, they just time out due to the same server overload.

sofi-ans (approved):

BenyFilho:
As I said before, this is not a good pattern for the project, and we should avoid this kind of approach.

```
} catch (_) {
    // Module may not be available in all versions
}
```

Adi1231234:
@BenyFilho Done, removed both catch (_) blocks.
One thing - if WhatsApp ever removes WAWebSyncGatingUtils, the require will now throw inside the main evaluate() block and break all event listeners, not just ciphertext recovery. That's why I originally had the try/catch there. Happy to keep it this way though if you prefer.

izofis:
Hello @Adi1231234 As far as I understand, @BenyFilho isn't telling you to remove try catch {}, he's referring to the fact that the catch block returns empty. You should provide a return message, for example.

BenyFilho:
> @BenyFilho Done, removed both catch (_) blocks. One thing - if WhatsApp ever removes WAWebSyncGatingUtils, the require will now throw inside the main evaluate() block and break all event listeners, not just ciphertext recovery. That's why I originally had the try/catch there. Happy to keep it this way though if you prefer.

But you understand that broken everything is better than kill silent, because we will know the exact place to look at

izofis (approved):

Adi1231234:
@BenyFilho Understood, thanks 🙏

Adi1231234:
Thank you very much for approving the PRs!! @BenyFilho 
I wanted to ask if there’s anything else you think should be addressed in this PR so we can merge it.

BenyFilho:
> Thank you very much for approving the PRs!! @BenyFilho I wanted to ask if there’s anything else you think should be addressed in this PR so we can merge it.

Only waiting for the community to validate it, and I asked users from Discord to test it, but you know, sometimes it's hard, people would rather complain to help

Adi1231234:
> > Thank you very much for approving the PRs!! @BenyFilho I wanted to ask if there’s anything else you think should be addressed in this PR so we can merge it.
> 
> Only waiting for the community to validate it, and I asked users from Discord to test it, but you know, sometimes it's hard, people would rather complain to help

Yeah, I totally understand you. Although I see that a few people already approved the PR, I assume that means they tested it?
@izofis @idolaman @IlayBacil1999 @idola9 @sofi-ans Did you test it?

sofi-ans:
@Adi1231234 Great PR! I've been running a fork with the same recovery mechanism in production and kept seeing false message_ciphertext_failed events. Turned out they were all unavailable fanout placeholders.

Dug into WhatsApp Web's source - `handlePlaceholderMsgsSeen` explicitly skips three subtypes before sending the PDO:
- `view_once_unavailable_fanout`
- `hosted_unavailable_fanout`
- `bot_unavailable_fanout`

These are unrecoverable by design - the server won't resend them. So the PDO gets silently dropped and the 15s failTimer fires as a false positive.

Adi1231234:
> @Adi1231234 Great PR! I've been running a fork with the same recovery mechanism in production and kept seeing false message_ciphertext_failed events. Turned out they were all unavailable fanout placeholders.
> 
> Dug into WhatsApp Web's source - `handlePlaceholderMsgsSeen` explicitly skips three subtypes before sending the PDO:
> 
> * `view_once_unavailable_fanout`
> * `hosted_unavailable_fanout`
> * `bot_unavailable_fanout`
> 
> These are unrecoverable by design - the server won't resend them. So the PDO gets silently dropped and the 15s failTimer fires as a false positive.

@sofi-ans Great catch, thank you! Just pushed the fix.

sofi-ans:
> > @Adi1231234 Great PR! I've been running a fork with the same recovery mechanism in production and kept seeing false message_ciphertext_failed events. Turned out they were all unavailable fanout placeholders.
> > Dug into WhatsApp Web's source - `handlePlaceholderMsgsSeen` explicitly skips three subtypes before sending the PDO:
> > 
> > * `view_once_unavailable_fanout`
> > * `hosted_unavailable_fanout`
> > * `bot_unavailable_fanout`
> > 
> > These are unrecoverable by design - the server won't resend them. So the PDO gets silently dropped and the 15s failTimer fires as a false positive.
> 
> @sofi-ans Great catch, thank you! Just pushed the fix.

That works great for me.
Thanks @Adi1231234

sofi-ans (approved):

dordor12:
works for me as well

idolaman:
Using it for a while, it works really well

IlayBacil1999 (approved):

IlayBacil1999:
I’ve been running this over the past few days and it’s been working flawlessly! It successfully recovered all the messages that were received as ciphertext. Thanks so much! @Adi1231234

aliyss (approved):

BenyFilho (approved):

PR #127090, opened 2026-03-02
fix: prevent stale change:ref listener from triggering false QR disconnect

Adi1231234:
Fixes #127089

## Problem

Two related bugs in the QR handling code:

1. The `change:ref` listener on `AuthStore.Conn` is never removed after a successful QR scan. When WhatsApp Web clears the ref during/after authentication (setting it to null/undefined), the listener fires `onQRChangedEvent` with an invalid ref, incrementing `qrRetries` past `qrMaxRetries` and triggering a spurious `"Max qrcode retries reached"` disconnect while authentication is already in progress.

2. The `qrRetries` counter never resets. Pre-scan QR refreshes consume retry attempts, so any post-scan QR event (even legitimate) immediately hits the limit. If linking fails and WA Web falls back to QR mode, the user gets zero retry attempts.

## Changes

Three layered fixes in `src/Client.js`, each addressing a different aspect:

1. **Reset `qrRetries` on `loading_screen`**: When the loading screen fires (QR was successfully scanned and authentication begins), reset the counter to 0. If the linking fails and WA Web falls back to QR, the user gets a fresh set of retry attempts.

2. **Null guard on `change:ref`**: Skip `onQRChangedEvent` when ref is null/undefined. Prevents malformed QR strings (starting with `"undefined,"`) from being emitted and counted as retry attempts.

3. **Listener cleanup on `change:hasSynced`**: Remove the `change:ref` listener when authentication completes (`change:hasSynced` on `AuthStore.AppState`, which is the same object as `WAWebSocketModel.Socket`). Prevents any stale QR events from firing after the client is fully connected.

## How it was found

Production logs showed a client reaching loading 95-99% after a successful QR scan, then disconnecting 3 minutes later. The final QR event had `qrLength=146` (vs normal `239`), confirming the ref was the literal string `"undefined"`. The `qrRetries` counter (already at 4 from normal pre-scan QR refreshes) was pushed to 5, exceeding `qrMaxRetries=4`.

PR #127083, opened 2026-02-28
fix: make inject() resilient to page navigation during initialization

Adi1231234:
## Summary

Fixes #127082

`inject()` uses `page.evaluate()` in polling loops to wait for `window.Debug?.VERSION` and `window.Store`. After sleep/resume, Chrome may perform an internal page navigation (IndexedDB recovery) that destroys the execution context, causing `page.evaluate()` to throw `"Execution context was destroyed"`.

This PR:
- Replaces the two manual `page.evaluate()` polling loops with `page.waitForFunction()`, which natively survives execution context destruction
- Moves the `framenavigated` listener registration to before `inject()` in `initialize()`, so the recovery mechanism is in place from the start

No retry logic, no try/catch wrapping, no error swallowing. Uses the correct Puppeteer API that handles execution context lifecycle internally.

### `src/Client.js`

1. **First polling loop (auth timeout):** Replace manual `while` loop with `page.evaluate()` -> `page.waitForFunction('window.Debug?.VERSION != undefined', { timeout })`

2. **Second polling loop (ready timeout):** Replace manual `while` loop with `page.evaluate()` -> `page.waitForFunction('window.Store != undefined', { timeout: 30000 })`

3. **`initialize()` ordering:** Move `this.pupPage.on('framenavigated', ...)` registration to before `await this.inject()` so that if inject fails, the framenavigated handler is already in place for recovery

### Why `waitForFunction` works

Puppeteer's `WaitTask` treats `"Execution context was destroyed"` as a non-fatal error (`getBadError()` returns `undefined`). When a new context is created after navigation, `IsolatedWorld` calls `taskManager.rerunAll()` to re-evaluate all waiting tasks in the new context. This is the intended Puppeteer mechanism for handling navigation during waits.

Verified in both Puppeteer 18.x (used by wwjs) and 24.x.

## Test plan

- [x] E2E tests proving `page.evaluate()` throws during navigation (the bug)
- [x] E2E tests proving `page.waitForFunction()` survives navigation (the fix)
- [x] A/B comparison tests (old code vs new code) with HTTP server
- [x] Full flow tests combining `framenavigated` + `waitForFunction`
- [x] Sanity tests confirming no regression when there is no navigation
- [x] All 11 tests passing

PR #201653, opened 2026-03-12
fix: prevent duplicate ready events on SPA re-injection

Adi1231234:
Consolidates #127083 and #127090 into one PR with a few more fixes on top.

## Root cause

`framenavigated` fires for iframes too (URLs like `:` and `chrome-error://chromewebdata/`), not just the main frame. Since inject() runs on every `framenavigated` without filtering, you get concurrent inject() calls racing each other, Backbone listeners stacking up (3-4 sets after a single QR scan), and eventually duplicate `ready` events.

I hit this in production - after a QR scan + destroy + re-init cycle, I was getting 3x READY events. In another case READY took ~10s (normally ~2s) because 3 concurrent LoadUtils calls were fighting over resources.

## What changed

Only `src/Client.js` is modified. The diff looks big because the `try/finally` guard re-indents the inject() body - use "Hide whitespace" for a cleaner view.

**The main fix:**
- Filter `framenavigated` to main frame only (`frame.parentFrame() !== null` -> skip). This is what actually solves the problem.

**Preventing re-injection when unnecessary:**
- Check if `WWebJS` is already loaded before re-injecting on navigation. Skips inject() on SPA-internal navigations where the store is still fine.
- `_injectInProgress` boolean guard so inject() can't run concurrently (safety net).

**Listener cleanup:**
- Store Backbone listener refs in `window._wwjsListeners`, `.off()` them before registering new ones on re-inject. This is what prevents the accumulation that causes duplicate READY.

**QR fixes (from #127090):**
- Null guard on `change:ref` - skip `onQRChangedEvent` when ref is null/undefined (was generating malformed QR strings starting with `"undefined,"`)
- Reset `qrRetries` when loading screen appears, so if linking fails and WA falls back to QR mode you get fresh retry attempts
- Remove `change:ref` listener once `change:hasSynced` fires (no more stale QR events after auth)

**Other:**
- try/catch on inject() from framenavigated (prevents unhandled rejection on detached frames during destroy)
- After registering `change:hasSynced` listener, check if already true and trigger manually (race condition where state transitions before listener is registered)
- Replace `while` + `sleep(200)` polling with `waitForFunction` for Debug.VERSION / socket state / WWebJS readiness

## How I tested

Ran the same 5 scenarios on both upstream and this branch with diagnostic logging:
- initialize() with existing session
- QR scan flow
- authenticated then destroy()
- QR scan after destroy()
- remote logout from phone then QR scan

Upstream had multiple inject() calls per session, listener accumulation, and duplicate READY in the destroy/re-init case. With this branch every scenario produces exactly 1 inject, 1 listener set, 1 READY. The ~10s READY issue is also gone.

All changes use upstream APIs only (WAWebSocketModel.Socket, WAWebCmd.Cmd).


Adi1231234:
Hey @BenyFilho, @williamcostaramos, @tuyuribr!

I consolidated my two previous PRs (#127083 and #127090) into this single, more comprehensive fix. After your great feedback on those PRs, I went back and did a deep dive into the root cause - turns out the core issue was `framenavigated` firing for iframes, not just the main frame.

On top of the original fixes, I added several more improvements (listener cleanup, concurrency guard, atomic hasSynced check, qrRetries reset) and ran thorough A/B diagnostic testing across 5 different real-world scenarios to make sure everything works correctly. The full results are in the PR description.

Note: the `try/finally` concurrency guard re-indents the `inject()` body, so the diff looks bigger than it is - "Hide whitespace" gives a cleaner view.

Would love it if you could take a look when you get a chance. Any feedback is very welcome!

Thanks :)

Adi1231234:
Hey @BenyFilho , thanks for the feedback on the previous PRs - I really appreciate you taking the time.

Quick note - yeah, I used AI to help write the PR description and for some of the investigation/research. Honestly who doesn't at this point. But the actual code changes are mine, and I personally ran all the testing scenarios on real WhatsApp sessions to validate everything works.

Either way, this is a real problem that I've been dealing with in production for a while now, and I put a lot of work into tracking down the root cause and fixing it properly. I also cleaned up the description to be more concise.

Would love to get your thoughts on the code whenever you have time. Happy to rework anything if needed!
Thanks!

Adi1231234:
This PR also fixes #201663.

The `onAuthAppStateChangedEvent` callback is registered via `exposeFunction`, so it runs in Node.js, not the browser. Calling `window.require()` there throws `ReferenceError: window is not defined`, which means the QR was never refreshed on UNPAIRED_IDLE. Wrapped the call in `pupPage.evaluate()` to fix it.

BenyFilho:
Please review your try/catch blocks and check if they are really necessary. If they are needed, the catch should throw the error instead of failing silently.

Adi1231234:
Hey @BenyFilho, thanks! Went through all the try/catch blocks:

Removed .off() catch - Backbone's .off() never throws. Verified on a live WhatsApp Web session.

Removed storeAvailable catch - evaluate() can't fail after framenavigated. Puppeteer automatically waits for the new execution context via waitForExecutionContext(), and the main frame is never disposed during navigation.

Removed inject() catch - matches upstream. If inject fails, the error should surface.

Kept .catch(() => { throw 'auth/ready timeout' }) (same error format as upstream) and .catch((_) => _) on waitForNavigation (already in upstream).

N0N4M3BNS:
On my side, some connection is still stuck at AUTHENTICATED.

tuyuribr (approved):

BenyFilho (approved):

estudiochapunov:
Field validation on Linux with `LocalAuth`, Node.js 24.14.0, Puppeteer 24.38.0, and its bundled Chrome for Testing 146.

During a fresh QR pairing with whatsapp-web.js 1.34.7 plus only the earlier navigation workaround, the client emitted `READY`, but WhatsApp Web performed an unsolicited top-level reload about five minutes later. During that transition we observed `Execution context was destroyed` and temporary Store / `getContact` failures, followed by a second `READY` roughly 24 seconds later.

The same delayed reload and cache-loading sequence was independently visible in a normal WebKitGTK browser session: the chat UI became usable, then WhatsApp Web refreshed itself, displayed its cache-loading progress screen, and only stabilized afterward. This suggests that the delayed navigation is part of the current WhatsApp Web lifecycle rather than a Puppeteer-only navigation.

After pinning merged commit `1780711` from this PR, a restored `LocalAuth` session produced exactly one `READY`, completed chat, contact, and message reads, and remained free of execution-context and listener errors for more than six minutes. We also applied #201848 for the `_serialized` to `$1` change; `getChats()` and message-model reads then completed without the opaque `r: r` failure.

Important limitation: we have not yet repeated the fresh QR flow with `1780711` in an isolated profile. Therefore, this validates the restored-session path and provides evidence about the cold-pairing lifecycle, but it should not be treated as proof that the cold path is fully fixed. We plan to test that separately without replacing the working profile.

Thank you for this fix. The main-frame filter, abortable injection, and listener cleanup match the failure sequence we observed in the field.

PR #201671, opened 2026-03-18
fix: use WA's own LID resolution for block/unblock

Adi1231234:
## Summary

block() and unblock() have two issues:

- **block()** passes a Chat object to `blockContact()`, which expects a Contact model. It works by accident because both have `.id`, but it's the wrong type. The `getChat()` call also creates an unnecessary chat as a side effect via `findOrCreateLatestChat`.
- **unblock()** uses the synchronous `Contact.get()` which returns `undefined` when the contact isn't in the local cache yet.

With blocklist migration enabled (`isBlocklistMigrated() === true`), `blockContact()` internally calls `S()` which needs either a LID or an existing chat with `accountLid` to resolve the target. The old `getChat()` workaround created a chat to satisfy this, but the proper way is to resolve the contact's LID directly.

This PR uses `WAWebBlockContactUtils.getContactToBlockOnlyUseIfNoAssociatedChat()`, which is WA's own utility for this exact scenario. It calls `getCurrentLidContact()` to convert PN contacts to their LID equivalent without creating unnecessary chats.

Tested in WhatsApp Web DevTools: block + unblock cycle works correctly for contacts with and without existing chats.

## Test plan
- [ ] Block a contact you have an existing chat with
- [ ] Block a contact you have never interacted with before
- [ ] Unblock both contacts
- [ ] Verify no unnecessary chats are created

### DIFF
--- a/src/structures/Contact.js
+++ b/src/structures/Contact.js
@@ -155,10 +155,18 @@ class Contact extends Base {
         if (this.isGroup) return false;
 
         await this.client.pupPage.evaluate(async (contactId) => {
-            const chat = await window.WWebJS.getChat(contactId);
+            const contact = await window
+                .require('WAWebCollections')
+                .Contact.find(contactId);
+            const resolved = window
+                .require('WAWebBlockContactUtils')
+                .getContactToBlockOnlyUseIfNoAssociatedChat(
+                    contact,
+                    'ChatListBlock',
+                );
             await window
                 .require('WAWebBlockContactAction')
-                .blockContact({ contact: chat });
+                .blockContact({ contact: resolved });
         }, this.id._serialized);
 
         this.isBlocked = true;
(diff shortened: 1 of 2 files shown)


BenyFilho (approved):

lindionez:
This resolved contact.unblock() for me, which hadn't been working for about 15 days. However, the contact still marks isBlocked: false even when the contact is blocked.

It was tested on 4 different numbers running the code.

Adi1231234:
@BenyFilho Thanks for the approval!
@lindionez Thanks for reporting this! That bug was actually there before my changes but I went ahead and fixed it anyway. The blocklist stores LIDs so when you look up a phone number contact it never finds it. I'm now using getAlternateUserWid() to check the right ID against the blocklist, which is how WA handles it internally. 
Would love to hear if it works on your end!

lindionez (approved):
THANK YOU VERY MUCH, I had been without this function for about 2 weeks, it worked perfectly.

aliyss (approved):

PR #201681, opened 2026-03-27
fix: RemoteAuth use consistent session key and prefer local session

lucas-campagna:
This patch makes RemoteAuth use a consistent session key when saving/extracting/checking sessions against remote stores, and prefers an existing local userDataDir instead of removing it. This fixes cases where a session uploaded to remote storage couldn't be found on restart due to mismatched session identifiers.\n\nChanges: use path.join(dataPath, sessionName) as the session key for store.save/sessionExists/extract/delete; prefer local userDataDir if present; add logging.

### DIFF
--- a/src/authStrategies/RemoteAuth.js
+++ b/src/authStrategies/RemoteAuth.js
@@ -66,10 +66,8 @@ class RemoteAuth extends BaseAuthStrategy {
 
     async beforeBrowserInitialized() {
         const puppeteerOpts = this.client.options.puppeteer;
-        const sessionDirName = this.clientId
-            ? `RemoteAuth-${this.clientId}`
-            : 'RemoteAuth';
-        const dirPath = path.join(this.dataPath, sessionDirName);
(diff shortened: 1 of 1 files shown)


src/authStrategies/RemoteAuth.js
```diff
@@ -166,35 +162,27 @@ class RemoteAuth extends BaseAuthStrategy {
...
+        // If a local user data dir already exists, prefer it over remote extraction
         if (pathExists) {
-            await fs.promises
-                .rm(this.userDataDir, {
-                    recursive: true,
-                    force: true,
-                    maxRetries: this.rmMaxRetries,
-                })
-                .catch(() => {});
+            console.log('[REMOTE AUTH] Local session found at:', this.userDataDir);
```
Adi1231234: Why do you `console.log` here?

Adi1231234 (changes requested):

lucas-campagna:
Updated fork branch with additional safety: ensure dataPath exists before extract and prefer local session. Please review.

PR #201697, opened 2026-04-06
feat: support big file downloads (read from WA cache + streaming)

Adi1231234:
## Description

`downloadMedia` re-downloads and decrypts media from WhatsApp's CDN on every call, even though the data is already sitting decrypted in the browser's memory. It also loads the entire file into a single base64 string, so a 25MB PDF becomes ~33MB in one allocation with no way to process it in chunks.

This PR fixes both problems:

**Read from WhatsApp's own caches instead of re-downloading**

When WhatsApp Web resolves a media message, it stores the result in `InMemoryMediaBlobCache` (250MB LRU pool, no per-file limit) and `mediaObject.mediaBlob`. This PR reads directly from those caches instead of going through `downloadAndMaybeDecrypt` and the CDN. For the `mediaBlob` path, `OpaqueData.forceToBlob()` returns the internal Blob directly without copying.

**Add `downloadMediaStream()` for large files**

Returns a Node.js `Readable` stream that reads the Blob in chunks via `evaluateHandle` + `blob.slice()`, so the full file never sits in Node.js memory at once. Chunk size defaults to 10MB and is configurable.

The cache-reading logic is extracted into a shared `window.WWebJS.resolveMediaBlob(msgId)` utility used by both methods.

Other changes: added `*.d.ts` to `.eslintignore`, added TypeScript interfaces for the new API, added `!streamdownload` example command, fixed `downloadMedia` return type to include `| undefined`.

### Test Details

Tested with images, videos, documents and audio messages across multiple accounts. Verified that cached media is read correctly and that streaming works for files up to 924MB. Confirmed `forceToBlob()` returns the internal Blob without copying via memory profiling.

### Environment

- Machine OS: Windows 11
- Phone OS: Android 14
- Library Version: 1.34.6
- WhatsApp Web Version: 2.3000.1036713926
- Browser Type and Version: Chromium 138.0.7204.251
- Node Version: 22.20.0

## Type of Change

## Checklist

- [x] Typings (e.g. `index.d.ts`) have been updated if necessary.
- [x] Usage examples (e.g. `example.js`) / documentation have been updated if applicable.


jeepch (changes requested):
Heads up: if evaluate/evaluateHandle on resultHandle throws, dispose() won't run and the blob stays stuck in the browser's memory until the app restarts (CDP keeps the inspector wrapper alive, which pins the underlying object). Worth wrapping in try/finally.

Adi1231234:
@jeepch Thanks! I fixed it 🙏

jeepch (approved):

idolaman:
I worked with something like this using CDP before, but this implementation is simpler. Nice work

liamiron (approved):
I had the same issue and also tried using CDP for it at first.
This PR worked better for me because it avoids the browser download flow and reads the resolved media in chunks instead.
I tested it on many small and large files over the last few days, and it worked consistently.

aliyss (approved):

BenyFilho (approved):

PR #201704, opened 2026-04-07
ci: add PR body validation against template

Adi1231234:
## Description

I had a PR closed because I didn't follow the template, and I noticed this happens a lot to other contributors too. Instead of maintainers having to check each PR manually and close the ones that don't comply, this adds an automated CI check that catches it right away.

It reads the template file at runtime, so if you ever change the template the check just works - nothing hardcoded.

It catches:
- Missing sections
- Subsections left empty (just the HTML comment placeholder, nothing else)
- Environment fields not filled in
- Checkbox sections where nothing was selected

Zero dependencies, just a few lines of JS inside actions/github-script.

### Test Details

Tested on my fork with both valid and invalid PR descriptions. Also ran 14 edge cases locally - empty body, raw template copy-pasted without changes, CRLF line endings, uppercase [X], fields with colons in the value, sections in a different order, multiline HTML comments, etc. All good.

### Environment

- Machine OS: Windows 11
- Phone OS: Android 14
- Library Version: 1.34.6
- WhatsApp Web Version: 2.3000.1036713926
- Browser Type and Version: Chromium 138.0.7204.251
- Node Version: 22.20.0

## Type of Change

## Checklist

- [ ] Typings (e.g. `index.d.ts`) have been updated if necessary.
- [ ] Usage examples (e.g. `example.js`) / documentation have been updated if applicable.

### DIFF
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -47,3 +47,53 @@ jobs:
                     echo "❌ PR title does not match the required format!"
                     exit 1
                   fi
+
+    validate-body:
+        name: Validate body
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/github-script@v7
+              with:
+                  script: |
+                      const tmpl = require('fs').readFileSync('.github/pull_request_template.md', 'utf8');
+                      const body = context.payload.pull_request.body || '';
+                      const stripComments = s => s.replace(/<!--[\s\S]*?-->/g, '');
+                      const cleanTmpl = stripComments(tmpl);
+                      const cleanBody = stripComments(body);
+
+                      function findMissingSections() {
+                        return [...cleanTmpl.matchAll(/^(#{2,3} .+)$/gm)]
+                          .filter(([, h]) => !body.includes(h))
+                          .map(([, h]) => `Missing: ${h}`);
+                      }
+
+                      function findEmptySubsections() {
+                        return [...cleanBody.matchAll(/^(### .+)\n([\s\S]*?)(?=\n##|$(?![\s\S]))/gm)]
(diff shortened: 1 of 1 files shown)


BenyFilho:
That was told to you before
_"You should be free when writing your PR description, as enforcing a strict structure would make it unnecessarily complicated."_

I'm closing it.

Adi1231234:
@BenyFilho Thanks for the feedback! Just want to understand - my earlier PR (#201694) was closed by another maintainer specifically because it didn't follow the template. So I'm a bit confused about what's expected. Should we follow the template or not? Happy to go either way, just want to know for future PRs.

BenyFilho:
> @BenyFilho Thanks for the feedback! Just want to understand - my earlier PR (#201694) was closed by another maintainer specifically because it didn't follow the template. So I'm a bit confused about what's expected. Should we follow the template or not? Happy to go either way, just want to know for future PRs.

Yeah, we should follow the template, but we don't want it to be so strict

Adi1231234:
> > @BenyFilho Thanks for the feedback! Just want to understand - my earlier PR (#201694) was closed by another maintainer specifically because it didn't follow the template. So I'm a bit confused about what's expected. Should we follow the template or not? Happy to go either way, just want to know for future PRs.
> 
> Yeah, we should follow the template, but we don't want it to be so strict

Thanks for clarifying. I’m not arguing for rigidity for its own sake - I’m trying to understand the actual contributor standard.

Right now the message contributors receive is mixed: in #201694 I was told to follow the template, while here I’m told strict enforcement is not desired.

If the template is optional guidance, it probably shouldn’t be used as a reason to close PRs. If some parts are mandatory, it would help to define which parts those are.

Otherwise contributors are left guessing, and enforcement becomes maintainer-dependent.

PR #201705, opened 2026-04-08
refactor(structures): optimize message loading with msgFindBefore

JuniorRibas:
## Description

Replaces `WAWebChatLoadMessages.loadEarlierMsgs` with `WAWebDBMessageFindLocal.msgFindBefore` when paginating messages until reaching `searchOptions.limit` in both `Chat` and `Channel`, resolving models through `MsgStore` when the returned entries are not yet serialized.

## Related Issue(s)

N/A

### Test Details

- `npx eslint` on the modified files: OK.
- Full `npm test` requires `WWEBJS_TEST_REMOTE_ID` (integration). Manual validation is recommended for message fetching with `limit` in chats/channels.

### Environment

- Machine OS: (fill in)
- Phone OS: (fill in)
- Library Version: 1.34.6
- WhatsApp Web Version: (fill in — `await client.getWWebVersion()`)
- Browser Type and Version: (fill in)
- Node Version: (fill in)

## Type of Change

- [ ] **Dependency change**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Breaking change**
- [ ] **Non-code change**
- [x] **Refactor / performance** (internal message loading)

## Checklist

- [x] My code follows the style guidelines of this project.
- [ ] All new and existing tests pass (`npm test`). _(integration suite requires environment variables)_
- [x] Typings have been updated if necessary. _(no public API changes)_
- [ ] Usage examples / documentation have been updated if applicable.

---

Closes https://github.com/wwebjs/whatsapp-web.js/issues/201706

Closes #201706

PR #201711, opened 2026-04-12
fix(structures): fix fetchMessages pagination and dropped in-memory msgs

themazim:
## Description

Follow-up to #201705. Builds on that refactor and fixes two cases where `Chat.fetchMessages` / `Channel.fetchMessages` could return fewer messages than requested — in particular, an empty array or a single anchor message where the caller expected up to `limit` messages.

Changes (both `src/structures/Chat.js` and `src/structures/Channel.js`, in the `!fromMeFilter && finite` fast-path introduced by #201705):

1. **Include in-memory `msgs` when building `merged`.**
   `loaded` only contains what `msgFindBefore` returned from the local IndexedDB message table. Messages that are in the in-memory `MsgCollection` but not yet persisted (or not returned by the range query) were being dropped. Appending `...msgs` into `merged` preserves them; the existing `dedupeByMsgId` step guarantees no duplicates.

   Symptom before the fix: chats whose requested window consisted mostly of recent in-memory messages (with nothing older resolvable via `msgFindBefore`) returned only the anchor message — reproduced and confirmed in the #201705 discussion thread.

2. **Preserve `msgs` on `msgFindBefore` `status: 404`.**
   When the anchor is not present in the local DB (`NoAnchorMessageError` → `status: 404`), the fast-path previously set `msgs = []`, throwing away the already-filtered in-memory messages. It now mirrors the `!anchorSerialized` fallback immediately above: sort by `t` and slice to `limit`. Rare in practice (`lastReceivedKey` is almost always present in the DB), but consistent with how the "cannot run the DB query" case is handled a few lines above.

No public API or type changes.

## Related Issue(s)

Builds on #201705.

### Test Details

- `npx eslint src/structures/Chat.js src/structures/Channel.js` — clean.
- Manual verification against a live WhatsApp Web session for both `Chat` and `Channel`:
  - `chat.fetchMessages({ limit: 50 })` on a chat with a large stored history → returns ~50 messages (regression guard for the single-message bug).
  - `chat.fetchMessages({ limit: 50 })` on a chat whose history mostly lives in the in-memory collection → now returns the in-memory messages instead of one or zero.
  - Same two scenarios on a channel via `channel.fetchMessages({ limit: 50 })`.

## Type of Change

- [ ] **Dependency change**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Breaking change**
- [ ] **Non-code change**

## Checklist

- [ ] All new and existing tests pass (`npm test`). _(integration suite requires `WWEBJS_TEST_REMOTE_ID`; eslint on the touched files is clean)_
- [x] Typings (e.g. `index.d.ts`) have been updated if necessary. _(no public API changes)_
- [ ] Usage examples (e.g. `example.js`) / documentation have been updated if applicable. _(not applicable)_

### DIFF
--- a/src/structures/Channel.js
+++ b/src/structures/Channel.js
@@ -356,6 +356,7 @@ class Channel extends Base {
                     const WAWebMsgKey = window.require('WAWebMsgKey');
                     const MsgStore = window.require('WAWebCollections').Msg;
 
+                    // msgFindByDirection is the newer API; fall back to msgFindBefore on older WA Web versions
                     const findBefore = async (anchorKey, count) => {
                         if (
                             typeof msgFindLocal.msgFindByDirection ===
@@ -445,7 +446,11 @@ class Channel extends Base {
                                 result?.status === 404 &&
                                 (!rawMessages || !rawMessages.length)
                             ) {
-                                msgs = [];
+                                // anchor not in local DB — fall back to in-memory msgs (same as the !anchorSerialized branch above)
+                                msgs.sort((a, b) => (a.t > b.t ? 1 : -1));
+                                msgs = msgs.slice(
+                                    -Math.min(limit, msgs.length),
+                                );
                             } else {
                                 let loaded = toMsgModels(rawMessages);
                                 const anchorMsg =
@@ -453,6 +458,7 @@ class Channel extends Base {
                                 let merged = [
                                     ...loaded,
                                     ...(anchorMsg ? [anchorMsg] : []),
+                                    // include in-memory msgs so recent (not-yet-persisted) messages aren't dropped
                                     ...msgs,
                                 ];
                                 merged = merged.filter(
--- a/src/structures/Channel.js
+++ b/src/structures/Channel.js
@@ -350,17 +350,170 @@ class Channel extends Base {
                 let msgs = channel.msgs.getModelsArray().filter(msgFilter);
 
                 if (searchOptions && searchOptions.limit > 0) {
-                    while (msgs.length < searchOptions.limit) {
-                        const loadedMessages = await window
-                            .require('WAWebChatLoadMessages')
-                            .loadEarlierMsgs(channel);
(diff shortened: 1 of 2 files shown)


mohamedhossam01:
I had the same issue mentioned here https://github.com/wwebjs/whatsapp-web.js/issues/201706 and followed your solution. Can confirm it's working as expected and I'm able to fetch messages again

N0N4M3BNS:
It's working perfectly for me. Thanks

BenyFilho:
What is the difference between https://github.com/wwebjs/whatsapp-web.js/pull/201705 and that one?
Is only that?

```
let merged = [
     ...loaded,
     ...(anchorMsg ? [anchorMsg] : []),
    // include in-memory msgs so recent (not-yet-persisted) messages aren't dropped
    ...msgs,
];
```

themazim:
> What is the difference between #201705 and that one? Is only that?
> 
> ```
> let merged = [
>      ...loaded,
>      ...(anchorMsg ? [anchorMsg] : []),
>     // include in-memory msgs so recent (not-yet-persisted) messages aren't dropped
>     ...msgs,
> ];
> ```


yes, and this:

```
                                // anchor not in local DB — fall back to in-memory msgs (same as the !anchorSerialized branch above)
                                msgs.sort((a, b) => (a.t > b.t ? 1 : -1));
                                msgs = msgs.slice(
                                    -Math.min(limit, msgs.length),
                                );

```

The initial PR (https://github.com/wwebjs/whatsapp-web.js/pull/201705) seems to be highly over engineered regarding the initial issue., now that I have dug way deeper into this. 

Simple method signature change would probably fix this as well.

BenyFilho:
If the difference is only that, make your suggestions in the original PR

themazim:
> If the difference is only that, make your suggestions in the original PR

as expected, the signature just changed. https://github.com/wwebjs/whatsapp-web.js/pull/201713

Issue #201717, opened 2026-04-15
Error: asUserWidOrThrow: wid is not a user wid when calling getContacts()


elhumbertoz:
## Issue Description
Calling `client.getContacts()` fails with an uncaught error: `asUserWidOrThrow: wid is not a user wid`. This seems to happen when the contact list contains non-user WIDs (like group JIDs) or new Linked ID (`@lid`) identifiers that the internal WhatsApp `asUserWidOrThrow `function does not validate as standard users.

## Reproduction Steps
1. Initialize the client and wait for the ready event.
2. Call await client.getContacts().
3. The process crashes with the following stack trace:

```
Error: asUserWidOrThrow: wid is not a user wid
    at s (hqMvLvPUOL4.js:79:180)
    at Object.I [as asUserWidOrThrow] (6ZCeJrDqvpN7oo8EPfVEEipN3BCR3yxG_Wl_HC4ZDPfu.js:910:3080)
    at window.WWebJS.getContactModel (pptr:evaluate;executeReadyLogic (Client.js:417:36):1030:26)
    at window.WWebJS.getContacts (pptr:evaluate;executeReadyLogic (Client.js:417:36):1082:22)
```

## Code Sample
```typescript
const contacts = await client.getContacts();
console.log(contacts);
```

## User Setup

| WhatsApp | Type |
|---|---|
| **Account Type** | WhatsApp Business |
| **Authentication Strategy** | LocalAuth |
| **WhatsApp Web Version** | 2.3000.1037376462 |
| **whatsapp-web.js Version** | 1.34.6 |

| Environment | Version |
|---|---|
| **Browser Type** | Google Chrome |
| **Browser Version** | 146.0.7680.76 (Compilación oficial) (64 bits) |
| **Phone OS Version** | Android 12 |
| **Running OS Version** | Windows / Ubuntu 24 |
| **Node.js Version** | v22.14.0 |

## Checklist
- [x] **I use the latest released version of whatsapp-web.js.**
- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have verified this is a library issue, not a problem with my code.

elhumbertoz:
[...]

PR #201718, opened 2026-04-15
fix: Error asUserWidOrThrow calling client.getContacts()

elhumbertoz:
Calling client.getContacts() fails with an uncaught error: asUserWidOrThrow: wid is not a user wid. 

## Description

Calling client.getContacts() fails with an uncaught error: asUserWidOrThrow: wid is not a user wid. This seems to happen when the contact list contains non-user WIDs (like group JIDs) or new Linked ID (@lid) identifiers that the internal WhatsApp asUserWidOrThrow function does not validate as standard users.

## Related Issue(s)

#201717

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -433,13 +433,13 @@ exports.LoadUtils = () => {
                     : meUser;
             participant = window
                 .require('WAWebWidFactory')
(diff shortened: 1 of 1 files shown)


BenyFilho:
Your changes are not solving the matter, because you changed the method in sendMessage, but you need to do it in getContactModel

PR #201721, opened 2026-04-19
fix(types): make Client constructor options optional

lumaXs:
## Fix: Client constructor options should be optional

The current type definition requires an argument:

```ts
constructor(options: ClientOptions)
```

However, the runtime allows instantiating without arguments:

```ts
new Client()
```

This results in a TypeScript error:

> Expected 1 arguments, but got 0. [2554]

## Proposed change

Make the options parameter optional:

```ts
constructor(options?: ClientOptions)
```

## Impact

- Aligns type definitions with runtime behavior
- Fixes incorrect TypeScript error
- Improves developer experience

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ import InterfaceController from './src/util/InterfaceController';
 
 declare namespace WAWebJS {
     export class Client extends EventEmitter {
-        constructor(options: ClientOptions);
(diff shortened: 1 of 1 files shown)


BenyFilho:
Currently, the lib is not 100% compatible with Typescript, maybe in the future

PR #201730, opened 2026-04-27
feat: Member tag

Davdadev:
feat: add group member tag support (setMemberTag, deleteMemberTag, getMemberTag)
Description
Adds support for WhatsApp's group member tag feature, which was rolled out in January 2026. Member tags allow users to set a custom label that appears under their name in a group chat and participant list (e.g. "Coach", "Organiser", "Project Manager"). Tags are group-specific, so a user can have different tags in different groups.
Changes

GroupChat.js — added three new methods:

setMemberTag(label) — sets or updates your member tag in the group (max 30 chars, plain text only)
deleteMemberTag() — removes your member tag from the group
getMemberTag() — returns your current member tag as a string, or null if not set

index.d.ts — added TypeScript type definitions for all three methods

How it works
Uses WAWebSendMemberLabelAction.sendMemberLabelMsg and WAWebMemberLabelsFrontendUtils.getMyMemberLabelStringForChat, both accessed via window.require() following the existing pattern in Utils.js. No changes to AuthStore.js or any other files were needed.
Deleting a tag is achieved by sending an empty string, matching WhatsApp Web's internal behaviour.
Testing
Manually verified in WhatsApp Web console:

sendMemberLabelMsg returns {messageSendResult: 'OK'} ✅
Tag appears under name in group after calling setMemberTag ✅
getMyMemberLabelStringForChat returns the correct string ✅
Empty string call successfully removes the tag ✅

Notes

Tags are capped at 30 characters and plain text only (no symbols, links, or special characters) — validated before calling the API
This feature requires WhatsApp Web to have rolled out member tags to the account — older sessions may not have access yet

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -38,6 +38,12 @@ declare namespace WAWebJS {
         acceptGroupV4Invite: (
             inviteV4: InviteV4Data,
         ) => Promise<{ status: number }>;
+        /** Sets or updates your member tag in this group. Max 30 characters, plain text only. */
+        setMemberTag(label: string): Promise<boolean>;
+        /** Deletes your member tag in this group. */
+        deleteMemberTag(): Promise<boolean>;
+        /** Gets your current member tag in this group, or null if not set. */
+        getMemberTag(): Promise<string | null>;
 
         /**Returns an object with information about the invite code's group */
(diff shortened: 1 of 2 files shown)


Davdadev:
I've tested and it works

PR #201825, opened 2026-07-12
feat: add call answering, outgoing calls and audio injection

### LINKED ISSUE #201746: Reject Call don't work
## Issue Description
client.once("ready", () => {
        if (typeof client.setRejectIncomingCalls === "function") {
            client.setRejectIncomingCalls(true).catch(() => {});
        }
    });
 client.on("call", async (call) => {
        if (call.fromMe) return;
        try {
            await call.reject();
            console.log(
                `[${tenantId}] Rejected incoming ${call.isVideo ? "video" : "voice"} call from ${call.from}`,
            );
        } catch (e) {
            console.error(`[${tenantId}] rejectCall:`, e.message);
        }
    });

## Reproduction Steps
client.on("call", async (call) => {
        if (call.fromMe) return;
        try {
            await call.reject();
            console.log(
                `[${tenantId}] Rejected incoming ${call.isVideo ? "video" : "voice"} call from ${call.from}`,
            );
        } catch (e) {
            console.error(`[${tenantId}] rejectCall:`, e.message);
        }
    });

## Code Sample
```typescript

```

## User Setup

| WhatsApp | Type |
|---|---|
| **Account Type** | WhatsApp Business |
| **Authentication Strategy** | LocalAuth |
| **WhatsApp Web Version** | latest |
| **whatsapp-web.js Version** | 1.34.7 |

| Environment | Version |
|---|---|
| **Browser Type** | cmd |
| **Browser Version** | v |
| **Phone OS Version** | 1.34.7 |
| **Running OS Version** | windows |
| **Node.js Version** | 1.34.7 |

## Checklist
- [x] **I use the latest released version of whatsapp-web.js.**
- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have verified this is a library issue, not a problem with my code.

alpersamur3:
## Description

Adds first-class call support: answer incoming calls, place outgoing calls, and
inject audio (e.g. a text-to-speech file) into an active call. Also fixes call
handling that no longer works on current WhatsApp Web.

New `Call` methods:
- `accept()` — answer the call (audio only by default, even for incoming video calls)
- `end()` — hang up an ongoing call
- `playAudio(media)` — play a `MessageMedia`/base64 audio clip into the call so the
  other party hears it; resolves when playback finishes, so clips can be chained
- `isConnected()` — whether the other party has answered

New `Client` methods:
- `call(chatId, { video, waitForAnswer, answerTimeout })` — place an outgoing
  voice/video call; with `waitForAnswer: true` it resolves once the callee answers
- `getActiveCall()` — the ongoing call (ringing/placing/connected), or `null`

Fixes:
- Incoming `call` event: current WhatsApp Web surfaces calls through
  `WAWebCallCollection`'s `activeCall` attribute; the previous internal-map `set`
  hook no longer fires, so `client.on('call')` never triggered. Now listens to
  `change:activeCall` (the map hook is kept as a fallback).
- `Call.reject()`: the previous signaling stanza no longer rejects on current
  WhatsApp Web; it now goes through the voip stack's `rejectCall()`.

Audio is captured from a synthetic Web Audio stream (WhatsApp's `getUserMedia` is
overridden), so **no microphone is required and it works headless**. The browser
flags needed for this are added automatically.

## Related Issue(s)

fixes #201746

### Test Details

Verified end-to-end against a real phone with live calls.

Outgoing voice call + audio + hang up:

```js
const call = await client.call('1XXXXXXXXXX', { waitForAnswer: true });
if (await call.isConnected()) {
    await call.playAudio(MessageMedia.fromFilePath('./hello.wav'));
    await call.end();
}
```

Incoming call handling:

```js
client.on('call', async (call) => {
    if (call.fromMe) return;
    await call.accept();                 // or await call.reject()
    await call.playAudio(MessageMedia.fromFilePath('./welcome.wav'));
    await call.end();
});
```

Confirmed: incoming voice/video calls fire the `call` event; `accept`, `reject`,
`end`, sequential `playAudio` and `getActiveCall` all work; the callee hears the
injected audio; video calls are answered audio-only; everything also works with a
headless browser and no microphone.

Note: the repo's `npm test` integration suite requires a configured live test
account (`WWEBJS_TEST_REMOTE_ID` etc.) and only covers messaging/groups, not
calls, so verification was done manually with real calls as shown above.

### Environment

- Machine OS: Windows 11
- Phone OS: Android 16
- Library Version: 1.34.7
- WhatsApp Web Version: 2.3000.1017054665
- Browser Type and Version: Google Chrome 149
- Node Version: 22.17.0

## Type of Change

## Checklist

- [ ] All new and existing tests pass (`npm test`) — the integration suite needs a live test account and only covers messaging/groups; verified manually with real calls.
- [x] Typings (`index.d.ts`) have been updated.
- [x] Usage examples (`example.js`) / documentation have been updated.

### DIFF
--- a/example.js
+++ b/example.js
@@ -697,7 +697,24 @@ client.on('call', async (call) => {
         call.from,
         `[${call.fromMe ? 'Outgoing' : 'Incoming'}] Phone call from ${call.from}, type ${call.isGroup ? 'group' : ''} ${call.isVideo ? 'video' : 'audio'} call. ${rejectCalls ? 'This call was automatically rejected by the script.' : ''}`,
     );
+
+    // Instead of rejecting, a call can also be answered and an audio clip
+    // (e.g. a text-to-speech file you generated) can be played into it:
+    // await call.accept(); // answers with audio only, even for video calls
+    // await call.playAudio(MessageMedia.fromFilePath('./welcome.mp3'));
+    // await call.end();
 });
+
+// Placing an outgoing call and speaking once the other party answers:
+// const call = await client.call('1234567890', { waitForAnswer: true });
+// if (await call.isConnected()) {
+//     await call.playAudio(MessageMedia.fromFilePath('./message.mp3'));
+//     await call.end();
+// }
+
+// Checking for an ongoing call and hanging it up:
+// const activeCall = await client.getActiveCall();
+// if (activeCall) await activeCall.end();
 
 client.on('disconnected', (reason) => {
     console.log('Client was logged out', reason);
--- a/index.d.ts
+++ b/index.d.ts
@@ -212,10 +212,7 @@ declare namespace WAWebJS {
         ): Promise<Message>;
 
         /** Send a reaction to a specific messageId */
-        sendReaction(
-            messageId: string,
-            reaction: string,
-        ): Promise<void>;
+        sendReaction(messageId: string, reaction: string): Promise<void>;
 
         /** Sends a channel admin invitation to a user, allowing them to become an admin of the channel */
         sendChannelAdminInvite(
@@ -344,6 +341,19 @@ declare namespace WAWebJS {
 
         /** Generates a WhatsApp call link (video call or voice call) */
         createCallLink(startTime: Date, callType: string): Promise<string>;
+
+        /** Places an outgoing call to the provided chat or phone number */
+        call(
+            chatId: string,
+            options?: {
+                video?: boolean;
+                waitForAnswer?: boolean;
+                answerTimeout?: number;
+            },
+        ): Promise<Call>;
+
+        /** Gets the call that is currently ongoing (ringing, being placed or connected), if any */
+        getActiveCall(): Promise<Call | null>;
 
         /**
          * Sends a response to the scheduled event message, indicating whether a user is going to attend the event or not
@@ -2444,6 +2454,18 @@ declare namespace WAWebJS {
 
         /** Reject the call */
         reject: () => Promise<void>;
+
+        /** Accept the call (audio only by default, even for incoming video calls) */
+        accept: (options?: { video?: boolean }) => Promise<boolean>;
+
+        /** End an ongoing call */
+        end: () => Promise<boolean>;
+
+        /** Play an audio clip into the ongoing call so the other party can hear it */
+        playAudio: (media: MessageMedia | string) => Promise<number>;
+
+        /** Indicates whether the call is currently connected (the other party has answered) */
+        isConnected: () => Promise<boolean>;
     }
 
     /** Message type List */
--- a/index.d.ts
+++ b/index.d.ts
@@ -349,6 +349,7 @@ declare namespace WAWebJS {
                 video?: boolean;
                 waitForAnswer?: boolean;
                 answerTimeout?: number;
+                injectAudio?: boolean;
             },
         ): Promise<Call>;
 
@@ -2456,7 +2457,10 @@ declare namespace WAWebJS {
         reject: () => Promise<void>;
 
         /** Accept the call (audio only by default, even for incoming video calls) */
-        accept: (options?: { video?: boolean }) => Promise<boolean>;
+        accept: (options?: {
+            video?: boolean;
+            injectAudio?: boolean;
+        }) => Promise<boolean>;
 
         /** End an ongoing call */
         end: () => Promise<boolean>;
--- a/src/Client.js
+++ b/src/Client.js
@@ -1172,7 +1172,18 @@ class Client extends EventEmitter {
                 if (!window._wwjsCallListener) {
                     window._wwjsCallListener = true;
                     WAWebCallCollection.on('change:activeCall', (call) => {
-                        emitCall(call);
+                        if (call) {
+                            emitCall(call);
+                        }
+                        // Restore the real microphone once the call is over so a
+                        // following normal call is not fed the injected stream.
+                        if (
+                            !call ||
+                            (typeof call.getState === 'function' &&
+                                call.getState() === 0)
+                        ) {
+                            window.WWebJS.teardownCallMediaStream?.();
+                        }
                     });
                 }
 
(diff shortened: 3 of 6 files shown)


alpersamur3:
Heads-up: I've built video support on top of this, in a follow-up branch (`feat/call-video-injection` on my fork). It reuses the same synthetic-media approach and adds:

- `Call.showImage(media)` and `Call.playVideo(media, { loop })`
- `orientation` (`landscape` / `portrait` / `auto`) and `resolution` options on `Client.call` / `Call.accept`
- `Call.setVideoResolution(px)` to change quality mid-call

Verified live end-to-end (portrait/landscape, mid-call source + resolution changes, video + injected audio together), including headless and after a cold restart.

Once this PR is merged I'll rebase that branch onto `main` and open it as a separate PR to keep the diffs clean — happy to fold it in here instead if you'd prefer.

alpersamur3:
@aliyss @lindionez Hi! Could you please take a look at this PR when you have a moment? Thanks!

qgustavor:
Other than fixing that issue, what's the intended use cases of this? As far as I understood, one can set up an automatic reply to calls and can auto-reject calls. That's it?

If it got it right, I wonder if you can consider those suggestions:

1. Can the `getUserMedia` override be optional? As it is I'm sure it will break normal call usage in [wwebjs-electron](https://github.com/AndyTargino/wwebjs-electron).
2. Can the API expose the audio stream?
3. Can it detect when other device took the call?

One potential use-case I think of is have it checking if I took a call, if I didn't took call after some time it would auto-answer, play a recording and record the answer (like a traditional answering machine). It would avoid a really common issue when people call me, I can't answer, and they don't send any message explaining why they called me. That said, I have Whisper already set up in my code, I can set it to transcribe their answer too.

alpersamur3:
@qgustavor Thanks again for the review, this was a good catch.

**Use cases:** it goes a bit beyond auto-reply and auto-reject. You can answer, reject and end calls, place outgoing calls, and inject audio into a live call with `playAudio`, so you can play a TTS file or a pre-recorded clip to the other party (clips chain, since `playAudio` resolves when playback finishes). Your answering machine idea fits it perfectly: auto-answer after a delay and play a recording. The missing piece for the full loop is recording the caller's reply, which is your point 2.

1. **Optional `getUserMedia` override: fixed.** Injection is now opt-in through an `injectAudio` option on `call()` and `accept()` (default true), and the override only feeds the synthetic stream while an injected call is actually active. When the call ends the real microphone is restored, so a following normal call, and wwebjs-electron, is left alone. I checked it both at the routing level (with injection off, `getUserMedia` hands back the real device track again) and with live calls: an injected call plays the clip, and a following `injectAudio: false` call goes out on the real microphone with no injected audio leaking in.

2. **Exposing the audio stream:** not in this PR, but it is the natural next step and it is what your record-the-answer flow needs. It means tapping the remote WebRTC track rather than the microphone entry point. I would like to do it as a follow-up.

3. **Detecting when another device answered:** WhatsApp signals this through the call collection (the active call clears or drops to the ended state when your phone or another linked device picks up), and I already use that signal to restore the microphone. I can surface it as an event so your auto-answer logic can bail out. I will add it as a follow-up.

I also pushed a small related fix in the same PR: outgoing calls now resolve to the freshly placed call instead of a stale previous one, so `Call.isConnected()` and `waitForAnswer` stay reliable when you place several calls in a session.

Thanks again, really appreciate the thoughtful feedback.

alpersamur3:
@qgustavor quick follow-up on your suggestions, and thanks for the rocket :)

**3 (detecting when another device took the call): done.** When an incoming call is answered on one of your other linked devices, WhatsApp records it as a `call_log` message with `callOutcome: 'AcceptedElsewhere'`, so an answering-machine flow can check that and skip auto-answering. I opened #201888 for it (it exposes `callOutcome`, `callDuration` and `isVideoCall` on the message).

**2 (exposing the audio stream to record the caller): not feasible on WhatsApp Web, unfortunately.** I dug into it: the call media does not travel as WebRTC media tracks. WhatsApp sets up several `RTCPeerConnection`s but they carry no audio senders or receivers, only SCTP data channels, and the audio is encoded/decoded by a WASM engine over those data channels. So there is no remote audio track to tap. Capturing it would mean hooking the WASM audio pipeline, which is too fragile to ship as a library feature. Your Whisper setup would still work if you can get the audio some other way, but the library cannot expose it from here.

I also opened #201889 (stacked on this one) which adds video into calls, in case it is useful.

Thanks again for the thoughtful review.

MalikHamzaSaleemi:
is this functionality is done ?
and working ?

MuhdSHiBiLi:
Bro @alpersamur3 , outgoing function is not working anymore now, ? on yesterday morning that working perfetly, but now not call to the persons, 
after ur i see, i add a feature in my subscription payment management system, when payment update its send pdf to that person, and  he got a annoncement as outgoing whatsapp call, ur receipt no. ........ is amount is update successfully, pdf file of the receipt send to ur whatsapp, now that calling feature is not working anymore


/root/.pm2/logs/bot-misbahul-out.log last 15 lines:
1|bot-misb | Background template and fonts loaded successfully.
1|bot-misb | LOADING SCREEN 100 WhatsApp
1|bot-misb | AUTHENTICATED
1|bot-misb | READY
1|bot-misb | WWebVersion = 2.3000.1047824257
1|bot-misb | [API] Manual payment received for ID: 197, Amount: ₹100, Method: Cash, TxId: , By: SHiBiLi M
1|bot-misb | [ANTI-BAN] Text receipt confirmation sent to +919xxxxxx
1|bot-misb | [ANTI-BAN] PDF receipt sent to +91974xxxxxx
1|bot-misb | [DEBUG_AK] Member ID 197 akExplanation raw: "> 100×-5  (വരിസംഖ്യ × Already Paid Months)"
1|bot-misb | [DEBUG_AK] cleanAkVal: "100×-5"
1|bot-misb | [DEBUG_AK] akMatch matched: true
1|bot-misb | [DEBUG_AK] Parsed values -> sub: 100, val2: -5, val3: undefined
1|bot-misb | [DEBUG] sendDetails check - bulkMedia: false, resolvedMsg: true, keepMediaHistory: false
1|bot-misb | [CALL NOTIFICATION] Generated message for 9197xxxxx: "പ്രിയപ്പെട്ട സൈദലവി, നിങ്ങളുടെ വരിസംഖ്യയിലേക്കായി തന്ന നൂറ് രൂപ അപ്ഡേറ്റ് ചെയ്തിട്ടുണ്ട്. | റെസിപ്റ്റ് നമ്പർ എം എച്ച് എസ് എൺപത്തി നാല്, ഈ റെസിപ്റ്റിന്റെ പി ഡി എഫ് ഫയൽ നിങ്ങളുടെ വാട്സപ്പിലേക്ക് അയച്ചിട്ടുണ്ട്. നന്ദി"
1|bot-misb | [CALL NOTIFICATION] Placing call to 919xxxxx (Attempt 1)...

showing the attempt, but not placing any call after the pdf sent, can you help me to solve

MuhdSHiBiLi:
@alpersamur3 @qgustavor @MalikHamzaSaleemi @claude 
anyone still get outgoing call functionality and inject a audio to that call?

alpersamur3:
@MuhdSHiBiLi @MalikHamzaSaleemi thanks for the report, it's fixed now. WhatsApp Web recently started asking for confirmation ("Start a WhatsApp call with X?") before any call that isn't started from its own call buttons, so `client.call()` hung on that popup and the offer was never sent. I pushed 5a78ab0, which starts the call with the same entry point and trust flag the call buttons use, so no popup is shown. Re-verified live on 2.3000.1048354754: the call goes out in about 1.5s, gets answered and the injected audio plays. #201889 is rebased on top of it, and video calls work too (image, clip and a mid-call resolution change).

@BenyFilho @tuyuribr @aliyss could you take a look when you get a chance? #201889 and #201881 are both built on top of this one, so reviewing it first would unblock them too. Happy to address any feedback quickly.

PR #201832, opened 2026-07-15
fix(client): add fallback for WhatsApp id._serialized renamed to id.$1

lindionez:
WhatsApp Web changed the serialized message ID property from _serialized to $1 in its July 2026 update. Any code reading id._serialized received undefined, causing a cascade of TypeError crashes across the entire event pipeline (messages, reactions, polls, media download, quoted messages, etc.).

Changes:
- Add Base._normalizeId() static helper that copies $1 into _serialized whenever _serialized is missing, keeping all downstream code compatible
- Apply Base._normalizeId() to all structure _patch() methods that assign his.id or 	his.wid: Message, Chat, Contact, Channel, GroupNotification, Broadcast, ClientInfo
- Fix GroupNotification chatId and uthor object reads to use || \
- Fix Message rom, 	o, uthor and inviteV4 reads to use || \
- Fix Utils.js browser-side: normalize msg.id in getMessageModel, add || \ fallback to ewMsgKey (fixes sendMessage returning empty),
  lastReceivedKey, msg.id.remote, chat.id, getMaybeMePnUser(), createWid(p.jid), and userJidToUserWid() calls
- Fix Client.js browser-side evaluates: reactions sender, poll vote sender and parentMsgKey, blocked contacts list, lid/phone in getContactLidAndPhone, userJidToUserWid in getCustomerNote, es.gid in cceptInvite, and participant.wid/createGroupResult.wid in createGroup
- Fix GroupChat.js browser-side evaluates: pWid, participant duplicate check, group invite message ID, and lid/phone in emoveParticipants, promoteParticipants, demoteParticipants
- Fix Contact.js browser-side: lid in unblock()

Closes #201829
Closes #201830

PR #201839, opened 2026-07-15
fix(message): find sent message indexed under alternate wid (lid/pn)

telmedola:
## Description

`client.sendMessage()` sometimes resolves with `undefined` even though the message is actually delivered to the recipient.

**Root cause:** in `window.WWebJS.sendMessage`, sending and retrieving the sent message are two separate steps:

1. `WAWebSendMsgChatAction.addAndSendMsgToChat(chat, message)` sends the message;
2. `Msg.get(newMsgKey._serialized)` looks it up in the collection to return it.

For contacts with a lid/phone-number wid pair, WhatsApp Web can index the sent message under the chat's **sibling wid** (lid ↔ pn), so the lookup by the original key finds nothing and the function silently returns `undefined` — after the message has already gone out. Callers then treat a delivered message as failed, which typically leads to duplicate sends.

**Fix:** when the lookup by the original key fails, the sent message is now searched:

1. in the chat's own collection by the random id part of the key (`m.id.id === newMsgKey.id`), which does not depend on the wid used to build the key;
2. by a key rebuilt with the alternate wid (`WAWebApiContact.getAlternateUserWid`), and in the sibling chat's collection.

If the message still cannot be found, the lookup context (attempted key, chat id, alternate wid, last message keys of the chat) is stored in `window.WWebJS.lastSendMessageDebug` as an internal diagnostic aid, and `undefined` is returned as before. No public API change; behavior for already-working cases is unchanged (the original lookup runs first).

## Related Issue(s)

Related to #3767

### Test Details

Reproduced in production with a customer-service system sending messages to contacts whose `getNumberId()` resolves to a `@lid` wid: the recipient received the message, but `sendMessage` resolved with `undefined` (`TypeError` in our caller when reading `response.id`). With this patch, `sendMessage` returns the proper `Message` object for those contacts; text and media sends to regular `@c.us` contacts were also re-tested and behave as before. <-- ajuste conforme sua validação do novo build

### Environment

- Machine OS: Debian 12 (Docker, node:20-bookworm-slim)
- Library Version: 1.34.7
- WhatsApp Web Version: 2.3000.1043191242
- Node Version: 20.20.0

## Type of Change

- [ ] **New feature** _(non-breaking change that adds functionality)_

## Checklist

- [x] Typings (e.g. `index.d.ts`) have been updated if necessary. _(no public API change)_
- [x] Usage examples (e.g. `example.js`) / documentation have been updated if applicable. _(not applicable)_

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -580,9 +580,65 @@ exports.LoadUtils = () => {
 
         if (options.waitUntilMsgSent) await sendMsgResultPromise;
 
-        return window
+        let sentMsg = window
             .require('WAWebCollections')
             .Msg.get(newMsgKey._serialized);
+
+        // The sent message can be indexed under a key built with the
+        // chat's sibling wid (lid <-> phone number); the random id part
+        // of the key does not depend on the wid, so look it up directly
+        // in the chat's own collection
+        if (!sentMsg) {
+            sentMsg = chat.msgs
+                .getModelsArray()
+                .find((m) => m.id.id === newMsgKey.id);
+        }
+
+        let altWid;
+        if (
+            !sentMsg &&
+            typeof chat.id?.isUser === 'function' &&
+            chat.id.isUser()
+        ) {
+            altWid = window
+                .require('WAWebApiContact')
+                .getAlternateUserWid(chat.id);
+            if (altWid) {
+                const altMsgKey = new (window.require('WAWebMsgKey'))({
+                    from: from,
+                    to: altWid,
+                    id: newId,
+                    participant: participant,
+                    selfDir: 'out',
+                });
+                sentMsg =
+                    window
+                        .require('WAWebCollections')
+                        .Msg.get(altMsgKey._serialized) ||
+                    window
+                        .require('WAWebCollections')
+                        .Chat.get(altWid)
+                        ?.msgs.getModelsArray()
+                        .find((m) => m.id.id === newMsgKey.id);
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


fernandomondo:
The real bug is that Meta changed the message payload.

Instead of returning newMsgKey._serialized, it is now returning newMsgKey.$1.

I don't know whether this change is permanent or if Meta will roll it back in the next few hours. However, if it is permanent, we'll need to update all relevant code to accept $1 as a fallback when _serialized is not available.

edit: 
Actually, there's already a PR open with this fix: https://github.com/wwebjs/whatsapp-web.js/pull/201832

telmedola:
> The real bug is that Meta changed the message payload.
> 
> Instead of returning newMsgKey._serialized, it is now returning newMsgKey.$1.
> 
> I don't know whether this change is permanent or if Meta will roll it back in the next few hours. However, if it is permanent, we'll need to update all relevant code to accept $1 as a fallback when _serialized is not available.
> 
> edit: Actually, there's already a PR open with this fix: #201832

Thanks — you're right, `_serialized` → `$1` is the actual root cause, and #201832 is the proper fix for it. I'll defer to that PR for the rename.

That said, the fallback in this PR is complementary rather than redundant: it looks the sent message up by the random id part of the key (`m.id.id === newMsgKey.id`), which doesn't depend on `_serialized` at all — so it also covers this incident and any future change to the key serialization, plus the lid/pn indexing edge case. Happy to rebase it on top of #201832 (keeping only the lookup fallbacks and the `lastSendMessageDebug` diagnostic) if that's useful, or close it if the maintainers prefer.

Issue #201846, opened 2026-07-16
Reaction msg keys keep _serialized undefined on the $1-rename builds


rmyndharis:
### Summary

`Reaction` is the one structure the `_serialized` → `$1` rename still gets past. On the July 2026 WhatsApp Web build (2.3000.x), `reaction.id._serialized` and `reaction.msgId._serialized` are both `undefined`, so a `message_reaction` event fires but can't be attributed to a message.

This is separate from #201830 — that one covers `Message`/`Chat`/`Contact` etc., and the fix in #201832 doesn't touch `Reaction.js`.

### Why it's missed

Every other `Base` subclass assigns `this.id = data.id;` in its constructor. `Reaction` is the only one whose ids are named something else, and it assigns them inside `_patch()`:

https://github.com/wwebjs/whatsapp-web.js/blob/1780711/src/structures/Reaction.js#L21
```js
this.id = data.msgKey;          // not data.id
```
https://github.com/wwebjs/whatsapp-web.js/blob/1780711/src/structures/Reaction.js#L51
```js
this.msgId = data.parentMsgKey; // not data.id
```

So a sweep for `this.id = data.id` doesn't reach it. Worth noting #201832 *does* touch the reaction handler in `Client.js` — it fixes `senderUserJid = sender._serialized || sender.$1` — but that's a `_serialized` **read**; these two keys are passed through as objects and never read there, so a read-driven sweep had nothing to catch.

### It breaks the published types

`index.d.ts` declares both properties as `MessageId`:

https://github.com/wwebjs/whatsapp-web.js/blob/1780711/index.d.ts#L2460-L2470
```ts
export class Reaction {
    id: MessageId;
    msgId: MessageId;
```

and `MessageId._serialized` is not optional:

https://github.com/wwebjs/whatsapp-web.js/blob/1780711/index.d.ts#L1383-L1388
```ts
export interface MessageId {
    _serialized: string;
```

So on an affected build the library hands consumers an object that doesn't satisfy its own declaration, and TypeScript won't flag the read.

### To reproduce

On WA Web 2.3000.x, react to a message and log the event:

```js
client.on('message_reaction', (reaction) => {
    console.log(reaction.msgId._serialized); // undefined
    console.log(reaction.msgId.$1);          // 'true_123@c.us_ABC'
});
```

### Impact

[...]

Issue #201845, opened 2026-07-16
Client.getState() and Client.getChats() throws r: r despite client being CONNECTED and WhatsApp Web fully loaded


infoshahnn:
## Issue Description
- I am using latest version of whatsapp-web.js library and when start application it getting logs below

0|whatsapp-reader  | ⏳ 100% WhatsApp
0|whatsapp-reader  | ✅ AUTHENTICATED
0|whatsapp-reader  | ======================================
0|whatsapp-reader  | ✅ WhatsApp READY
0|whatsapp-reader  | ======================================
0|whatsapp-reader  | WhatsApp Web Version: 2.3000.1043270046
0|whatsapp-reader  | 🌐 URL: https://web.whatsapp.com/
0|whatsapp-reader  | 📄 Title: (53) WhatsApp Business
0|whatsapp-reader  | {
0|whatsapp-reader  |   body: '53\n' +
0|whatsapp-reader  |     'All\n' +
0|whatsapp-reader  |     'Unread\n' +
0|whatsapp-reader  |     '53\n' +
0|whatsapp-reader  |     'Favourites\n' +
0|whatsapp-reader  |     'Groups\n' +
0|whatsapp-reader  |     '28\n' +
0|whatsapp-reader  |     'Rice\n' +
0|whatsapp-reader  |     'यशस्वी कमोडिटी(Aivaग्रुप)\n' +
0|whatsapp-reader  |     '15:17\n' +
0|whatsapp-reader  |     "Don't Msg Yashasvi\n" +
0|whatsapp-reader  |     ': \n' +
0|whatsapp-reader  |     'Photo\n' +
0|whatsapp-reader  |     'IPGA Market Update,,,,,\n' +
0|whatsapp-reader  |     '15:22\n' +
0|whatsapp-reader  |     'Ipga market update\n' +
0|whatsapp-reader  |     ': \n' +
[...]

Issue #201844, opened 2026-07-16
message.downloadMedia() fails on every media message


neweality-a11y:
## Issue Description

`message.downloadMedia()` fails on **every** media message (images, videos, documents) since the night of **July 14→15, 2026**, coinciding with a WhatsApp Web frontend update. Confirmed to be the same root cause as #201828 (`r: r` error), but even after patching around that specific error, media resolution still fails completely for messages in **LID-based chats** — the message cannot be found at all via `Msg.get()`, `Msg.getMessagesById()`, or any in-memory model fallback.

The `message`/`message_create` **events themselves fire correctly** — only `downloadMedia()` (and by extension `resolveMediaBlob`) fails to locate/resolve the underlying message model.

## Reproduction Steps

1. Connect with `LocalAuth`
2. Receive any message with media (image, video, or PDF document) from a contact whose chat uses the new LID identifier format (e.g. `228874680049760@lid`)
3. Call `await msg.downloadMedia()`
4. Error thrown / null returned depending on version (see timeline below)

**Expected behavior:** `downloadMedia()` returns the base64 media data.

**Actual behavior:** Either throws a generic `r: r` error (pre-patch) or silently returns `null` (post-patch, see workaround below) because the message model cannot be resolved by any lookup strategy.

## Timeline / Evidence

| Time (CEST) | Event |
|---|---|
| 14 Jul, 23:15 | Last confirmed working `message`/media reception (whatsapp-web.js v1.34.7, no version pinning) |
| 15 Jul, 01:16 | Routine heartbeat restart. `authenticated` fires, but `ready` **never fires** afterward |
| 15 Jul, ~all day | Every reconnect: `authenticated` fires but `ready` hangs indefinitely; app force-restarts every 3 min (pre-ready) / 2h (post-ready) via internal heartbeat |
| 15 Jul, 22:47 onward | `ready` eventually fires (after long delays), but `downloadMedia()` starts throwing on **every single media message** with a minified error resolving to the string `"r"` |
| 16 Jul, 00:35 | WhatsApp Web pushes "stable current" version `2.3000.1043234067-alpha` (per wppconnect.io/whatsapp-versions) — `ready` event issue appears resolved |
| 16 Jul, 08:01 | Further WhatsApp Web version `2.3000.1043263898-alpha` released — `downloadMedia()` still broken |
[...]

PR #201848, opened 2026-07-16
fix: tolerate WhatsApp Web renaming the message key _serialized to $1

kristoflouviere:
## The bug

WhatsApp Web renamed the **message key's** `_serialized` property to `$1`.

`getChatModel` reads `chat.lastReceivedKey._serialized` ([Utils.js#L984-L997](https://github.com/wwebjs/whatsapp-web.js/blob/main/src/util/Injected/Utils.js#L984-L997)), which now yields `undefined` and reaches IndexedDB as `Msg.getMessagesById([undefined])`:

```
```

`getChats()` maps `getChatModel` over every chat under `Promise.all`, so **one** affected chat rejects the entire batch — surfacing to callers as an opaque minified error (widely reported as **`r: r`**) while the client is `CONNECTED` and WhatsApp Web is fully loaded.

**Chat ids are not affected** — they still carry `_serialized`. That is why this doesn't look like the `$1` rename, and why fixes that normalize ids in the Node-side `structures/` layer don't reach it: the throw happens *in-page*, inside `getChatModel`, before anything is returned.

## Evidence

Measured on two live sessions (WA Web `2.3000.1043280533`, whatsapp-web.js `1.34.7`) by catching the error **inside** `page.evaluate` so the in-page stack survives the CDP boundary (otherwise it collapses to `name: message`, i.e. `r: r`):

| | chats | with `lastReceivedKey` | of those, having `_serialized` | `getChatModel` result |
|---|---|---|---|---|
| session A | 1253 | 916 | **0** | 916 threw · 337 ok |
| session B | 515 | 498 | **0** | — |

Chats with no last message skip the block entirely — which is exactly the observed 337 / 916 split. Sampled keys look like:

```
keys: [fromMe, remote, id, $1]      _serialized: undefined      $1: "true_1252928…@lid_3A7…"
```

Re-running the same lookup with `_serialized ?? $1` returned the message on **every** affected chat.

## The fix

Adds `window.WWebJS.getMsgKeyId(key)`:

- prefers `_serialized`, so behaviour is **unchanged** wherever it is still present;
- returns `undefined` when neither exists, so callers skip the lookup rather than querying a store with a nullish key.

Applied to the three places a **raw** (un-`serialize()`d) message key is read: `getChatModel`, `sendMessage`, `editMessage`.

`getMessageModel` is deliberately untouched — it reads a `serialize()`d model, which still produces `_serialized`. `msg.id.remote._serialized` is also untouched: `remote` is a *chat* id, which the rename didn't affect.

Only `getChatModel` is verified against live sessions; `sendMessage`/`editMessage` read a raw key of the same shape and get the same one-token change. The helper is a no-op where `_serialized` still exists, so it cannot regress either path.

Refs #201845, #201844, #201846

---
Prettier + the existing style are unchanged (`prettier --check` passes; diff is 1 file, +24/−7).

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -836,6 +836,20 @@ exports.LoadUtils = () => {
             });
         }
 
+        // The rename reaches the serialized model too: `msg.id` comes back carrying `$1` and no
+        // `_serialized`, so every consumer of `message.id._serialized` — including this library's own
+        // Message structure and anything keyed on a message id — silently receives undefined. Restore
+        // it here, where `msg.id.remote` is already normalised, so the whole downstream surface keeps
+        // working rather than each caller having to know about `$1`.
+        if (typeof msg.id === 'object' && msg.id._serialized == null) {
+            const serializedId = window.WWebJS.getMsgKeyId(msg.id);
+            if (serializedId) {
+                msg.id = Object.assign({}, msg.id, {
+                    _serialized: serializedId,
+                });
+            }
+        }
+
         delete msg.pendingAckUpdate;
 
         return msg;
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -582,7 +582,7 @@ exports.LoadUtils = () => {
 
         return window
             .require('WAWebCollections')
-            .Msg.get(newMsgKey._serialized);
+            .Msg.get(window.WWebJS.getMsgKeyId(newMsgKey));
     };
 
     window.WWebJS.editMessage = async (msg, content, options = {}) => {
@@ -625,7 +625,9 @@ exports.LoadUtils = () => {
         await window
             .require('WAWebSendMessageEditAction')
             .sendMessageEdit(msg, content, internalOptions);
-        return window.require('WAWebCollections').Msg.get(msg.id._serialized);
+        return window
+            .require('WAWebCollections')
+            .Msg.get(window.WWebJS.getMsgKeyId(msg.id));
     };
 
     window.WWebJS.toStickerData = async (mediaInfo) => {
@@ -917,6 +919,20 @@ exports.LoadUtils = () => {
         };
     };
 
+    /**
+     * The serialized id of a message key, tolerating WhatsApp Web having renamed the key's
+     * `_serialized` property to `$1`. Reading the old name now returns `undefined`, which reaches
+     * IndexedDB as a `.get(undefined)` and throws
+     * `DataError: Failed to execute 'get' on 'IDBObjectStore': No key or key range specified.`
+     * `_serialized` is preferred so nothing changes wherever it is still present (chat ids, for
+     * example, are unaffected by the rename). Returns undefined when neither exists, so callers can
+     * skip the lookup instead of querying a store with a nullish key.
+     * @param {Object} key a message key, e.g. `chat.lastReceivedKey` or a raw `msg.id`
+     * @returns {string|undefined}
+     */
+    window.WWebJS.getMsgKeyId = (key) =>
+        key?._serialized ?? key?.$1 ?? undefined;
+
     window.WWebJS.getChats = async () => {
         const chats = window.require('WAWebCollections').Chat.getModelsArray();
         const chatPromises = chats.map((chat) =>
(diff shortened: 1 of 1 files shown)


jonnytest1 (approved):
lgtm

dolav:
Tested this and it perfectly fixes the r: r IndexedDB crashes on my headless Oracle Cloud setup. +1 for this surgical approach!

vaibhavpandeyvpz:
It worked perfectly for me. Thanks! But this ultimately requires patch in https://github.com/wwebjs/whatsapp-web.js/pull/201847 as well

vaibhavpandeyvpz:
Few misses still:

## #1 — Poll votes

This is a **real uncovered path**, assuming `pollUpdateParentKey` underwent the same `_serialized` → `$1` rename.

When a live poll-vote update arrives, `Client.js` does:

```js
const parentMsgKey = vote.pollUpdateParentKey;

let parentMessage = Msg.get(parentMsgKey._serialized);
if (!parentMessage) {
    const fetched = await Msg.getMessagesById([
        parentMsgKey._serialized,
    ]);
}
```

If that key now looks like:

```js
{ $1: '...', remote: ..., id: ... }
```

both lookups receive `undefined`. Unlike issue #1, the first lookup occurs directly in the event-processing path, so it may throw and prevent the poll-vote event from being emitted.

The resulting `PollVote` also exposes the key unchanged:

```js
this.parentMsgKey = data.parentMsgKey;
```

So consumers reading `vote.parentMsgKey._serialized` would also get `undefined`.

### Scope

This mainly affects:

- live `vote_update`/poll-vote events;
- resolving the poll creation message associated with a vote;
- consumers accessing the vote’s parent message ID.

It does **not necessarily break** `message.getPollVotes()`. That method begins with the poll message’s primary `msg.id`, which your `getMessageModel` normalization already repairs. The returned vote objects may still expose an unnormalized `parentMsgKey`, though.

### Suggested minimal fix

Use `getMsgKeyId` in the injected poll-vote handler and normalize the key before it crosses into `PollVote`:

```js
const parentMsgKeyId = window.WWebJS.getMsgKeyId(parentMsgKey);

let parentMessage = parentMsgKeyId
    ? Msg.get(parentMsgKeyId)
    : null;

if (!parentMessage && parentMsgKeyId) {
    const fetched = await Msg.getMessagesById([parentMsgKeyId]);
    parentMessage = fetched?.messages?.[0] || null;
}
```

Then either serialize `parentMsgKey` there or normalize it in `PollVote`, as done for `Reaction`.

---

## #2 — Nested message IDs

There are two nested fields:

```js
message.latestEditMsgKey
message.protocolMessageKey
```

Both are declared as `MessageId`, meaning callers reasonably expect:

```js
key._serialized
```

Your normalization in `getMessageModel` repairs only:

```js
msg.id
```

It does not recursively repair those nested keys.

### `latestEditMsgKey`

This appears to be **mostly an exposed-data/API consistency issue**. Within `whatsapp-web.js@1.34.7`, it is assigned but not otherwise consumed:

```js
this.latestEditMsgKey = data.latestEditMsgKey;
```

Therefore, the library itself probably continues working. Only consumers that access:

```js
message.latestEditMsgKey._serialized
```

would observe the break. If your project does not use that property, you can likely ignore it.

### `protocolMessageKey`

This one has a narrow internal consequence during “delete for everyone” events:

```js
if (message.protocolMessageKey)
    revoked_msg.id = { ...message.protocolMessageKey };
```

If `protocolMessageKey` contains only `$1`, the emitted `revoked_msg.id` also lacks `_serialized`. Consequently:

```js
revokedMessage.id._serialized
```

would be undefined for listeners of `message_revoke_everyone`.

It likely does **not prevent the revocation event itself** from firing. It mainly leaves the reconstructed revoked message with an incomplete public ID.

## Practical priority

| Path | Likely impact | Recommendation |
|---|---|---|
| Poll-vote parent key | Event processing may fail | **Worth fixing** if polls matter |
| `protocolMessageKey` | Revoked message ID missing | Fix if handling delete-for-everyone events |
| `latestEditMsgKey` | Public nested ID missing | Low priority unless consumed |

vaibhavpandeyvpz:
This is the patch that's working for me https://github.com/vaibhavpandeyvpz/wappmcp/tree/main/patches until changes in this PR are released

jonnytest1:
Btw , has there been a discussion about bundling the current whatsappweb into the package , or agressively caching them post deploy so upstream changes arent as likely to break things ? (I imagine whatsapp has to keep up the older apis for a while anyways ) 🤔

Adi1231234:
Confirmed @vaibhavpandeyvpz's poll-vote point on a live 2.3000.x session: `vote.pollUpdateParentKey` is a `MsgKey` with `_serialized` now `undefined` (the value moved to `$1`), so `Msg.getMessagesById([undefined])` throws the same `r`. Running it through `getMsgKeyId` resolves it.

Happy to help however's easiest for you - I can open a full PR with all the fixes bundled together, or if you'd rather keep everything in this one, I'm glad to send the changes your way. Whatever works best. 🙂

XxQuickSilverZz:
> Confirmed @vaibhavpandeyvpz's poll-vote point on a live 2.3000.x session: `vote.pollUpdateParentKey` is a `MsgKey` with `_serialized` now `undefined` (the value moved to `$1`), so `Msg.getMessagesById([undefined])` throws the same `r`. Running it through `getMsgKeyId` resolves it.
> 
> Happy to help however's easiest for you - I can open a full PR with all the fixes bundled together, or if you'd rather keep everything in this one, I'm glad to send the changes your way. Whatever works best. 🙂

please tell me all changes...

PR #201881, opened 2026-08-03
feat(call): add experimental WhatsApp Web call controls

lokoamigo:
## Description

Adds an experimental call-control API backed by WhatsApp Web's internal VoIP
controllers.

This PR adds support for:

- starting outgoing 1:1 calls with `client.startCall(contactId, options)`
- starting outgoing group calls with `client.startGroupCall(contactIds, options)`
- reading the active call with `client.getActiveCall()`
- accepting the active incoming call with `client.acceptCall(callId?)` /
  `call.accept()`
- rejecting the active incoming call with `client.rejectCall(callId?)` /
  `call.reject()`
- ending the active call with `client.endCall(callId?)` / `call.end()`
- inviting a participant with `client.addParticipantToCall(contactId, callId?)`
  / `call.addParticipant(contactId)`
- removing a participant with
  `client.removeParticipantFromCall(contactId, callId?)` /
  `call.removeParticipant(contactId)`

The implementation intentionally keeps private WhatsApp Web calls in the
injected browser layer and fails clearly when the required internal controllers
are unavailable. It does not send custom call signaling stanzas.

Typings, focused unit tests, and experimental documentation are included.

## Related Issue(s)

N/A

### Test Details

Tested with focused unit coverage for the new call APIs:

- `npx mocha tests/startCall.js --timeout 5000`
- `npx eslint src/Client.js src/structures/Call.js src/util/Injected/Utils.js tests/startCall.js`
- `git diff --check`

The full test suite was also attempted with `WWEBJS_TEST_CLIENT_ID`,
`WWEBJS_TEST_REMOTE_ID`, and `PUPPETEER_EXECUTABLE_PATH` configured:

- `PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium WWEBJS_TEST_CLIENT_ID=test WWEBJS_TEST_REMOTE_ID=<test-contact> npm test`

That run reached `53 passing / 6 failing`. The focused call tests passed, but
the full suite did not complete in this local environment because the shared
authenticated `LocalAuth` session was locked by an earlier browser instance
(`The browser is already running for .../.wwebjs_auth/session-test`). For that
reason, the full-suite checklist item is left unchecked below.

Manual runtime validation was performed against a live WhatsApp Web session:

- started an outgoing voice call
- ended an outgoing call
- invited an additional participant into an active call
- started an outgoing group call
- rejected an incoming call
- accepted an incoming voice call detected through `getActiveCall()`

`acceptCall()` was live-tested after `getActiveCall()` detected an incoming
active call; the existing `call` event did not fire in that run before polling
detected the call. `removeParticipantFromCall()` was verified by runtime module
inspection and unit tests, but was not live-completed end-to-end because it
requires an accepted multi-party call and additional consenting test
participants.

A proof-of-concept consumer project was used during development to validate the
API shape in a real integration scenario:

- https://github.com/lokoamigo/whatsapp-teamspeak-bridge

This was used as supplemental validation only; the library changes are covered
by the tests listed above.

### Environment

- Machine OS: Manjaro Linux
- Phone OS: 
- Library Version: 1.34.7
- WhatsApp Web Version: 
- Browser Type and Version: Chromium 150.0.7871.186
- Node Version: v26.4.0

## Type of Change

- [x] **Non-code change** _(documentation, README, etc.)_

## Checklist

- [ ] All new and existing tests pass (`npm test`).
- [x] Typings (e.g. `index.d.ts`) have been updated if necessary.
- [x] Usage examples (e.g. `example.js`) / documentation have been updated if applicable.

### DIFF
--- a/example.js
+++ b/example.js
@@ -697,7 +697,24 @@ client.on('call', async (call) => {
         call.from,
         `[${call.fromMe ? 'Outgoing' : 'Incoming'}] Phone call from ${call.from}, type ${call.isGroup ? 'group' : ''} ${call.isVideo ? 'video' : 'audio'} call. ${rejectCalls ? 'This call was automatically rejected by the script.' : ''}`,
     );
+
+    // Instead of rejecting, a call can also be answered and an audio clip
+    // (e.g. a text-to-speech file you generated) can be played into it:
+    // await call.accept(); // answers with audio only, even for video calls
+    // await call.playAudio(MessageMedia.fromFilePath('./welcome.mp3'));
+    // await call.end();
 });
+
+// Placing an outgoing call and speaking once the other party answers:
+// const call = await client.call('1234567890', { waitForAnswer: true });
+// if (await call.isConnected()) {
+//     await call.playAudio(MessageMedia.fromFilePath('./message.mp3'));
+//     await call.end();
+// }
+
+// Checking for an ongoing call and hanging it up:
+// const activeCall = await client.getActiveCall();
+// if (activeCall) await activeCall.end();
 
 client.on('disconnected', (reason) => {
     console.log('Client was logged out', reason);
--- a/index.d.ts
+++ b/index.d.ts
@@ -369,6 +369,13 @@ declare namespace WAWebJS {
         ): Promise<Call>;
 
         /**
+         * Gets the number of participants in the currently active WhatsApp
+         * call, or null when no call or countable roster is available.
+         * @experimental Depends on private WhatsApp Web internals.
+         */
+        getActiveCallParticipantCount(): Promise<number | null>;
+
+        /**
          * Adds an individual WhatsApp contact to the currently active call.
          * @experimental Depends on private WhatsApp Web internals and may reject
          * when no supported internal call controller is detected.
--- a/index.d.ts
+++ b/index.d.ts
@@ -357,11 +357,33 @@ declare namespace WAWebJS {
         getActiveCall(): Promise<Call | null>;
 
         /**
+         * Starts an outgoing WhatsApp group call.
+         * @experimental Depends on private WhatsApp Web internals and may reject
+         * when no supported internal group call controller is detected.
+         */
+        startGroupCall(
+            contactIds: string[],
+            options?: {
+                video?: boolean;
+            },
+        ): Promise<Call>;
+
+        /**
          * Adds an individual WhatsApp contact to the currently active call.
          * @experimental Depends on private WhatsApp Web internals and may reject
          * when no supported internal call controller is detected.
          */
         addParticipantToCall(contactId: string, callId?: string): Promise<void>;
+
+        /**
+         * Removes an individual WhatsApp contact from the currently active call.
+         * @experimental Depends on private WhatsApp Web internals and may reject
+         * when no supported internal call controller is detected.
+         */
+        removeParticipantFromCall(
+            contactId: string,
+            callId?: string,
+        ): Promise<void>;
 
         /**
          * Sends a response to the scheduled event message, indicating whether a user is going to attend the event or not
@@ -2480,6 +2502,9 @@ declare namespace WAWebJS {
 
         /** Add an individual WhatsApp contact to the active call */
         addParticipant: (contactId: string) => Promise<void>;
+
+        /** Remove an individual WhatsApp contact from the active call */
+        removeParticipant: (contactId: string) => Promise<void>;
     }
 
     /** Message type List */
--- a/index.d.ts
+++ b/index.d.ts
@@ -357,6 +357,13 @@ declare namespace WAWebJS {
         getActiveCall(): Promise<Call | null>;
 
         /**
+         * Adds an individual WhatsApp contact to the currently active call.
+         * @experimental Depends on private WhatsApp Web internals and may reject
+         * when no supported internal call controller is detected.
+         */
+        addParticipantToCall(contactId: string, callId?: string): Promise<void>;
+
+        /**
          * Sends a response to the scheduled event message, indicating whether a user is going to attend the event or not
          * @param response The response code to the event message. Valid values are: `0` for NONE response (removes a previous response) | `1` for GOING | `2` for NOT GOING | `3` for MAYBE going
          * @param eventMessageId The event message ID
@@ -2470,6 +2477,9 @@ declare namespace WAWebJS {
 
         /** Indicates whether the call is currently connected (the other party has answered) */
         isConnected: () => Promise<boolean>;
+
+        /** Add an individual WhatsApp contact to the active call */
+        addParticipant: (contactId: string) => Promise<void>;
     }
 
     /** Message type List */
--- a/index.d.ts
+++ b/index.d.ts
@@ -212,10 +212,7 @@ declare namespace WAWebJS {
         ): Promise<Message>;
 
         /** Send a reaction to a specific messageId */
-        sendReaction(
-            messageId: string,
-            reaction: string,
-        ): Promise<void>;
+        sendReaction(messageId: string, reaction: string): Promise<void>;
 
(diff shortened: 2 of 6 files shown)


lokoamigo:
I noticed that PR #201825 already implements part of the same call-control functionality covered here, particularly starting 1:1 calls and rejecting or ending active calls.

I wasn’t aware of that PR when I started this work, so the overlap was unintentional. This PR still adds some functionality that does not appear to be covered there, mainly group-call support and adding or removing participants from an active call.

I’m happy to rework this PR to avoid maintaining two competing implementations—for example, by rebasing the group-call and participant-management changes on top of #201825 and removing the duplicated 1:1 call lifecycle code.

@aliyss, @lindionez , please let me know which approach and public API you would prefer.

lokoamigo:
Implemented the missing incoming-call control pieces:

  - added client.getActiveCall() returning Call | null
  - added client.acceptCall(callId?)
  - added call.accept()
  - updated typings, docs, and focused unit tests
  - live-tested accepting an incoming voice call via getActiveCall() + acceptCall(call.id)

One relevant finding: in the live test, the existing call event did not fire before polling detected the active incoming call, so the docs now
  mention using getActiveCall() polling for stronger incoming-call detection.

lokoamigo:
Status update:

  I found and fixed one more runtime issue while validating the call APIs in my POC.

  The existing incoming `call` event hook was not reliable with the current WhatsApp Web `WAWebCallCollection` shape. Manual `getActiveCall()` / `acceptCall()` worked, but consumers relying on `client.on('call', ...)`
  did not always receive the incoming call event.

  New fixes pushed:

  - `6099ed0` observes incoming calls from the actual call collection and active-call changes
  - `67998b5` deduplicates incoming call events by call id
  - `f7a67c9` prevents duplicate call listener registration when event listeners are reattached

  Runtime validation was performed through the POC integration. After rebuilding the consumer against `f7a67c9`, an incoming WhatsApp voice call produced the expected flow:

  - `call` event observed once
  - auto-accept scheduled
  - `acceptCall()` completed successfully

  This also confirmed that the framework-level event path and the explicit call-control methods now work together for incoming calls.

alpersamur3:
@lokoamigo thanks for flagging this and for the very generous offer to rework, much appreciated.

For context, #201825 is intentionally focused: it adds call answering, outgoing 1:1 calls, reject/end, and its main differentiator is audio injection (`playAudio`), which feeds a synthetic media stream into a live call so a bot can play a TTS file or a recording to the other party. It is already in review with feedback addressed and now has unit tests, and it does not touch group calls or participant management, so your group-call and add/remove-participant work is genuinely additive rather than a duplicate.

Interesting that we both independently hit and fixed the same incoming-call issue (the old hook not firing on the current `WAWebCallCollection` shape); good sign that fix is needed.

To avoid two implementations of the 1:1 basics, one option would be to land #201825 first since it is already under review, and build the group-call and participant APIs on top of it (either you rebase the additive parts onto it, or as a follow-up PR), so accept/reject/end/getActiveCall live in one place. Happy to coordinate on the shared surface (naming, the `getActiveCall` shape) so they line up. Whatever sequencing the maintainers prefer works for me though.

Thanks again, and nice work on the group-call side.

lokoamigo:
Small update: @gitlost808 added an API for retrieving the participant count of the currently active call.

The new `client.getActiveCallParticipantCount()` checks the active `WAWebCallCollection` call and reads the serialized call participant roster, returning the number of participants when available and `null` when there is no active/countable call.

The addition also includes the corresponding TypeScript definition and focused tests covering both the no-active-call case and participant counting.

Thanks @gitlost808 for the contribution!

lokoamigo:
@alpersamur3 thanks, that approach makes sense to me.

I'm happy to rebase my branch on top of #201825 / your `feat/call-answering-audio` branch and remove the overlapping 1:1 call lifecycle implementation from my side. That should leave one implementation for the shared `getActiveCall` / accept / reject / end functionality, while keeping the group-call and participant-management work here as the additive part.

I also checked @gitlost808's recent participant-count addition before doing that. As far as I can see, `getActiveCallParticipantCount()` is not duplicated by #201825 at the moment. Your branch does already carry participant data on the `Call` model/incoming-call path, but `getActiveCall()` itself does not currently expose that participant roster and there is no equivalent participant-count API.

So my preference would be to retain that addition when rebasing. If it turns out to become redundant once the two implementations are combined, though, I'm also completely fine with removing it rather than ending up with two competing ways/states for representing the same information.

I'll align the remaining API surface with #201825 during the rebase so the group-call and participant features build on top of the same call implementation. It will take me some time, though, as I'm currently busy with other projects.

@aliyss, @lindionez, please review the current state first so I can incorporate all feedback and implement the remaining changes in one go.

alpersamur3:
@lokoamigo sounds great, thanks for aligning on this.

Agreed on keeping `getActiveCall` / accept / reject / end in #201825 as the shared base, with your group-call and participant-management work layered on top. And `getActiveCallParticipantCount()` is a clean additive API that #201825 does not cover, so it makes sense to keep it on your side. No rush on the rebase given you are busy.

@aliyss @lindionez since #201825 is turning into the shared base here (my video PR #201889 already stacks on it, and lokoamigo plans to rebase the group-call and participant work on top of it too), it would help everyone if #201825 could be reviewed first: it gives lokoamigo a stable foundation to rebase onto and avoids three PRs carrying overlapping 1:1 call code. I also have #201888 (small and independent, based on `main`) which surfaces how a call ended on `call_log` messages, e.g. `AcceptedElsewhere` when answered on another device, which complements the incoming-call detection discussed here. Happy to address feedback on any of them quickly.

lokoamigo:
Thanks @alpersamur3, 
I rebased the branch onto your call-answering/audio PR as discussed.

I kept your 1:1 call lifecycle surface as the base and dropped the overlapping older implementation from this branch. This PR now only carries the additive pieces on top: 

1. group call start
2. add/remove participants
3. participant count
4. related tests/typings

I also adjusted the downstream teamspeak-whatsapp-bridge against the rebased API and tested it locally. It "works on my machine" ;-)

alpersamur3:
@lokoamigo awesome, thanks for the quick rebase! nice that it came together so cleanly. the group-call and participant-management side is a solid addition on top of the shared base, so together they cover the call feature pretty well. hope the maintainers can take a look soon.

lokoamigo:
Hi @aliyss @lindionez,

it has been a while since the last update, and the requested alignment/rebase onto #201825 was completed back in August. Is there anything currently blocking a review or anything else you’d like me to change before this can move forward?

I’d appreciate a quick status update, even if it’s just to know whether this is still something you’re interested in merging.

If not let me at least know.

Thanks in advance.

alpersamur3:
@lokoamigo heads-up: WhatsApp Web now shows a "Start a WhatsApp call with X?" confirmation before any 1:1 call that isn't flagged as a user gesture, so `startWAWebVoipCall` hangs on that popup and no call goes out (reported on #201825). I pushed the fix to #201825 as 5a78ab0: it passes `{ entryTrust: 'user_gesture' }` the same way the call buttons do. Your branch is still based on a522119, so it doesn't have the fix yet. I checked that your three commits cherry-pick cleanly on top of the updated branch. Group calls aren't affected right now since only the 1:1 path checks consent, but WA's own group-call buttons pass the same `{ entryTrust: 'user_gesture' }` as the last argument to `startWAWebVoipGroupCallFromWids`, so it may be worth passing it there too in case the check gets extended.

lokoamigo:
Oh is that so? I guess the feature is not rolled out on my end yet. I'll try to reproduce it and fix this issue.

lokoamigo:
@alpersamur3  I rebased onto your latest branch and added the same user_gesture flag to group calls. I tested both 1:1 and group calls live. Both start without the popup and work as expected. All 36 call tests, ESLint, and Prettier pass. The updated branch is pushed.

PR #201888, opened 2026-08-06
feat: expose call outcome on call_log messages

alpersamur3:
## What

Exposes the outcome of `call_log` messages on the `Message` structure: `message.callOutcome`, `message.callDuration` and `message.isVideoCall`.

## Why

There was no way to tell how a call ended. In particular, when an incoming call is answered on one of the account's other linked devices, `callOutcome` is `'AcceptedElsewhere'`, which lets a bot know it should not auto-answer. Other common values include `'Completed'`, `'Missed'`, `'Rejected'`, `'Cancelled'` and `'Failed'`.

Requested in the call PR discussion (#201825).

## How

`call_log` messages already carry the outcome in their serialized model; this just surfaces it (along with the duration and the video flag) on the `Message` structure. `call_log` messages are delivered through the existing `message` / `message_create` events, so no new event is needed.

## Usage

```js
    if (msg.type === 'call_log' && msg.callOutcome === 'AcceptedElsewhere') {
        // The call was answered on another linked device
    }
});
```

## Tested

- Verified against a live session: an incoming call answered on another device produced a `call_log` message with `callOutcome: 'AcceptedElsewhere'`; a normal completed call read `'Completed'`.
- Unit tests added in `tests/callLog.js`.

### DIFF
--- a/example.js
+++ b/example.js
@@ -67,6 +67,14 @@ client.on('ready', async () => {
 client.on('message', async (msg) => {
     console.log('MESSAGE RECEIVED', msg);
 
+    if (msg.type === 'call_log') {
+        // Call log messages carry the outcome of a call, e.g. 'AcceptedElsewhere'
+        // when the call was answered on one of your other linked devices
+        console.log(
+            `Call from ${msg.from} ended with outcome: ${msg.callOutcome}`,
+        );
+    }
+
     if (msg.body === '!ping reply') {
         // Send a new message as a reply to the current one
         msg.reply('pong');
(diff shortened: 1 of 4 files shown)


alpersamur3:
Re-checked against the current WA Web (2.3000.1048354754): `call_log` messages still carry `callOutcome`, `callDuration` and `isVideoCall`, so this is still good to go.

@BenyFilho @tuyuribr @aliyss this one is small and independent of the other call PRs, if you have a moment to review it.

PR #201893, opened 2026-08-11
fix(client): prevent concurrent injection crash

MySSoN:
## Problem

Concurrent `Client.inject()` calls can reach `exposeFunctionIfAbsent()` at the
same time during page navigation or authentication.

Because checking `window[name]` and calling `page.exposeFunction()` are separate
operations, both calls can pass the check. The second registration then throws:

```text
Failed to add page binding with name onQRChangedEvent:
window['onQRChangedEvent'] already exists!
```

The error can escape from the async `framenavigated` listener as an unhandled
rejection and terminate the Node.js process.

## Changes

- Serialize `inject()` calls per `Client` instance.
- Preserve parallel initialization between different clients.
- Avoid a queue deadlock during authentication recovery.
- Treat only the known duplicate Puppeteer binding error as idempotent.
- Catch failures from the async `framenavigated` flow and emit `inject_error`.
- Continue propagating unrelated Puppeteer errors.

## Testing

- `node --check src/Client.js`
- `node --check src/util/Puppeteer.js`
- `git diff --check`
- Verified that the duplicate binding error is ignored.
- Verified that unrelated Puppeteer errors still propagate.

A live WhatsApp session is still required for full QR/logout/reconnect testing.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -103,6 +103,9 @@ class Client extends EventEmitter {
 
         this.currentIndexHtml = null;
         this.lastLoggedOut = false;
+        // Keep injection work isolated per Client. Different Client instances
+        // still inject in parallel, while calls for this page are serialized.
+        this._injectQueue = Promise.resolve();
 
         Util.setFfmpegPath(this.options.ffmpegPath);
     }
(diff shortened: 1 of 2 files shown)


MySSoN (commented):
Review this bug fix please

Issue #201862, opened 2026-07-27
Errors r: r on client.getChats() and t: t on client.sendMessage(media)


GarfieldMSB:
## Issue Description
Hello, I am using the latest commit from the master branch (github:pedroslopez/whatsapp-web.js) and I am facing minified execution errors when trying to fetch chats and send media messages.

**Error Logs**:
Crash on getChats():
`
Error: r: r
    at #evaluate (puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:391:56)
    at async ExecutionContext.evaluate (puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:277:16)
    at async Client.getChats (whatsapp-web.js/src/Client.js:1729:23)
`

Crash on sendMessage() with media:
`
Error: t: t
    at #evaluate (puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:391:56)
    at async ExecutionContext.evaluate (puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:277:16)
    at async Client.sendMessage (whatsapp-web.js/src/Client.js:1593:25)

`

## Reproduction Steps
You can reproduce this issue by running the following minimal script. It simply authenticates and attempts to call getChats() and send a base64 image to the bot's own number.

## Code Sample
```typescript
const { Client, LocalAuth, MessageMedia } = require('whatsapp-web.js');
const qrcode = require('qrcode-terminal');

const client = new Client({
    authStrategy: new LocalAuth()
});

client.on('qr', (qr) => {
    qrcode.generate(qr, { small: true });
});

client.on('ready', async () => {
    console.log('Client is ready!');
    
    // Reproducing `getChats()` crash (r: r)
    try {
        console.log('Testing getChats()...');
        const chats = await client.getChats();
        console.log(`Success: Found ${chats.length} chats`);
    } catch (e) {
        console.error('Crash on getChats():', e);
    }
    
    // Reproducing `sendMessage(media)` crash (t: t)
    try {
        console.log('Testing sendMessage with media...');
        const targetId = client.info.wid._serialized; // Send to itself
        const media = new MessageMedia('image/png', 'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==');
        await client.sendMessage(targetId, media);
[...]

PR #201840, opened 2026-07-15
fix: dual-compat for WA Web 2.3000.1043xxx ($1 vs _serialized, downloadMedia)

erickmourasilva:
## Summary
- Add `widSerialized()` / `normalizeSerialized()` so WID/MsgKey work with both `_serialized` (legacy) and `$1` (WA Web 2.3000.1043xxx+).
- Wire helpers into `getMessageModel` and `getChatModel` (fixes `getChatById` / typing crashes with `r: r`).
- Dual-read LID/PN in `getContactLidAndPhone` and Message `_patch` from/to/author.
- Use dual message id for `downloadMedia` / `downloadMediaStream`.
- In `resolveMediaBlob`, fall back to `DownloadManager.downloadAndMaybeDecrypt` when the in-memory blob is missing or looks thumbnail-sized vs `msg.size`.

## Compatibility
- Old WA Web: unchanged path via `_serialized`.
- New WA Web: falls back to `$1` and mirrors onto `_serialized` for node-side consumers.

## Test plan
- [ ] Text receive + send on WA Web **before** 2.3000.1043xxx
- [ ] Text receive + send on WA Web **2.3000.1043xxx+**
- [ ] `msg.getChat()` / typing indicator does not throw
- [ ] Incoming image: `downloadMedia()` returns full media (bytes ≈ `msg.size`, not thumbnail)
- [ ] `getContactLidAndPhone` still returns lid/pn strings

Closes #201831

Made with [Cursor](https://cursor.com)

Issue #201836, opened 2026-07-15
When sending message, IDMessage returns empty since yesterday..


yairmelamed:
## Issue Description
When sending message, IDMessage returns empty since yesterday..

## Reproduction Steps
Just run client.sendMessage(

## Code Sample
```typescript
return await this.client.sendMessage(number_details._serialized, mediaFile, options);
```

## User Setup

| WhatsApp | Type |
|---|---|
| **Account Type** | WhatsApp Business |
| **Authentication Strategy** | LocalAuth |
| **WhatsApp Web Version** | 2.3000.1031490220-alpha |
| **whatsapp-web.js Version** | 1.34.7 |

| Environment | Version |
|---|---|
| **Browser Type** | Chromium |
| **Browser Version** | 120 |
| **Phone OS Version** | Android or Iphone |
| **Running OS Version** | Ubuntu 24 |
| **Node.js Version** | 18 |

## Checklist
- [x] **I use the latest released version of whatsapp-web.js.**
- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have verified this is a library issue, not a problem with my code.

fatibhussain:
i am getting same issue since yesterday response : TypeError: Cannot read properties of undefined (reading 'id')
while check sending message obj is undefined as well

Shvan11:
same error here!

PR #201901, opened 2026-09-01
fix(client): sendMessage returns undefined on newer WA Web builds

netobasilio:
## Description

On WA Web `2.3000.1043xxx+` a `MsgKey` no longer exposes `_serialized`. The value is
still there — reachable through the minified `$1` property and through `toString()` —
but the named property is gone.

`WWebJS.sendMessage` ends by looking the freshly created message up by that property:

```js
return window.require('WAWebCollections').Msg.get(newMsgKey._serialized);
```

On these builds this is `Msg.get(undefined)`, which returns nothing, so
`Client.sendMessage` resolves to `undefined` — **even though the message was actually
delivered**. Callers that check the return value report a successful send as a failure,
and no message id is available for tracking acks.

This PR falls back to `$1` and then to `toString()`, keeping `_serialized` first so
older builds take exactly the same path as before.

### Relationship to #201840

#201840 covers the **read** path (`getMessageModel`, `getChatModel`, `Message._patch`,
`downloadMedia`) and does not touch this line. The two changes are complementary and do
not overlap:

- With #201840 alone, `Client.sendMessage` still resolves to `undefined`, because the
  lookup fails before `getMessageModel` is ever reached.
- With this PR alone, `sendMessage` returns the `Message` again, but its node-side
  `id._serialized` is still `undefined` — that part is what `normalizeSerialized` in
  #201840 fixes.

I kept this patch minimal and self-contained on `main` rather than building on #201840,
so it can land independently. Happy to rebase it onto that PR's helpers
(`widSerialized`) instead if the maintainers prefer a single entry point.

## Related Issue(s)

Related to #201836 (`sendMessage` returning an empty message id) and #201862.

I have deliberately not used a closing keyword: those reports do not state the WA Web
build precisely enough for me to confirm they are the same root cause, and I would
rather not auto-close someone else's issue on an assumption.

### Test Details

Verified against a live WhatsApp Web session, sending to a self-chat, with the library
copied to a scratch directory so the same session could be reused between runs. No
other patches were applied, so the result isolates this change.

**Before (unpatched `main`):**

```
sendMessage returned UNDEFINED
```

**After (this patch, nothing else changed):**

```
sendMessage returned Message (id._serialized=undefined)
```

The message was delivered in both runs — confirming the send itself was never broken,
only the lookup of its own result. The remaining `id._serialized=undefined` on the node
side is the part #201840 addresses.

Supporting observations from the same session, for whoever picks this up:

- A freshly built `MsgKey` has own properties `fromMe`, `remote`, `id`, `$1` — no
  `_serialized`; its prototype only carries `constructor`, `toString`, `clone`, `equals`.
- `$1` and `toString()` return the same string
  (`true_<remote>_<id>` / `true_<remote>_<id>_<participant>` for groups), which is why
  either works as a fallback.

`npm run check` passes (ESLint clean; Prettier reports only a pre-existing warning on
`index.d.ts`, which is present on unmodified `main` and is untouched here).

`npm test` could not be run: the suite requires `WWEBJS_TEST_REMOTE_ID` and a dedicated
paired account, and it fails to load without them on unmodified `main` too. I did not
point it at a production session. The manual verification above is what I have.

### Environment

- Machine OS: Ubuntu 24.04.4 LTS
- Phone OS: not recorded
- Library Version: 1.34.7 (change authored against `main` @ 942d236, where the affected
  line is byte-identical)
- WhatsApp Web Version: 2.3000.1046520132
- Browser Type and Version: Chrome/146.0.7680.31 (puppeteer bundled Chromium)
- Node Version: v20.20.2

## Type of Change

- [ ] **New feature** _(non-breaking change that adds functionality)_

## Checklist

- [ ] All new and existing tests pass (`npm test`). — see Test Details: the suite needs a
      dedicated paired test account; `npm run check` passes.
- [x] Typings (e.g. `index.d.ts`) have been updated if necessary. — no typing change; the
      public signature is unchanged, this only stops it returning `undefined`.
- [x] Usage examples (e.g. `example.js`) / documentation have been updated if applicable. —
      not applicable, internal lookup only.

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -580,9 +580,19 @@ exports.LoadUtils = () => {
 
         if (options.waitUntilMsgSent) await sendMsgResultPromise;
 
-        return window
-            .require('WAWebCollections')
-            .Msg.get(newMsgKey._serialized);
+        // Newer WA Web builds (2.3000.1043xxx+) no longer expose `_serialized` on a
+        // MsgKey; the same value is only reachable through `$1` or `toString()`.
+        // Without a fallback the just-created message is looked up by `undefined`,
+        // so this resolves to `undefined` and `Client.sendMessage` returns
+        // `undefined` even though the message was sent.
+        const sentMsgId =
+            newMsgKey._serialized ??
+            newMsgKey.$1 ??
+            (typeof newMsgKey.toString === 'function'
+                ? newMsgKey.toString()
+                : undefined);
+
+        return window.require('WAWebCollections').Msg.get(sentMsgId);
     };
 
     window.WWebJS.editMessage = async (msg, content, options = {}) => {


emrecirik:
**Confirmed on WhatsApp Web 2.3000.1047868043** (whatsapp-web.js 1.34.7, puppeteer 24.38.0, Chrome, Node 22).

Dumping the MsgKey of a just-sent message from the page shows the own properties are `fromMe, remote, id, self, $1`, with `$1 = "true_<redacted>@lid_3EB0…_out"` and `_serialized` undefined — so `Msg.get(undefined)` resolves to `undefined` while the message is in fact delivered (it shows up in the chat with ack 3).

One suggestion: instead of fixing this single call site, defining a `_serialized` getter on `window.require('WAWebMsgKey').prototype` restores every page-side read at once — this same lookup, `editMessage` (`Msg.get(msg.id._serialized)`) and `chat.lastReceivedKey._serialized` all break the same way on this build. I'm running that shape in production and it fixes all three.

netobasilio:
Switched to the prototype getter in 1a94695, with a setter so builds that still assign `_serialized` in the constructor keep working. The node-side model still needs #201840.

Issue #201884, opened 2026-08-05
chat.fetchMessages() give error t: t


reinforcers:
## Issue Description
chat.fetchMessages() give error t: t
I try replace `_serialized` with `$1` in some files but still nothing works
```
t: t
    at #evaluate (/home/xxxxxx/node_modules/whatsapp-web.js/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:391:56)
    at async ExecutionContext.evaluate (/home/xxxxxx/node_modules/whatsapp-web.js/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/ExecutionContext.js:277:16)
    at async IsolatedWorld.evaluate (/home/xxxxxx/node_modules/whatsapp-web.js/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/IsolatedWorld.js:100:16)
    at async CdpFrame.evaluate (/home/xxxxxx/node_modules/whatsapp-web.js/node_modules/puppeteer-core/lib/cjs/puppeteer/api/Frame.js:362:20)
    at async CdpPage.evaluate (/home/xxxxxx/node_modules/whatsapp-web.js/node_modules/puppeteer-core/lib/cjs/puppeteer/api/Page.js:830:20)
    at async PrivateChat.fetchMessages (/home/xxxxxx/node_modules/whatsapp-web.js/src/structures/Chat.js:204:24)
    at async /home/xxxxxx/Work/nodejs/server/wa-server.js:262:18
```

## Reproduction Steps
chat.fetchMessages()

## Code Sample
```typescript
chat.fetchMessages()
```

## User Setup

| WhatsApp | Type |
|---|---|
| **Account Type** | Standard |
| **Authentication Strategy** | LocalAuth |
| **WhatsApp Web Version** | 2.3000.1044515310 |
| **whatsapp-web.js Version** | 1.34.7 |

| Environment | Version |
|---|---|
| **Browser Type** | Chrome |
| **Browser Version** | 146.0.7680.31 |
| **Phone OS Version** | Android |
| **Running OS Version** | Linux Mint |
| **Node.js Version** | 20.18.2 |

## Checklist
- [x] **I use the latest released version of whatsapp-web.js.**
- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have verified this is a library issue, not a problem with my code.

MuhdSHiBiLi:
[...]

PR #201910, opened 2026-09-03
fix(Utils): resolve getChats() CDP 'r: r' on WhatsApp Web 2.3000.x ($1 rename + fault-isolated getChatModel)

cripperzprodigy:
## Problem Description

On WhatsApp Web `2.3000.x` (e.g. `2.3000.1046738589`), calling `client.getChats()` consistently throws a minified CDP evaluation error:
```text
Evaluation failed: r: r
    at Client.getChats (Client.js:1669:23)
```

### Root Causes
1. **Property Rename (`_serialized` → `$1`):** WhatsApp Web's minified bundler renamed internal WID and key object serialization properties from `_serialized` to `$1`. In `src/util/Injected/Utils.js`, calls like `.createWid(chat.id._serialized)` pass `undefined`, throwing minified `r: r` exceptions inside `#evaluate`.
2. **`Promise.all` Fail-Fast Trap:** `window.WWebJS.getChats` and `window.WWebJS.getChannels` map all chats through `getChatModel()` inside an unprotected `Promise.all`. If an account is in multiple groups/channels and ANY single chat throws (e.g. stale group metadata, kicked from group, LID migration glitch in `toPn`, or revoked newsletter), the entire `getChats()` batch rejects.

---

## Proposed Changes

1. **`_serialized || $1` Fallbacks:**
   Added fallback lookups across all 9 direct crash sites in `Injected/Utils.js` (e.g., `chat.id._serialized || chat.id.$1`).

2. **`__ensureSerializedIds` Model Normalizer:**
   Hooked a normalizer into `getMessageModel`, `getChatModel`, and `getContactModel` ensuring that models returned across the CDP boundary into Node.js always guarantee `id._serialized`.

3. **`__getChatModelSafe` Fault-Tolerant Resolver:**
   - Routes `getChats()` and `getChannels()` through `__getChatModelSafe`: if metadata enrichment fails for an individual group or channel, it falls back to basic chat serialization (`chat.serialize()`) instead of rejecting the whole batch.
   - Wrapped `groupMetadata.update()` and `newsletterMetadata.update()` in `try/catch` error boundaries.
   - Filtered unresolvable entries with `results.filter(Boolean)`.

---

## Verification
- Validated with automated unit tests including vm-sandbox simulation of throwing group metadata and unmigrated LID participants.
- Verified in live production on an active account with 2,560 contacts and multiple groups on WhatsApp Web `2.3000.1046738589` — `getChats()` successfully recovers offline message history without errors.

### DIFF
--- a/src/util/Injected/Utils.js
+++ b/src/util/Injected/Utils.js
@@ -3,6 +3,61 @@
 exports.LoadUtils = () => {
     window.WWebJS = {};
 
+    // ISSUE-388 (Phase 2): WhatsApp Web 2.3000.x renamed the internal wid
+    // property `_serialized` to `$1` on live model objects. The Utils.js
+    // crash sites use `obj._serialized || obj.$1` fallbacks (Phase 1), but
+    // the model builders below return `model.serialize()` output whose `id`
+    // may ALSO be missing `_serialized` if the serializer dropped it.
+    // Node-side structures (Chat.js:22 `this.id = data.id`, then
+    // `this.id._serialized` on every method) require that field, so every
+    // model we hand back to `page.evaluate()` callers must be normalized.
+    window.WWebJS.__widToSerialized = (wid) => {
+        if (wid === null || wid === undefined) return wid;
+        if (typeof wid === 'string') return wid;
+        if (typeof wid._serialized === 'string') return wid._serialized;
+        if (typeof wid.$1 === 'string') return wid.$1;
+        if (typeof wid.user === 'string' && typeof wid.server === 'string') {
+            return `${wid.user}@${wid.server}`;
+        }
+        return undefined;
+    };
+
+    // ISSUE-388 (Phase 2): Normalize a serialized model so all id-like
+    // fields carry the `_serialized` string the Node-side structures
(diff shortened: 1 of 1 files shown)


JFuentesTr (approved):

emrecirik:
**Confirmed on WhatsApp Web 2.3000.1047868043** (whatsapp-web.js 1.34.7, puppeteer 24.38.0, Chrome, Node 22).

For the record, the underlying page-side error behind the minified `r: r` is:

```
```

i.e. an `undefined` key (the missing `_serialized`) reaching IndexedDB. `chat.fetchMessages()` fails identically (#201884 looks like the same root cause). Restoring `_serialized` makes `getChats()` return all 344 chats and `fetchMessages()` work again.

One robustness note on the `$1` fallback: that name is a minified, build-specific symbol and will change on the next rename. Resolving it by value rather than by name has survived every build I've tested:

```js
for (const value of Object.values(key)) {
    if (typeof value === 'string' && /^(true|false)_/.test(value)) return value;
}
```

(as a last resort after `_serialized` and before the `user@server` reconstruction).

